PR #63, opened 2026-01-16
v3/LMS: Align settings page with LMS layout for better skinnability.

michaelherger:
* Remove the knob configuration from the LMS settings page: it only covered one knob, and was redundant with the configuration page built-in to the bridge.
* Don't use sync HTTP requests. Always use async calls.
* Add missing localization for a few buttons.
* Add input validator for the port.
* Leave prefs handling to the built-in handler in the base class for settings.

## Release Notes

* **New Features**
  * Redesigned knob management interface displaying knobs in a dynamic list with per-knob edit actions.
  * Added knob status information including Firmware and Battery status indicators.

* **Bug Fixes**
  * Implemented stricter port validation to ensure values fall within the valid range (1024–65535).

* **Changes**
  * Removed individual knob power and rotation preference settings from the configuration page.
  * Restructured settings interface for improved layout and organization.


muness:
@coderabbitai review

PR #64, opened 2026-01-16
v3/LMS: Implement a subscription to player state changes in LMS

michaelherger:
In LMS we can listen to events happening to a player. This change would listen for playback state changes.

This PR is to start the discussion. If you implemented an endpoint in the bridge which the plugin could call to trigger a metadata update, I could take care of the LMS side of things.

### DIFF
--- a/lms-plugin/Helper.pm
+++ b/lms-plugin/Helper.pm
@@ -476,6 +476,17 @@ sub writeKnobConfig {
     }
 }
 
+# Notify helper of playback state change
+sub notifyPlaybackStateChange {
+    my ($class, $client) = @_;
+    return unless running() && $client;
+
+    my $zoneId = 'lms:' . $client->id;
+    # TODO: Implement notification logic (e.g., HTTP request to helper API)
+
+    $log->debug("Notified helper of playback state change for zone $zoneId");
+}
+
 # Get knob status from running helper (if available)
 sub knobStatus {
     my $class = shift;
--- a/lms-plugin/Plugin.pm
+++ b/lms-plugin/Plugin.pm
@@ -43,6 +43,8 @@ sub initPlugin {
         Plugins::UnifiedHiFi::Helper->start;
     }
 
+    Slim::Control::Request::subscribe(\&_onPlaylistEvent, [['power', 'pause', 'stop', 'mixer', 'playlist'], ['newsong', 'pause', 'stop', 'play', 'volume']]);
+
     $log->info("Unified Hi-Fi Control plugin initialized");
 }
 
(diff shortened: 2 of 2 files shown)


muness:
From working on #68, I looked into LMS's notification system. Here's a spec outline:

## LMS Notification Subscription

**Commands:**
```
subscribe pause,play,stop,playlist,mixer,client
```

**Notifications received:**
```
04:20:00:12:23:45 pause
04:20:00:12:23:45 mixer volume 25
04:20:00:12:23:45 playlist newsong ...
```

## Design Considerations

1. **Connection type:** LMS CLI uses persistent TCP (telnet-style on port 9090), not HTTP. Current adapter uses HTTP JSON-RPC for commands. Would need a separate TCP connection for subscriptions.

2. **Useful notifications:**
   - `pause` / `play` / `stop` - playback state changes
   - `mixer` - volume changes
   - `playlist` - track changes (newsong, etc.)
   - `client` - player connect/disconnect

3. **Architecture options:**
   - Keep HTTP for commands (simple request/response)
   - Add TCP subscription for real-time state updates
   - Reduce/eliminate polling interval

**Reference:** [LMS CLI Documentation](https://github.com/elParaguayo/LMS-CLI-Documentation/blob/master/LMS-CLI.md#0.1_NC) - see `listen` and `subscribe` commands.

cc @mherger - would this approach work well with the LMS plugin?

michaelherger:
> cc @mherger - would this approach work well with the LMS plugin?

Your summary isn't 100% complete: we have a subscription mechanism over http. It's from before web sockets, [CometD](https://cometd.org). It "simulated" persistent connections using long-polling: a request would not be disconnected immediately, but only after a long time (don't know the details), and then the client would immediately re-connect. During that time messages could be sent both ways.

That said: this would only be necessary/recommended if you wanted to support LMS without a plugin. With the plugin it's LMS which could call the bridge on events. That way the bridge wouldn't need to have a persistent connection to subscribe to events. The subscription would be handled in the plugin, which in case of an event would push that to the bridge using some endpoint in the bridge.

I'd probably rely on the plugin for simplicity reasons. It could be extended to talk to any bridge, providing its own implementation for those users who don't care about supporting other platforms.

muness:
Superseded by #107 which implements CLI subscription approach instead of HTTP callbacks

michaelherger:
I personally believe this was the wrong decision. Keeping up a persistent connection certainly has a much higher overhead in terms of resource usage as well as code maintenance. Another network connection, another protocol. The exponential back-off alone (thanks for thinking of that!) will probably be more lines of code than the suggested solution where the LMS plugin would listen for events and call the bridge using a protocol that's being used already anyway.

In addition to efficiency it will add another source of configuration issues: what if a user decided to have the CLI listener on a different port? What if they had their desktop firewall misconfigured to block access to the CLI port?

Calling the bridge, which is needed to work anyway, would have been dead simple in comparison.

PR #65, opened 2026-01-16
v3/LMS: Get rid of some of the redundancy evaluating the binary to be used

michaelherger:
There were multiple implementations to try to figure out the platform, and what binary to use on each platform. Resulting in broken support for eg. `linux-armv7l`. Consolidate all evaluation in one place.

## Summary by CodeRabbit

* **New Features**
  * Added support for Linux ARMv7l systems

* **Improvements**
  * Enhanced platform detection and binary selection for improved reliability across different architectures
  * Better logging and debugging capabilities for troubleshooting binary-related issues

### DIFF
--- a/lms-plugin/Settings.pm
+++ b/lms-plugin/Settings.pm
@@ -106,7 +106,9 @@ sub beforeRender {
     # Add template variables
     $params->{'running'}    = Plugins::UnifiedHiFi::Helper->running();
     $params->{'webUrl'}     = Plugins::UnifiedHiFi::Helper->webUrl();
-    $params->{'binaries'}   = [Plugins::UnifiedHiFi::Helper->binaries()];
+    # Single binary per platform now - dropdown only shows if size > 1 (never)
+    my $platformBinary = Plugins::UnifiedHiFi::Helper::BINARY_MAP->{Plugins::UnifiedHiFi::Helper->detectPlatform()};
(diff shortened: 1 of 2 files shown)


muness:
@coderabbitai review

lms-plugin/Settings.pm
```diff
@@ -106,7 +106,9 @@ sub beforeRender {
     # Add template variables
     $params->{'running'}    = Plugins::UnifiedHiFi::Helper->running();
     $params->{'webUrl'}     = Plugins::UnifiedHiFi::Helper->webUrl();
-    $params->{'binaries'}   = [Plugins::UnifiedHiFi::Helper->binaries()];
+    # Single binary per platform now - dropdown only shows if size > 1 (never)
+    my $platformBinary = Plugins::UnifiedHiFi::Helper::BINARY_MAP->{Plugins::UnifiedHiFi::Helper->detectPlatform()};
+    $params->{'binaries'}   = $platformBinary ? [$platformBinary] : [];
```
michaelherger, line 111: I'm sorry about missing this. Do you expect there ever to be more than one executable on any platform? The old code would never have returned more than one either. We could further simplify this and the HTML template to not care about a list.

I know I have such a selector in one of my plugins (Spotty), and right now I'm wondering why... I'm probably the only person who would switch between the default and some custom build every now and then. But could this be a use case?

PR #105, opened 2026-01-18
feat: add LMS auto-discovery via UDP broadcast

### LINKED ISSUE #48: LMS discovery
I see that the LMS connection is configured by the user - however, good news is that LMS supports being discovered.
So discovery could be used to provide the default and then the user could override fields if really needed for some reason or if discovery failed - or if there is more than one LMS in the network.

It is how the players find LMS - and LMS uses it itself to find other LMS instances that are on the same subnet.

You can see related code at https://github.com/LMS-Community/slimserver/blob/776e969ec5f8101f20f7687f525d42674ea52900/Slim/Networking/Discovery.pm#L108

You can see it in action using socat
`echo -ne "eIPAD\x00NAME\x00JSON\x00UUID\x00VERS\x00" | socat -t5 - udp-datagram:255.255.255.255:3483,broadcast |od -Ax -bc`

It is possible that some OS will not allow a full broadcast ... in which case you could work out the current subnet and then just broadcast to that.
The response is TLV (Type-Length-Value) with the requested fields.

Type:
IPAD is the IP address of LMS - usually not returned in which case the IP address of the response is the LMS IP address (example of use is LMS in Docker Bridge mode)
NAME is the server (library) name
JSON is the port that you use subsequently
UUID is the unique id of this server instance ... in theory a server could change IP address but the UUID should persist
VERS is supposed to be the LMS version but it is not reliable since it fakes a response so that players with old firmware do not refuse to talk to it

Length is binary

muness:
Closes #48

## Summary
- Implement LMS server discovery using UDP broadcast on port 3483 (Squeezebox/Slim protocol)
- Add `lms_discovery.rs` module with TLV parsing for discovery responses
- Auto-configure LMS when exactly one server is found on startup (no manual config needed)
- Add `GET /lms/discover` endpoint for manual discovery and listing available servers
- Comprehensive test coverage (11 tests) for TLV parsing edge cases

## Test plan
- [x] `cargo test` passes (all 33 lib tests + integration tests)
- [x] `cargo clippy` passes with no warnings
- [x] API contract test updated for new `/lms/discover` endpoint
- [ ] Manual test: Start bridge without LMS config, verify auto-discovery finds server
- [ ] Manual test: Call `GET /lms/discover` to see available servers

## API Changes
Added new endpoint:
- `GET /lms/discover?timeout_ms=3000` - Discover LMS servers on the local network via UDP broadcast

Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added automatic LMS server discovery during startup for seamless configuration.
  * New API endpoint to discover available LMS servers with configurable timeout.
  * Auto-configuration when exactly one server is detected; manual guidance provided for zero or multiple servers.

### DIFF
--- a/src/adapters/lms.rs
+++ b/src/adapters/lms.rs
@@ -16,6 +16,7 @@ use tokio::time::interval;
 use tokio_util::sync::CancellationToken;
 use tracing::debug;
 
+use crate::adapters::lms_discovery::discover_lms_servers;
 use crate::bus::{BusEvent, PlaybackState, SharedBus, VolumeControl, Zone};
 use crate::config::get_config_dir;
 
@@ -424,6 +425,59 @@ impl LmsAdapter {
         }
     }
 
+    /// Attempt auto-discovery and configure if exactly one server is found.
+    /// Returns Ok(true) if auto-configured, Ok(false) if no single server found, Err on failure.
+    ///
+    /// Only auto-configures if:
+    /// - No existing configuration (host is None)
+    /// - Exactly one LMS server responds to discovery
+    pub async fn auto_discover_and_configure(&self) -> Result<bool> {
+        // Don't auto-configure if already configured
+        if self.is_configured().await {
+            tracing::debug!("LMS already configured, skipping auto-discovery");
+            return Ok(false);
+        }
+
+        tracing::info!("Attempting LMS auto-discovery...");
+
+        let servers = discover_lms_servers(None).await?;
+
+        match servers.len() {
+            0 => {
+                tracing::info!("LMS auto-discovery: no servers found");
+                Ok(false)
+            }
+            1 => {
+                let server = &servers[0];
+                tracing::info!(
+                    "LMS auto-discovery: found single server '{}' at {}:{}",
+                    server.name,
+                    server.host,
+                    server.json_port
+                );
+                // Auto-configure with discovered settings
+                self.configure(server.host.clone(), Some(server.json_port), None, None)
+                    .await;
+                Ok(true)
+            }
+            n => {
+                tracing::info!(
+                    "LMS auto-discovery: found {} servers, not auto-configuring (manual selection required)",
+                    n
+                );
+                for server in &servers {
+                    tracing::info!(
+                        "  - '{}' at {}:{}",
+                        server.name,
+                        server.host,
+                        server.json_port
+                    );
+                }
+                Ok(false)
+            }
+        }
+    }
+
     /// Configure the LMS connection
     pub async fn configure(
         &self,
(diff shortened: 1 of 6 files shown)


muness:
## Superego Review

Overall: **Well-implemented feature** that correctly follows the Squeezebox/Slim TLV protocol. The code is clean, well-documented, and has good test coverage.

### Protocol Correctness

The implementation correctly handles the LMS discovery protocol:

- **Port 3483** - Correct standard discovery port
- **TLV format** - Properly parses the 4-byte tag + 1-byte length + value structure
- **Response marker** - Correctly handles both `'E'` (uppercase response) and `'e'` (lowercase echo)
- **Request packet** - `"eIPAD\x00NAME\x00JSON\x00UUID\x00VERS\x00"` matches the expected format for requesting these fields

One observation: The protocol reference shows individual TLV values are capped at 255 bytes and total response at 1450 bytes. The 1024-byte receive buffer in `discover_lms_servers()` is slightly smaller than the 1450-byte max, but in practice server names rarely approach these limits so this should be fine for real-world use.

### Security Considerations

**UDP broadcast is acceptable here** because:
1. Discovery only runs when LMS is unconfigured (no constant broadcast traffic)
2. The socket binds to port 0 (ephemeral), not a well-known port
3. Only reads responses, doesn't execute arbitrary data
4. Uses `String::from_utf8_lossy` for safe string handling of untrusted data

**Minor consideration:** The `/lms/discover` endpoint exposes local network topology (LMS server IPs/names). If this API is exposed externally, you may want to consider rate limiting or access controls. For a home hi-fi bridge, this is likely fine.

### Error Handling

Good error handling overall:
- Graceful fallback when IPAD field is missing (uses source address)
- Handles truncated responses without panicking
- Invalid JSON port falls back to default 9000
- Missing NAME field correctly rejects the response
- Timeout handling with deadline-based loop is robust

### Test Coverage

Excellent test coverage (11 tests) covering:
- Empty/invalid responses
- Minimal valid responses
- Full responses with all fields
- Missing IPAD fallback
- Invalid port parsing
- Truncated data
- Lowercase response marker
- Serialization round-trip

**Suggestion:** Consider adding one integration test that mocks a UDP response to test the full `discover_lms_servers()` flow, though this is lower priority given the thorough unit tests.

### Auto-Configuration Logic

The auto-configure behavior is sensible:
- Only activates when not already configured
- **Exactly one server** = auto-configure (good UX for typical single-server setups)
- **Zero servers** = helpful error message pointing to manual config
- **Multiple servers** = lists all options, requires manual selection

This is the right design - it "just works" for the common case while providing clear guidance for edge cases.

### Minor Suggestions (Non-blocking)

1. **Buffer size comment:** Consider adding a comment explaining why 1024 bytes is sufficient (typical responses are ~100-200 bytes):
   ```rust
   // 1024 bytes is sufficient - typical responses are ~100-200 bytes,
   // max TLV response is 1450 bytes but server names rarely approach limits
   let mut buf = [0u8; 1024];
   ```

2. **Discovery timeout configurability:** The 3-second default is reasonable. For slower networks or Docker environments, users can already pass `timeout_ms` to the API endpoint, which is good.

### Verdict

**Approve** - Clean implementation that correctly follows the protocol, handles edge cases well, and provides good UX for automatic server discovery. The code is well-documented with references to the protocol specification.

PR #106, opened 2026-01-18
feat: organize config files into unified-hifi/ subdirectory (#76)

### LINKED ISSUE #76: V3: configuration file naming
At present there are a few different configuration files stored in the LMS prefs directory - but perhaps naming should be changed to make it clearer where they are from.
LMS does not force a sub-directory or naming prefix - so to avoiding name clashes is down to individual developers. Therefore using a consistent prefix might help.
Currently seen
knob-config.json
knobs.json
lms-config.json
app-settings.json

Perhaps precede all with unifiedhifi_

Another way to do it is to store in a folder in "cache".

@mherger perhaps something to document in https://lyrion.org/reference/music-service-plugin/ - but perhaps have a generic section that applies to all LMS hosted plugins

muness:
Closes #76

## Summary

- Organizes config files into a `unified-hifi/` subdirectory to reduce clutter
- Implements automatic migration on startup (moves files from config dir root to subdirectory)
- Adds backwards-compatible reading (checks subdirectory first, falls back to root for legacy files)
- Updates all adapters and modules to use new config file helpers

## New Structure

```
~/.config/unified-hifi-control/
└── unified-hifi/
    ├── app-settings.json
    ├── lms-config.json
    ├── hqp-config.json
    ├── hqp-zone-links.json
    ├── roon_state.json
    └── knobs.json
```

## Migration Safety

- Won't overwrite existing files in subdirectory
- Handles cross-device moves (copy + delete fallback)
- Also migrates `roon_state.json` from XDG_DATA_HOME on Linux (where data dir differs from config dir)

## Test Plan

- [x] All existing tests pass
- [x] New tests cover migration scenarios
- [x] Clippy passes with no new warnings
- [x] TDD approach followed per AGENTS.md

Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Configuration files are now organized in a unified subdirectory with automatic migration from legacy locations.
  * Added backward-compatible fallback to support existing config file locations.

* **Chores**
  * Simplified knob store initialization to remove explicit directory parameters.
  * Refactored configuration file path resolution for centralized management.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3649,6 +3649,19 @@ dependencies = [
 ]
 
 [[package]]
+name = "tempfile"
+version = "3.24.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "655da9c7eb6305c55742045d5a8d2037996d61d8de95806335c7c86ce0f82e9c"
+dependencies = [
+ "fastrand",
+ "getrandom 0.3.4",
+ "once_cell",
+ "rustix",
+ "windows-sys 0.61.2",
+]
+
+[[package]]
 name = "thiserror"
 version = "1.0.69"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -4254,6 +4267,7 @@ dependencies = [
  "sha2",
  "ssdp-client",
  "syn",
+ "tempfile",
  "thiserror 2.0.17",
  "tokio",
  "tokio-stream",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -133,6 +133,7 @@ tokio-test = "0.4"
 serial_test = "3"
 syn = { version = "2", features = ["full", "parsing", "visit"] }
 walkdir = "2"
+tempfile = "3"
 
 [profile.release]
 lto = true
--- a/src/adapters/hqplayer.rs
+++ b/src/adapters/hqplayer.rs
@@ -25,7 +25,7 @@ use crate::bus::{
     BusEvent, NowPlaying as BusNowPlaying, PlaybackState, SharedBus, TrackMetadata,
     VolumeControl as BusVolumeControl, VolumeScale, Zone as BusZone,
 };
-use crate::config::get_config_dir;
+use crate::config::{get_config_file_path, read_config_file};
 
 const HQP_CONFIG_FILE: &str = "hqp-config.json";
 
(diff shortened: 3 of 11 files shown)


muness:
## Superego Review

Overall this is a well-structured PR with good safety measures. The migration logic is sound and the backwards compatibility approach is correct. A few observations:

### What Works Well

1. **Migration safety**: The `migrate_single_file` function correctly:
   - Skips if source does not exist
   - Skips if destination already exists (preserves user data)
   - Handles cross-device moves with copy+delete fallback
   - Logs clearly at each step

2. **Backwards compatibility**: `read_config_file()` checks subdirectory first, then falls back to root - this is the right approach for seamless upgrades.

3. **Test coverage**: Good tests for the key migration scenarios including the "don't overwrite existing" case.

### Observations (Non-Blocking)

1. **Migration order in main.rs**: Migration runs *after* `load_config()` but *before* adapters load their configs. This is correct - the main `config.toml` stays at root (not migrated), while adapter JSON files get migrated before adapters initialize. Just noting this is intentional and works correctly.

2. **Roon adapter difference**: Unlike other adapters that use `read_config_file()` for fallback reading, the Roon adapter uses `get_config_file_path()` directly for writes/reads. This is fine because:
   - Migration handles moving `roon_state.json` from both config and data dirs
   - After migration, file will be in the subdirectory
   - Roon creates the parent dir before writing (line 558-562)

3. **Cross-device edge case**: In `migrate_single_file`, if the copy succeeds but delete fails, the user ends up with files in both locations. The subdir takes precedence on reads, so this is safe but leaves orphan files. The warning log is appropriate.

4. **Data dir migration for roon_state.json**: Good catch handling the Linux case where XDG_DATA_HOME differs from XDG_CONFIG_HOME - this avoids losing existing Roon pairing state.

### Test Gap (Minor)

Consider adding a test for the cross-device fallback path (copy+delete). Currently only `std::fs::rename` success is tested. Could mock/force the rename failure to ensure the fallback works. Not blocking - the implementation looks correct.

### Verdict

**Approve** - Migration logic is safe, backwards compatibility works correctly, no data loss risks. The careful handling of "don't overwrite existing" and cross-device moves shows good defensive programming.

---
*Reviewed by Superego (Claude Code)*

PaulWebster:
For LMS the top-level directory will have to be considered carefully.
Should it be writing to OS/User location?
I know that it should not be writing to InstalledPlugins/Plugins/UnifiedHiFi because the plugin installer removes that directory during the update process.

I do something like this ... but @mherger might have a one-liner.


```
my $serverprefs = preferences('server');
my $servercache = $serverprefs->get('cachedir');
		
if ( $servercache ) {
	$cacheDir = catdir($servercache, $pluginName);
			
	# Create cache directory if needed
	mkdir $cacheDir unless -d $cacheDir;

	if ( -w $cacheDir ) {
                  # can save stuff in here ...

```
Then Plugin can pass config location to Bridge when it runs it

PaulWebster:
Ah - I see LMS Plugin code already has stuff for this and passes to Bridge
https://github.com/open-horizon-labs/unified-hifi-control/blob/d5965ce9dff67734bfb5f976b3f671293e6fdfc3/lms-plugin/Helper.pm#L460C5-L460C39

PR #107, opened 2026-01-18
feat: add event-driven LMS updates via CLI subscription (#77)

### LINKED ISSUE #77: v3: LMS poll interval should be configurable
## Problem
The LMS poll interval is fixed at 2 seconds in the Rust version, whereas it was configurable in the Node.js version via config file.

On low-powered LMS servers (e.g., RPi 3B) with many players (22+), this causes high CPU usage on the LMS server.

## Proposed Solution
- Make the poll interval configurable (environment variable or config file) as a stopgap while waiting for the larger polling architecture changes.
- Default could remain at 2 seconds, but allow users to increase it (e.g., 5-10 seconds) for resource-constrained systems.
- Use a different polling for inactive, perhaps with exponential back off up to 30s

## Current Location
`src/adapters/lms.rs` - `POLL_INTERVAL` constant is hardcoded to 2 seconds.

muness:
Closes #77

## Summary

Adds event-driven LMS updates via CLI subscription on port 9090, replacing constant polling with push-based notifications.

### Core Implementation
- **CLI Subscription**: Persistent TCP connection to LMS CLI (port 9090)
- **Event Types**: Handles playlist, mixer (volume + mute), power, and client events
- **Reconnection**: Exponential backoff (5s → 60s cap), resets on successful connection
- **TCP Timeout**: 120s read timeout detects unresponsive LMS

### Adaptive Polling
- **With subscription active**: Poll every 30s (configurable via `LMS_POLL_INTERVAL × 15`)
- **Without subscription**: Poll every 2s (configurable via `LMS_POLL_INTERVAL`)
- Polling continues as fallback when CLI subscription fails

### Configuration
- `LMS_POLL_INTERVAL` env var: Base poll interval in seconds (default: 2)

### Event Parsing
- URL-decodes player IDs (MAC addresses as `00%3A04%3A20%3A...`)
- Parses space-separated event format per LMS CLI spec
- 15 unit tests for parsing edge cases

## Test Plan
- [x] Unit tests for CLI event parsing (15 tests)
- [x] All existing tests pass (207 total)
- [x] Clippy clean
- [ ] Manual test with LMS server

## Summary by CodeRabbit

* **New Features**
  * Real-time LMS event subscription delivering instant updates for playlist, mixer, power, and client changes (including proper decoding of track/player names).

* **Performance Improvements**
  * Adaptive polling that shortens update intervals while a subscription is active for faster UI refreshes.

* **Reliability Improvements**
  * Robust reconnection/backoff, read-timeouts, and lifecycle handling for more resilient live updates and timely propagation of now-playing, volume, power, and zone changes.

* **Tests**
  * Extensive parsing and event-driven tests covering varied event formats and edge cases.

### DIFF
--- a/src/adapters/lms.rs
+++ b/src/adapters/lms.rs
@@ -60,6 +60,8 @@ const POLL_INTERVAL_WITH_SUBSCRIPTION: Duration = Duration::from_secs(30);
 const CLI_RECONNECT_DELAY: Duration = Duration::from_secs(5);
 /// Maximum reconnection delay (exponential backoff cap)
 const CLI_MAX_RECONNECT_DELAY: Duration = Duration::from_secs(60);
+/// TCP read timeout for CLI subscription (detect unresponsive LMS)
+const CLI_READ_TIMEOUT: Duration = Duration::from_secs(120);
 
 // =============================================================================
 // CLI Event Parsing
@@ -1006,16 +1008,23 @@ async fn run_cli_subscription(
                 break;
             }
             Err(e) => {
+                // Check if we had a successful connection before failing
+                let was_connected = {
+                    let mut s = state.write().await;
+                    let was_active = s.cli_subscription_active;
+                    s.cli_subscription_active = false;
+                    was_active
+                };
+
+                // Reset backoff if we had a working connection that later failed
+                if was_connected {
+                    retry_delay = CLI_RECONNECT_DELAY;
+                }
+
                 warn!(
                     "CLI subscription error: {}. Reconnecting in {:?}...",
                     e, retry_delay
                 );
-
-                // Mark subscription as inactive
-                {
-                    let mut s = state.write().await;
-                    s.cli_subscription_active = false;
-                }
 
                 // Wait before reconnecting (with shutdown check)
                 tokio::select! {
@@ -1024,8 +1033,10 @@ async fn run_cli_subscription(
                         break;
                     }
                     _ = tokio::time::sleep(retry_delay) => {
-                        // Exponential backoff up to max
-                        retry_delay = (retry_delay * 2).min(CLI_MAX_RECONNECT_DELAY);
+                        // Exponential backoff up to max (only if initial connection failed)
+                        if !was_connected {
+                            retry_delay = (retry_delay * 2).min(CLI_MAX_RECONNECT_DELAY);
+                        }
                     }
                 }
             }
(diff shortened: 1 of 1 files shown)


muness:
## Superego Review

Overall this is a well-structured implementation. The approach of layering event-driven updates on top of polling fallback is sound. A few observations:

### LMS CLI Protocol Implementation

**Correct:**
- The subscribe command format `subscribe playlist,mixer,power,client\n` matches the LMS CLI documentation
- URL-decoding of player IDs (MAC addresses encoded as `00%3A04%3A20%3A...`) is handled correctly
- Event line parsing (space-separated, URL-encoded fields) aligns with LMS behavior

**Minor concern:**
- The `newsong` event parsing assumes the track name is at index 3 and index at 4. Per LMS docs, the format is `<playerid> playlist newsong [<current_title>] [<playlist index>]`. Both are optional, which the code handles with `Option`. However, the track name could contain URL-encoded spaces which would appear as a single token after decoding - this looks correctly handled by decoding after splitting.

### TCP Connection Handling

**Good:**
- Exponential backoff with cap (5s -> 60s max) for reconnection
- Clean shutdown handling via `CancellationToken` in both connection and read loops
- Subscription state properly tracked for polling interval adjustment

**Potential issues:**

1. **No TCP keepalive or read timeout** - If the LMS server becomes unresponsive without closing the connection (e.g., firewall drops packets silently), `read_line` could block indefinitely. Consider adding:
   ```rust
   let stream = TcpStream::connect(&addr).await?;
   stream.set_nodelay(true)?;  // Optional: reduce latency
   // TCP keepalive would need socket2 crate for fine-grained control
   ```
   Or wrap the read in a `tokio::time::timeout()`.

2. **Backoff reset** - The retry_delay is reset to the initial value only implicitly when `connect_and_subscribe` returns `Ok(())`. On successful connection that later fails, the backoff doesn't reset. Consider resetting `retry_delay = CLI_RECONNECT_DELAY` after a successful subscribe.

### Event Parsing Edge Cases

The parsing is solid. A few observations:

1. **Mixer events default to 0** - Line 134: `let value = parts.get(3).and_then(|s| s.parse().ok()).unwrap_or(0);` - If the value parsing fails, defaulting to 0 could cause unexpected volume changes. Consider making this return `Unknown` instead.

2. **Muting events parsed as volume** - The test `test_parse_cli_event_mixer_muting` shows muting events are parsed as `Mixer { param: "muting", value: 1 }`, but `handle_cli_event` only handles `param == "volume"` (line ~1090). Muting events are silently ignored. This is probably fine, but worth noting the LMS comment at line 1098: `is_muted: false, // LMS doesn't expose mute via CLI events` - actually it does via `mixer muting`, it's just not handled.

### Adaptive Polling

**Correct approach:**
- Polling slows to 30s when subscription active, resumes at 2s on disconnect
- The polling task checks `cli_subscription_active` each tick and adjusts

**Minor inefficiency:**
- Creating a new `interval` on each interval change (lines 681-685) resets the timer. The first tick after changing intervals will occur immediately. This is probably fine but could cause a burst of polls right after subscription connects/disconnects.

### Concurrency and Resource Handling

**Good:**
- Write locks are held briefly (update cached state, release)
- The CLI task and polling task share state via `Arc<RwLock<LmsState>>`
- Spawned tasks properly respond to cancellation

**No leaks detected:**
- TCP stream dropped when reader/writer go out of scope
- Tasks exit cleanly on shutdown

### Test Coverage

The 15 unit tests for `parse_cli_event` are comprehensive:
- All event types covered (playlist variants, mixer, power, client)
- Edge cases: empty lines, single tokens, unencoded player IDs, special characters
- URL decoding verified

**Missing test coverage:**
- No integration tests for `handle_cli_event` or `connect_and_subscribe` (would require mocking, reasonable to skip)
- No tests for the adaptive polling interval logic
- No tests verifying the exponential backoff behavior

### Recommendations

1. Consider adding a read timeout to prevent hanging on unresponsive connections
2. Reset backoff delay after successful subscription
3. Either handle `mixer muting` events or document why they're intentionally ignored

Overall: **Good to merge** with the above as optional follow-up improvements. The core implementation is correct and the event-driven approach will significantly reduce CPU usage for multi-player setups.

PR #118, opened 2026-01-19
More LMS plugin cleanup

michaelherger:
* Remove support for different log levels: The bridge helper is set up to pipe all output to the null device. Therefore there's no reason to have a configuration for the log level. If we wanted to support logging from the bridge we could use the plugin's `plugin.unifiedhifi` logging configuration and map this to the bridge's levels.
* Remove port setting for the bridge binary. This would rarely be needed at all. Leave the pref there, so people in dire could edit the configuration file to change port.
* Remove support for multiple binaries. It wasn't really supporting more than one anyway.

## Summary by CodeRabbit

* **Changes**
  * Simplified settings UI by removing port, binary selection, and log level options.
  * Added localized messages for binary download status and live download progress.
  * Knob editor now opens the knobs management list instead of individual knob detail pages.
  * Knobs without a name display a default label ("The Knob") in the list.


muness:
Appreciate the improvements! #119 changes made it easier to port the changes manually:

  - Process reaping (->wait after ->die) to prevent zombies
  - String externalization for download status messages
  - Knob edit link fix (route to /knobs since per-knob URLs aren't supported)
  - Remove debug file logging.

PR #129, opened 2026-01-21
refactor(lms): Bundle all binaries in universal ZIP

### LINKED ISSUE #125: v3/LMS: binary handling is overly complicated
There is support for a binary within the installation folder (for dev purposes), as well as a way to download a binary for the leaner plugin installation archives. Both of them are supported by two or three subs, even if they're all only used once, and only executing a single line of code. This makes it hard to follow the flow for no good reason.

LMS has helpers to find binaries, once binary folders are registered with it. Instead of dealing with the binaries in the plugin, you could simply register the two paths at initialisation time:

```perl
Slim::Utils::Misc::addFindBinPaths(pluginBinDir(), bundledPublicDir());

my $platform = $class->detectPlatform();
my $binaryName = BINARY_MAP->{$platform};

my $binary = Slim::Utils::Misc::findBin($binaryName);
```

That way you actually wouldn't even need to deal with two paths, as you could simply stick your own local build in any folder represented in your `PATH` environment variable. One less path to deal with in the plugin. Oh, and it know Windows needs an extension, too. So no need to deal with this in the plugin.

You decided not to merge my removal of the multiple binaries handling, despite the fact that only one binary is supported. Why have a setting which will never be shown? Why have a method returning a single binary, to then stuff that single value in an array of length 1, which would not render the list? Should the plugin support multiple binaries? If so, let's fix it. Otherwise all that code can go away.

So to summarize my suggestions:
* remove support for the `bin` pref - it's not really supported and can cause problems if you decide to store binaries elsewhere in the future
* remove handling of a locally available binary in the plugin's `Bin` folder
* use LMS' built in binary lookup handling - only register your binary download folder for lookups
* store a local copy of the binary in a folder which is represented in the search path already

muness:
## Summary
- Replace download-on-demand with pre-bundled universal ZIP containing all platform binaries
- Use LMS's built-in `findBin()` with platform folder structure
- Remove ~485 lines of download/async code

## Changes
- **build.yml**: Replace `build-lms-bootstrap` + `build-lms-full` matrix with single `build-lms-universal`
- **Helper.pm**: Remove download machinery, use `Slim::Utils::Misc::findBin()`
- **Settings.pm**: Remove binary dropdown
- **basic.html**: Simplify status UI
- **strings.txt**: Remove download-related strings

## ZIP Structure
```
Bin/
  darwin/unified-hifi-control
  x86_64-linux/unified-hifi-control
  aarch64-linux/unified-hifi-control
  arm-linux/unified-hifi-control
  MSWin32-x64-multi-thread/unified-hifi-control.exe
  public/
```

Closes #125

🤖 Generated with [Claude Code](https://claude.ai/code)

## Release Notes

* **New Features**
  * LMS plugin now uses universal binary packaging with all platform binaries bundled together.

* **Bug Fixes**
  * Simplified binary management and discovery process.

* **UI Changes**
  * Removed manual binary selection interface from plugin settings.
  * Added "Binary Missing" alert when binary is unavailable.
  * Streamlined settings interface for improved clarity.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -73,7 +73,6 @@ jobs:
       build_qnap_arm: ${{ steps.decide.outputs.build_qnap_arm }}
       build_docker: ${{ steps.decide.outputs.build_docker }}
       build_linux_packages: ${{ steps.decide.outputs.build_linux_packages }}
-      lms_matrix: ${{ steps.decide.outputs.lms_matrix }}
     steps:
       - name: Decide what to build
         id: decide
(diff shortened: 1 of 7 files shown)


muness (commented):
## Review (Opus channeling @michaelherger's perspective)

Looking at this through the lens of the original issue feedback...

### What's right

1. **Using findBin() correctly** - The new bin() method does exactly what I suggested - register path, call findBin, let LMS handle platform detection.

2. **Platform folder structure** - Using darwin/, x86_64-linux/, aarch64-linux/, arm-linux/, MSWin32-x64-multi-thread/ matches LMS convention.

3. **Removed the download machinery** - No more downloadBinary(), downloadFile() (which were duplicated anyway), ensureBinary(), async HTTP handlers. That's the bulk of the complexity gone.

4. **Removed the bin pref** - Settings.pm no longer manages binary selection. That setting was never really used anyway.

### Questions

1. Web assets in Bin/public/ - does the bridge find them correctly via CONFIG_DIR?

2. Updates now come with plugin updates. That's actually the right answer.

### Overall

Helper.pm: ~575 lines → ~295 lines. The download machinery, platform detection, caching logic - all gone. Remaining code is just process management which is actually useful.

This is what I was asking for.

muness:
## Superego Review

### Issues Found & Fixed

- [x] **Documentation inconsistency** - docs/gh-release.md referenced `build-lms-full` → updated to `build-lms-universal`
- [x] **Dead preference** - Plugin.pm still had `bin => undef` → removed

### Not Issues

- ~~**Windows process spawning** - Original code also used `/bin/sh`. Not a regression.~~

### Verified Correct

- [x] Clean removal of download code - no dead remnants
- [x] Correct LMS API usage (`addFindBinPaths` + `findBin`)
- [x] Proper async pattern in `knobStatus()`
- [x] CI uses correct platform folder naming
- [x] Template uses LMS `WRAPPER setting` pattern
- [x] macOS Gatekeeper handling with `xattr -cr`

### Result

Helper.pm: ~575 → ~295 lines (-49%)
Total: -485 lines net

muness:
Fixed Windows process spawning - now uses `Proc::Background` directly on Windows instead of `/bin/sh` wrapper.

PR #139, opened 2026-01-22
fix(roon): Prevent high CPU and ensure graceful disconnect handling

### LINKED ISSUE #128: High CPU usage and duplicate processes when Roon server disconnects
## Summary

Bridge goes into high CPU state with duplicate processes when Roon server is disconnected.

## Environment

- Version: v3.1.0-rc.0
- Installation: Roon Extension Manager on Pi-like machine
- Bridge runs separately from Roon server

## Symptoms

1. **Knob connectivity lost** - Knob didn't connect to bridge when turned on; required bridge restart + knob power cycle
2. **High CPU/temperature** - Machine temperature rose significantly around the time Roon server disconnected
3. **Duplicate processes** - Two `/app/unified-hifi-control` processes running, both consuming high CPU
4. **Restart fixes it** - Restarting via Extension Manager returns to normal

## Reproduction

1. Run bridge on separate machine from Roon server
2. Shut down Roon server
3. Observe: bridge spawns duplicate process, high CPU consumption

## Expected Behavior

Bridge should gracefully handle Roon disconnection without resource leak or process duplication.

## Workaround

Restart bridge after Roon server goes down (or run bridge on same machine as Roon server).

## Investigation Notes

Likely related to Roon reconnection logic in `src/adapters/roon.rs` - may be spawning reconnect tasks that don't get cleaned up, or mDNS discovery going into a tight loop.

muness:
## Summary
- Fix Roon adapter event loop to prevent high CPU when server disconnects
- Add cancellation handling to gracefully respond to shutdown requests
- Extend spawn_cancellation_lint test to catch JoinSet::spawn patterns

## Test plan
- [x] Extended spawn_cancellation_lint to detect the bug (test failed as expected)
- [x] Fixed code (test now passes)
- [x] All 195+ tests pass
- [x] Superego review completed

Fixes #128


muness:
## Superego Review

**The fix looks solid.** The approach correctly addresses both problems identified, and the TDD methodology is sound.

### What works well

1. **The core fix in `src/adapters/roon.rs` (lines 614-631)** is correct:
   - `tokio::select!` enables cancellation via the shutdown token
   - The `let Some(...) = event_result else { break }` pattern properly handles channel close (no more infinite spin)
   - Good tracing messages for both shutdown paths

2. **The lint test extension** (`tests/spawn_cancellation_lint.rs` lines 63-68, 98-109) correctly catches `handles.spawn()` patterns, not just `tokio::spawn()`. This is important because `JoinSet::spawn` has the same cancellation requirements.

3. **Consistency with other adapters** - Checked the LMS adapter's polling loop (lines 772-801) and it already uses the same `tokio::select!` pattern. This fix brings the Roon adapter in line with the rest of the codebase.

### One observation

The outer reconnection loop in `start_internal()` (line 228-276) also has proper cancellation handling via `tokio::select!` at two points:
- Line 235-243: Wraps the entire `run_roon_loop` call
- Line 265-274: Checks before the retry sleep

This is thorough - shutdown can interrupt both the active loop and the reconnection backoff.

### Minor note

The `handles.spawn()` pattern detection is broad (catches any method named `spawn`), which is pragmatic for this codebase. False positives would just cause extra scrutiny of already-correct code.

### Verdict

The fix is well-reasoned and the implementation is clean. The TDD approach (extend lint test to catch the bug, then fix the code) is exactly right for this kind of structural invariant.

muness:
Closing - will recreate with proper branch workflow (feature branch -> v3)

PR #164, opened 2026-01-25
fix(lms): Emit NowPlayingChanged during polling, propagate CLI errors

### LINKED ISSUE #163: LMS now_playing data not propagating to knob endpoint
## Problem

After the v3.1.5 release (LMS reconnect improvements), the `/knob/now_playing` endpoint shows stale data for LMS zones while the `/lms/player/{id}` endpoint shows correct current data.

## Reproduction

1. Play music from Tidal (or any non-Internet Radio source) on an LMS player
2. Query `/lms/player/{player_id}` - shows correct track info
3. Query `/knob/now_playing?zone_id=lms:{player_id}` - shows stale/old track info

## Observed Behavior

**LMS endpoint (correct):**
```json
{
  "title": "Best for Last",
  "artist": "Adele",
  "album": "19",
  "artwork_url": "...tidal.com/images/..."
}
```

**Knob endpoint (stale):**
```json
{
  "line1": "FLOAT",
  "line2": "TIM & THE GLORY BOYS",
  "image_key": "...tunein.com/..."
}
```

## Root Cause (suspected)

The aggregator is not receiving `NowPlaying` updates from the LMS adapter. The LMS adapter fetches data correctly but the event bus connection to the aggregator appears broken.

## Related

- Likely introduced by #161 (unify retry logic in AdapterHandle)
- May be related to #147 or #150

muness:
## Summary

- Polling now emits `NowPlayingChanged` events when track metadata changes, ensuring aggregator stays updated even without CLI subscription
- CLI subscription failure propagates error to trigger AdapterHandle retry with exponential backoff

## Problem

After v3.1.5, `/knob/now_playing` showed stale data while `/lms/player/{id}` showed correct data. Root cause: LMS adapter's polling loop updated local cache but didn't emit bus events for the aggregator.

## Test plan

- [ ] Build and run locally
- [ ] Verify `/knob/now_playing` updates when changing tracks on LMS
- [ ] Verify CLI subscription reconnects after LMS restart

Fixes #163

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved per-player change detection, batched event publishing (now including cleared metadata), and more resilient background polling so CLI failures no longer stop updates.

* **Tests**
  * Added an integration regression test ensuring Now Playing events report cleared metadata (None) after metadata removal; added test setup cleanup to avoid config leakage.

* **Chores**
  * Updated repository ignore header and added a new ignore entry.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,9 +12,10 @@ firmware/version.json
 # Tailwind standalone CLI binary (downloaded via `make setup-tailwind`)
 /tailwindcss
 
-# Superego
+# Open Horizon Labs tools
 .superego/
 .wm/
+.ba/
 /target/
 # Note: Cargo.lock is committed for reproducible binary builds
 dist/
--- a/src/adapters/lms.rs
+++ b/src/adapters/lms.rs
@@ -3,15 +3,38 @@
 //! Implements the JSON-RPC protocol over HTTP and CLI event subscription over TCP.
 //! Documentation: http://HOST:9000/html/docs/cli-api.html
 //!
-//! ## Event-Driven Updates
+//! ## Architecture (Issue #165)
 //!
-//! This adapter supports real-time event notifications via the LMS CLI telnet protocol.
-//! When enabled, the adapter maintains a persistent TCP connection to LMS port 9090 and
-//! subscribes to `playlist,mixer,power,client` events. This dramatically reduces CPU usage
-//! compared to polling, especially with many players (22+).
+//! This module contains two logically separate concerns that share state:
 //!
-//! The polling fallback runs at reduced frequency (15x base interval) when subscription is active.
-//! The base poll interval can be configured via `LMS_POLL_INTERVAL` env var (default: 2 seconds).
+//! 1. **Polling** (HTTP JSON-RPC on port 9000)
+//!    - Discovers LMS server and players
+//!    - Polls player status at configurable interval
+//!    - Primary mechanism - always works
+//!
+//! 2. **CLI Subscription** (TCP telnet on port 9090)
+//!    - Subscribes to real-time events: playlist, mixer, power, client
+//!    - Enhancement for faster updates and lower CPU
+//!    - Optional - polling continues if CLI unavailable
+//!
+//! ## Interaction Model
+//!
+//! The two paths coordinate via a single shared flag: `cli_subscription_active`
+//!
+//! ```text
+//! CLI connects    → flag = true  → Polling slows to 30s interval
+//! CLI fails/exits → flag = false → Polling speeds to 2s interval (immediate)
+//! CLI reconnects  → flag = true  → Polling slows again
+//! ```
+//!
+//! Currently both run within a single adapter. Future refactor (Issue #165) will
+//! split into two independent adapters, each with AdapterHandle retry, sharing
+//! only the `cli_subscription_active` flag.
+//!
+//! ## Configuration
+//!
+//! - `LMS_POLL_INTERVAL`: Base poll interval in seconds (default: 2)
+//! - When CLI active, polling runs at 15x base interval (default: 30s)
 
 use anyhow::{anyhow, Result};
 use async_trait::async_trait;
--- a/src/adapters/lms.rs
+++ b/src/adapters/lms.rs
@@ -85,6 +85,15 @@ const CLI_READ_TIMEOUT: Duration = Duration::from_secs(120);
 // CLI Event Parsing
 // =============================================================================
 
+/// Now playing update data for bus emission
+struct NowPlayingUpdate {
+    player_id: String,
+    title: Option<String>,
+    artist: Option<String>,
+    album: Option<String>,
+    image_key: Option<String>,
+}
+
 /// Parsed CLI event from LMS
 #[derive(Debug, Clone, PartialEq)]
 pub enum CliEvent {
@@ -931,14 +940,7 @@ async fn update_players_internal(
         { state.read().await.players.keys().cloned().collect() };
 
     // Collect updates to emit after releasing the lock
-    // NowPlayingChanged: (player_id, title, artist, album, image_key)
-    let mut now_playing_updates: Vec<(
-        String,
-        Option<String>,
-        Option<String>,
-        Option<String>,
-        Option<String>,
-    )> = Vec::new();
+    let mut now_playing_updates: Vec<NowPlayingUpdate> = Vec::new();
     // LmsPlayerStateChanged: (player_id, state)
     let mut state_updates: Vec<(String, String)> = Vec::new();
     // VolumeChanged: (player_id, volume)
@@ -996,13 +998,13 @@ async fn update_players_internal(
 
         if now_playing_changed {
             // Emit even when metadata clears (all fields empty) so UI can update
-            now_playing_updates.push((
-                player.playerid.clone(),
-                to_option(&player.title),
-                to_option(&player.artist),
-                to_option(&player.album),
-                player.artwork_url.clone().or(player.coverid.clone()),
-            ));
+            now_playing_updates.push(NowPlayingUpdate {
+                player_id: player.playerid.clone(),
+                title: to_option(&player.title),
+                artist: to_option(&player.artist),
+                album: to_option(&player.album),
+                image_key: player.artwork_url.clone().or(player.coverid.clone()),
+            });
         }
 
         if state_changed {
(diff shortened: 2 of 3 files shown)


muness:
## Design Options Considered

### Option A: Independent Parallel Tasks

```
┌─────────────────┐     ┌─────────────────┐
│  Polling Task   │     │   CLI Task      │
│  (always runs)  │     │  (own retry)    │
└────────┬────────┘     └────────┬────────┘
         │                       │
         │  NowPlayingChanged    │  NowPlayingChanged
         └───────────┬───────────┘
                     ▼
               [Aggregator]
```

- Polling: emits events, runs at slow rate when CLI active
- CLI: has own retry loop with exponential backoff, independent of polling
- Adapter restart: only when polling fails critically

**Pros:**
- CLI failure doesn't disrupt polling
- Maximum uptime for polling

**Cons:**
- Two retry mechanisms (AdapterHandle + CLI-specific)
- More complex code, duplicated retry logic

---

### Option B: CLI Failure Restarts Adapter

```
┌─────────────────────────────────┐
│         AdapterHandle           │
│    (manages retry/restart)      │
└────────────────┬────────────────┘
                 ▼
┌─────────────────────────────────┐
│    run() = Polling + CLI        │
│    Either failure → Err         │
└─────────────────────────────────┘
```

- If CLI fails → adapter returns `Err` → AdapterHandle retries whole adapter
- Polling restarts too (brief interruption during backoff)

**Pros:**
- Single retry mechanism (AdapterHandle)
- Simpler code, no duplicate retry logic

**Cons:**
- CLI flakiness restarts polling unnecessarily
- ~5-60s polling gap during reconnection

---

### Option C: Polling Primary, CLI Optional Enhancement

```
┌─────────────────┐
│  Polling Task   │──────► NowPlayingChanged (always)
│  (primary)      │
└────────┬────────┘
         │ spawns
         ▼
┌─────────────────┐
│   CLI Task      │──────► NowPlayingChanged (when active)
│  (enhancement)  │
└─────────────────┘
```

- Polling always emits events
- CLI is "nice to have" - reduces polling frequency when active
- CLI has own retry but failure is non-fatal
- Adapter only restarts if polling fails

**Pros:**
- Robust - always works via polling
- CLI failure is graceful degradation

**Cons:**
- Two retry mechanisms
- CLI might stay broken forever if it fails but polling works

muness:
## Decision: Option B

We chose **Option B: CLI Failure Restarts Adapter** for these reasons:

### 1. Single Retry Mechanism

AdapterHandle already provides robust retry with exponential backoff (5s initial, 60s max, reset after 30s stable run). Duplicating this logic for CLI-specific retry adds complexity and potential for bugs.

### 2. Acceptable Trade-off

The polling gap during CLI reconnection (~5s in most cases) is acceptable because:
- Knob clients poll every 250ms - they'll recover quickly
- Web UI uses SSE which reconnects automatically
- The aggregator retains last-known state

### 3. Consistent Architecture

Other adapters (Roon, OpenHome, UPnP) use AdapterHandle retry for all failures. LMS should follow the same pattern rather than having special-case retry logic.

### 4. CLI Reconnection is Guaranteed

With Option B, CLI reconnection is **always** attempted on failure. With Option C, if CLI fails but polling works, CLI stays broken forever - it only gets another chance if the whole adapter restarts.

### What This PR Does

1. **Polling emits `NowPlayingChanged`** - ensures aggregator gets updates even without CLI
2. **CLI failure returns `Err`** - triggers AdapterHandle retry, which restarts both polling and CLI

### Timeline Example (CLI drops after 30s)

```
0s   - run() starts, CLI connects, polling starts
30s  - CLI drops unexpectedly
30s  - run() returns Err
30s  - polling task cancelled
35s  - AdapterHandle retries (5s backoff, reset because ran >30s)
35s  - run() starts fresh, CLI connects, polling starts
```

The 5s gap is negligible for real-world usage.

muness:
## Final Design: Option C Variation - Adaptive Polling with CLI Enhancement

After discussion, we landed on a hybrid approach that ensures polling always works while CLI is an optional enhancement.

### Architecture

```
┌─────────────────────────────────────────────────────────────┐
│                         run()                               │
├─────────────────────────────────────────────────────────────┤
│  1. Spawn polling (FAST: 2s initially)                      │
│  2. Spawn CLI subscription task                             │
│     ├─ SUCCESS → cli_subscription_active=true               │
│     │            polling switches to SLOW (30s)             │
│     └─ FAILURE → warn, cli_subscription_active=false        │
│                  polling stays FAST (2s)                    │
│  3. Wait for polling to complete (or fail)                  │
│  4. Only polling failure triggers AdapterHandle retry       │
└─────────────────────────────────────────────────────────────┘
```

### State Machine

```
                    ┌──────────────┐
                    │  SLOW + CLI  │ ◄─── CLI connected
        CLI ok      │   (30s)      │      (cli_subscription_active=true)
          ┌─────────┴──────────────┴─────────┐
          │                                   │
    ┌─────▼─────┐                      ┌──────▼──────┐
    │  Starting │                      │  FAST only  │
    │ (FAST 2s) │                      │    (2s)     │
    └─────┬─────┘                      └─────────────┘
          │                                   ▲
          └───────────────────────────────────┘
                    CLI fails (non-fatal)
```

### Why This Design

1. **Polling always works** - Core functionality never depends on CLI
2. **CLI is an enhancement** - Reduces polling load when available, not required
3. **CLI failure is graceful** - Just switches to fast polling, no restart
4. **Single retry trigger** - Only polling failure (LMS unreachable) causes AdapterHandle retry
5. **CLI gets retry on adapter restart** - If polling eventually fails and adapter restarts, CLI gets another chance

### Behavior Examples

**CLI available:**
```
0s  - Polling starts (FAST 2s)
1s  - CLI connects, cli_subscription_active=true
1s  - Polling switches to SLOW (30s)
... - CLI handles real-time events, polling syncs every 30s
```

**CLI unavailable (port blocked):**
```
0s  - Polling starts (FAST 2s)
1s  - CLI connection refused, warn logged
1s  - cli_subscription_active stays false
... - Polling continues at FAST (2s), CLI disabled for this run
```

**CLI disconnects mid-run:**
```
0s  - Both running (CLI active, polling SLOW)
60s - CLI connection drops
60s - cli_subscription_active=false, warn logged
62s - Polling detects change, switches to FAST (2s)
... - Polling handles everything until adapter restart
```

muness:
## Superego Review

### Code Review: LMS Adapter Polling & CLI Changes

This is a solid improvement to the LMS adapter that addresses two distinct concerns:

### 1. **NowPlayingChanged Events from Polling** ✅

The addition of `NowPlayingChanged` event emission from the polling path is well-structured:

**What I like:**
- **Lock discipline is correct**: Collecting updates into a vec and emitting *after* releasing the write lock avoids holding locks while doing I/O (publishing to the bus).
- **Change detection logic is reasonable**: Comparing title, artist, album, artwork_url, and coverid covers the meaningful "now playing" state.
- **Empty title guard**: The `!player.title.is_empty()` check prevents spurious events when nothing is actually playing.
- **New players handled elsewhere**: Correctly delegates to `ZoneDiscovered` for new players.

### 2. **CLI/Polling Task Restructuring** ✅

The refactoring from "CLI drives lifecycle, polling is background" to "Polling drives lifecycle, CLI is optional" is a meaningful architectural improvement.

**What I like:**
- **Clear responsibility**: Polling is the primary mechanism; CLI subscription is an optimization for real-time events.
- **Graceful degradation**: CLI failure is non-fatal—polling continues with fast interval (2s vs 30s).
- **Proper cleanup**: `cli_task.abort()` + `await` ensures no orphaned tasks.
- **Good comments**: The comments explaining the intent are helpful.

### Finding: CLI clean exit doesn't reset polling interval

> **When CLI subscription ends cleanly (not via error), should `cli_subscription_active` be set to `false` so polling switches back to fast mode?**

When CLI ends with `Ok(())`, `cli_subscription_active` isn't explicitly set to `false`. If CLI disconnects gracefully (e.g., server shutdown), polling will stay at the slow 30s interval until the next full adapter restart.

**Recommendation:** Set `cli_subscription_active = false` on clean exit too.

---

*Reviewed by superego*

muness:
Fixed superego finding: `cli_subscription_active` is now always set to `false` when CLI subscription ends (both Ok and Err cases), so polling switches to fast interval regardless of how CLI exited.

muness:
## Addressed CodeRabbit feedback

### 1. Metadata clearing ✅
Fixed: `NowPlayingChanged` is now emitted when metadata clears (e.g., track stops). Empty strings are converted to `None` so clients can distinguish between "no data" vs "empty string".

### 2. CLI retry loop - Intentional design choice
Added comment explaining why CLI doesn't retry independently:
- Polling is the primary mechanism and always works
- CLI is an optimization, not a requirement  
- AdapterHandle already provides retry with exponential backoff
- CLI gets a fresh attempt on adapter restart
- Avoids duplicate retry logic

See PR comments for full design discussion (Options A, B, C).

### 3. Regression test ✅
Added test `polling_emits_now_playing_changed_including_cleared` that:
- Sets track metadata on mock LMS server
- Calls `update_players()` 
- Verifies `NowPlayingChanged` event with track info
- Clears metadata on server
- Verifies `NowPlayingChanged` event with `None` values

PR #166, opened 2026-01-25
refactor(lms): Split CLI subscription into independent adapter (#165)

### LINKED ISSUE #165: refactor(lms): Split CLI subscription into separate adapter with AdapterHandle
## Summary

The LMS adapter currently combines two distinct concerns:
1. **HTTP JSON-RPC polling** - discovers players, fetches status
2. **TCP CLI subscription** - real-time event streaming on port 9090

These are coupled via a `cli_subscription_active` flag that controls polling interval. When CLI fails, it stays dead until the whole adapter restarts. This is a regression from older behavior that had CLI retry logic.

## Problem

From PR #164 discussion and logs:
- CLI failure (timeout, connection closed) leaves the adapter in polling-only mode permanently
- Old code had `"Reconnecting in 5s..."` retry logic that recovered from transient failures
- Current design relies on AdapterHandle retry, but AdapterHandle only restarts on **polling** failure, not CLI failure

## Proposed Solution

Split into two independent adapters, each with its own AdapterHandle:

### LmsPollingAdapter
- Handles HTTP JSON-RPC on port 9000
- Discovers players, polls status
- Reads `cli_subscription_active` flag to adjust poll interval:
  - Fast (2s) when CLI is down
  - Slow (30s) when CLI is up
- Has its own retry via AdapterHandle

### LmsCliAdapter  
- Handles TCP CLI subscription on port 9090
- Subscribes to `playlist,mixer,power,client` events
- Sets `cli_subscription_active = true` on connect, `false` on any failure
- Has its own retry via AdapterHandle
- Reads LMS host from shared state or config

### Shared State
- `Arc<AtomicBool>` for `cli_subscription_active` flag
- Or existing `Arc<RwLock<LmsState>>` with the flag inside
- This is the **only** coupling between the two adapters

## Benefits

1. **Reuses AdapterHandle retry** - no custom retry loops needed
2. **Clear separation of concerns** - each adapter does one thing
3. **Independent failure/recovery** - CLI can fail and retry without affecting polling
4. **Simpler code** - each adapter is smaller and more focused
5. **Better resilience** - CLI reconnects automatically after transient failures

## Implementation Notes

- CLI adapter needs LMS host - can read from config file or wait for polling adapter to discover
- Both adapters spawned independently in `spawn_adapters()`
- Startup order doesn't matter - CLI just retries until LMS is available
- Polling always works, CLI is pure enhancement

## Related

- PR #164: Fixed polling to emit events, documented CLI non-retry as "design choice"
- Discussion in PR #164 led to this architectural insight

## Labels
refactor, lms, adapters

muness:
## Summary

Separates LMS into two adapters with independent AdapterHandle retry:

- **LmsAdapter**: HTTP polling only (port 9000)
- **LmsCliAdapter**: CLI subscription only (port 9090)

Both share `LmsState` and coordinate via `cli_subscription_active` flag:
- CLI connects → flag=true → polling slows to 30s
- CLI fails → flag=false → polling speeds to 2s

## Benefits

- CLI gets independent retry via AdapterHandle
- CLI failure/recovery doesn't affect polling
- Cleaner separation of concerns
- Each adapter simpler and more focused

## Implementation

- Added `LmsCliAdapter` struct implementing `AdapterLogic` + `Startable`
- Factory function `create_lms_adapters()` creates both with shared state
- Removed CLI task spawning from `LmsAdapter::run()`
- Updated `main.rs` to use factory and register both adapters

## Testing

- All existing tests pass
- CLI adapter gets independent retry on failure
- Polling continues at fast interval when CLI is down

Closes #165

## Summary by CodeRabbit

* **Refactor**
  * LMS split into two startable components (polling adapter and CLI subscription adapter) with shared state; CLI adapter added to available adapters.

* **New Features**
  * API and server UI now show CLI subscription status and poll interval; richer LMS player/status data and a players list endpoint.

* **Bug Fixes**
  * Standardized prefixed output IDs across adapters and stricter matching in the aggregator.

* **Tests**
  * New tests enforce prefixed output IDs and volume/event reporting conventions.

* **Documentation**
  * Added Git Workflow guidance to AGENTS.md.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -23,6 +23,13 @@ Use GitHub for all task tracking:
 - Describe what changed and how to test
 - Request review from coderabbit and superego
 
+### Git Workflow
+**DO NOT force push (`git push --force` or `git push -f`)**
+- This project uses squash merges, so commit history cleanup is unnecessary
+- Force pushing breaks checkouts for anyone tracking the branch
+- Force pushing loses SHA references (builds, comments, reviews)
+- Just push new commits - they all get squashed on merge anyway
+
 ---
 
 ## Code Review
--- a/src/adapters/lms.rs
+++ b/src/adapters/lms.rs
@@ -973,8 +973,8 @@ async fn update_players_internal(
 
     // Collect updates to emit after releasing the lock
     let mut now_playing_updates: Vec<NowPlayingUpdate> = Vec::new();
-    // LmsPlayerStateChanged: (player_id, state)
-    let mut state_updates: Vec<(String, String)> = Vec::new();
+    // State updates: (player_id, player_name, state)
+    let mut state_updates: Vec<(String, String, String)> = Vec::new();
     // VolumeChanged: (player_id, volume)
     let mut volume_updates: Vec<(String, i32)> = Vec::new();
 
@@ -1040,7 +1040,11 @@ async fn update_players_internal(
         }
 
         if state_changed {
-            state_updates.push((player.playerid.clone(), player.state.clone()));
+            state_updates.push((
+                player.playerid.clone(),
+                player.name.clone(),
+                player.state.clone(),
+            ));
         }
 
         if volume_changed {
@@ -1067,9 +1071,16 @@ async fn update_players_internal(
         });
     }
 
-    // Emit LmsPlayerStateChanged events for state changes (play/pause/stop)
-    for (player_id, state) in state_updates {
+    // Emit state change events (play/pause/stop)
+    for (player_id, player_name, state) in state_updates {
         debug!("Polling detected state change for {}: {}", player_id, state);
+        // Publish ZoneUpdated so aggregator updates state
+        bus.publish(BusEvent::ZoneUpdated {
+            zone_id: PrefixedZoneId::lms(&player_id),
+            display_name: player_name,
+            state: state.clone(),
+        });
+        // Also publish LmsPlayerStateChanged for SSE/UI
         bus.publish(BusEvent::LmsPlayerStateChanged { player_id, state });
     }
 
@@ -1290,8 +1301,8 @@ async fn handle_cli_event(
                 Ok(status) => {
                     let zone_id = PrefixedZoneId::lms(&player_id);
 
-                    // Update cached state
-                    {
+                    // Update cached state and get player name for ZoneUpdated
+                    let player_name = {
                         let mut s = state.write().await;
                         if let Some(player) = s.players.get_mut(&player_id) {
                             player.state = status.state.clone();
@@ -1304,10 +1315,20 @@ async fn handle_cli_event(
                             player.album = status.album.clone();
                             player.artwork_url = status.artwork_url.clone();
                             player.coverid = status.coverid.clone();
+                            player.name.clone()
+                        } else {
+                            player_id.clone() // Fallback to player_id if not in cache
                         }
-                    }
+                    };
 
-                    // Publish bus events
+                    // Publish ZoneUpdated so aggregator updates state
+                    bus.publish(BusEvent::ZoneUpdated {
+                        zone_id: zone_id.clone(),
+                        display_name: player_name,
+                        state: status.state.clone(),
+                    });
+
+                    // Also publish LmsPlayerStateChanged for SSE/UI
                     bus.publish(BusEvent::LmsPlayerStateChanged {
                         player_id: player_id.clone(),
                         state: status.state.clone(),
--- a/src/adapters/lms.rs
+++ b/src/adapters/lms.rs
@@ -493,6 +493,8 @@ pub struct LmsStatus {
     pub players: Vec<LmsPlayerInfo>,
     /// Whether CLI subscription is active (real-time events vs polling)
     pub cli_subscription_active: bool,
+    /// Effective poll interval in seconds (2s base, 30s when CLI active)
+    pub poll_interval_secs: u64,
 }
 
 /// Summary information about an LMS player for status reporting
@@ -700,6 +702,12 @@ impl LmsAdapter {
     /// Get connection status
     pub async fn get_status(&self) -> LmsStatus {
         let state = self.state.read().await;
+        let base_interval = get_poll_interval();
+        let effective_interval = if state.cli_subscription_active {
+            get_poll_interval_with_subscription()
+        } else {
+            base_interval
+        };
         LmsStatus {
             connected: state.connected,
             host: state.host.clone(),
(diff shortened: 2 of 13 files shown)


muness:
## Code Review: LMS Adapter Split (#165)

Overall this is a clean refactor that achieves the goal of giving CLI subscription independent retry via AdapterHandle. The design is sound and the implementation is straightforward.

### What Works Well

1. **Clean separation** - The split is surgical. `LmsAdapter` now focuses purely on polling, while `LmsCliAdapter` handles CLI subscription. The shared state coordination via `cli_subscription_active` flag is unchanged.

2. **Factory function** - `create_lms_adapters()` is a nice pattern that ensures both adapters share the same state and RPC client. This prevents accidental misconfiguration.

3. **Consistent patterns** - `LmsCliAdapter` follows the same `AdapterLogic` + `Startable` patterns as other adapters, making it easy to understand.

4. **Documentation updated** - Module docs now accurately reflect the two-adapter architecture.

### Suggestions

**1. Consider exposing `state` and `rpc` via accessor methods instead of `pub(crate)`**

```rust
// Current
pub(crate) state: Arc<RwLock<LmsState>>,
pub(crate) rpc: LmsRpc,

// Alternative - more encapsulated
impl LmsAdapter {
    pub(crate) fn shared_state(&self) -> Arc<RwLock<LmsState>> {
        self.state.clone()
    }
    pub(crate) fn shared_rpc(&self) -> LmsRpc {
        self.rpc.clone()
    }
}
```

This is minor - current approach works fine for a factory function in the same file.

**2. `handle_command` returns a "command not supported" response**

The `LmsCliAdapter::handle_command` returns `success: false`. This is correct behavior, but worth confirming the coordinator won't route commands to `lms-cli` prefix. Since both adapters share the `"lms"` prefix, commands will likely be dispatched to both. Does the coordinator handle this gracefully when one returns `success: false`?

If command routing is by first match or similar, this should be fine. Just want to flag in case there's duplicate command handling.

**3. Minor: Logging clarity**

```rust
info!("LMS CLI adapter starting for {}", host);
```

Consider distinguishing this more clearly from the main adapter's logs during debugging:

```rust
info!("[CLI] LMS CLI adapter starting for {}", host);
```

Or using a span/target. Not critical.

### Questions

1. **Shutdown coordination**: Both adapters have independent shutdown tokens. Is there a scenario where stopping the main `LmsAdapter` should also stop the CLI adapter? Currently they're fully independent, which seems intentional for the retry use case.

2. **Configuration changes at runtime**: If `lms.configure()` is called while adapters are running (host change), does `LmsCliAdapter` pick up the new host on its next retry? Looks like yes, since it reads `state.host` in `run()` - just confirming this is the intended behavior.

### Verdict

This is a solid refactor. The split achieves the goal cleanly and follows established patterns. The only item worth double-checking is the command routing behavior with shared prefixes.

Nice work!

muness:
## Review: Issue #165 - Split LMS Adapter into Polling + CLI

This is a well-executed refactor that addresses a real architectural problem. Good work.

### What's Working Well

**1. Clear Problem → Solution Path**
The issue is well-understood: previously CLI and polling were entangled in one adapter, meaning CLI failures could interfere with polling retry logic. The split gives each their own `AdapterHandle` with independent retry—polling continues reliably while CLI retries independently.

**2. Minimal Coordination via Shared Flag**
The `cli_subscription_active` flag is elegant—it's the only shared state between the two adapters. Polling reads it to adjust interval (fast when CLI down, slow when CLI up). No complex synchronization needed.

**3. Factory Function Pattern**
`create_lms_adapters()` ensures the shared state is set up correctly and makes the coupling explicit. Users can't accidentally create an `LmsCliAdapter` without the shared state.

**4. Architecture Lint Tests**
The new `state_changes_emit_zone_updated` test is excellent—it codifies the lesson learned (that adapter-specific events don't update the aggregator) so it can't regress. The documented schema in `bus_event_schema_documented` serves as living documentation.

**5. ZoneUpdated Emission**
The fix to emit `ZoneUpdated` alongside `LmsPlayerStateChanged` solves the real bug: CLI events were updating internal cache but not propagating to the aggregator.

### Observations (Not Blockers)

**1. Dual Event Emission Pattern**
You're now emitting both `ZoneUpdated` (for aggregator) and `LmsPlayerStateChanged` (for SSE/UI) in multiple places. The comment "Also publish LmsPlayerStateChanged for SSE/UI" is good, but consider whether `LmsPlayerStateChanged` is still needed at all. If the SSE layer subscribes to `ZoneUpdated`, you could remove the adapter-specific event entirely. Worth a follow-up ticket to evaluate.

**2. Lint Test Heuristic May Need Tuning**
The new lint in `architecture_lint.rs` looks for `zone_id = PrefixedZoneId::` in the preceding 2000 chars. This could have false negatives if the variable is named differently (e.g., `prefixed_id`) or passed through a function. Not a problem today, but worth noting for future maintenance.

**3. CLI Adapter Command Handling**
The CLI adapter returns `success: false` for all commands. This is correct (polling adapter handles commands), but the coordinator routes commands by prefix ("lms"). With two adapters sharing the same prefix, does command routing still work correctly? Worth verifying—looks like LmsAdapter's `handle_command` would be the one registered, so this should be fine.

**4. Poll Interval Display Logic Duplication**
The API handler (`lms_config_handler`) calculates `effective_poll_interval` using hardcoded multiplier (`poll_interval_secs * 15`). This duplicates knowledge from wherever the actual polling interval is calculated. Consider extracting this to a shared constant or function.

**5. AGENTS.md Git Workflow Guidance**
Good addition. The "squash merge" context makes the "no force push" rule understandable rather than arbitrary.

### Summary

This is a clean refactor that:
- Solves a real problem (independent retry for CLI vs polling)
- Adds guardrails to prevent regression (lint tests)
- Maintains the system's existing behavior while improving resilience

The architecture is sound—shared state is minimal, each adapter is independently restartable, and the failure modes are well-handled (CLI down → polling speeds up; polling down → adapter restart).

---
*Review generated via `sg review pr`*

muness:
## Volume Fix Added (94f86f9)

Fixed a bug where LMS CLI relative volume changes were being treated as absolute values.

**Problem:** When adjusting volume via iPhone/remote controls, LMS CLI sends relative changes like `-3` (decrease by 3%). The code was treating this as an absolute volume of -3, causing nonsensical displays like "-3 state".

**Solution:**
- Added `is_relative` flag to detect `+`/`-` prefixed values
- Changed mixer value from `i32` to `f32` to support LMS's 2.5% step size
- Applied relative changes to cached volume instead of using as absolute
- Clamped results to 0-100 range

New tests added for absolute, relative positive (+2.5), and relative negative (-3) volume events.

PR #168, opened 2026-01-25
fix(roon): Prevent volume snap to max when API returns None

muness:
## Summary
- Fixes bug where Roon dB-based outputs (like Poly Mojo 2) would snap to max volume when API returns `None` transiently
- Now only emits `VolumeChanged` when we have a valid value
- Defaults to min volume (safest) for zone discovery when value is missing

## Problem
The Roon API's `Volume.value` is `Option<f32>` and can be `None` in transient states. Previously:
- `unwrap_or(0.0)` was used
- For dB zones (where 0 dB = max volume), this caused a dangerous jump to maximum volume

## Test plan
- [x] All existing tests pass
- [x] Manual test with dB-based output (if available)
- [x] Verify volume events are not emitted when value is None

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented emitting invalid or transient volume values during zone updates; safer handling for dB-based zones.
  * Volume defaults now derive from the actual range and use an inferred safe scale/minimum.

* **Diagnostics**
  * Added per-output debug logging for volume details during zone conversions.

* **Tests**
  * Added a lint enforcing safe volume-default handling (note: lint block was accidentally duplicated).

### DIFF
--- a/src/adapters/roon.rs
+++ b/src/adapters/roon.rs
@@ -537,6 +537,18 @@ fn convert_zone(roon_zone: &RoonZone) -> Zone {
         length: np.length,
     });
 
+    // Log output volume info for debugging
+    for o in &roon_zone.outputs {
+        tracing::debug!(
+            "Zone '{}' output '{}': volume={:?}",
+            roon_zone.display_name,
+            o.display_name,
+            o.volume
+                .as_ref()
+                .map(|v| format!("value={:?} min={:?} max={:?}", v.value, v.min, v.max))
+        );
+    }
+
     let outputs = roon_zone
         .outputs
         .iter()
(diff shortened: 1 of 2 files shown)


muness:
## Superego Review

### 1. Is this the right fix for the safety issue?

**Yes, this is a correct and safe fix.** The changes address both places where `vol.value` is consumed:

**Zone discovery:** Defaulting to `min` instead of `50.0` is the right call. For dB zones where the range might be -64 to 0, defaulting to min (-64 dB) is the quietest/safest. The old `50.0` default could have been out of range entirely for some zones.

**VolumeChanged emission:** Silently skipping the event when `value` is `None` is appropriate. This avoids propagating bad data downstream.

### 2. Should we add a regression test?

**Yes, a unit test would be valuable.** The `roon_zone_to_bus_zone` function is pure and testable. A test could verify:

```rust
#[test]
fn zone_discovery_defaults_to_min_when_value_none() {
    // Create a zone with volume.value = None, min = -64.0
    // Assert the BusVolumeControl.value equals -64.0, not 50.0 or 0.0
}
```

### 3. Edge cases and concerns

1. **Consistency check:** Change detection compares `ov.value != vol.value` where both are `Option<f32>`. If old was `Some(x)` and new is `None`, this returns `true` (changed), but then the event is suppressed. That's fine - no downstream impact.

2. **is_muted still uses unwrap_or(false):** Safe default, no action needed.

3. **The comment is helpful:** The inline comment explains *why* this matters (dB zones, 0 = max). Future maintainers will thank you.

### Verdict

**Approved.** The fix is sound. A unit test for `roon_zone_to_bus_zone` with `value: None` would be a good addition.

muness:
## Addressed CodeRabbit Feedback

1. **Inconsistent volume range defaults** (Major): Fixed by using `get_volume_range()` for consistent defaults between `roon_zone_to_bus_zone` and `change_volume`. Also now infers the scale from the range (dB if max <= 0, else Percentage).

2. **Mute-only updates dropped** (Minor): Fixed by using last known value when current value is None but mute state changed. Code now does:
   ```rust
   let value_to_use = vol.value.or_else(|| old_vol.and_then(|ov| ov.value));
   ```

Both issues are now addressed in the latest push.

muness:
@coderabbitai review

muness:
## Added Unit Tests for CodeRabbit

Added regression tests for `roon_zone_to_bus_zone` in `src/adapters/roon.rs`:

1. **`roon_zone_to_bus_zone_db_scale_value_none_defaults_to_min`**: Tests dB zone (max <= 0) with `value=None` defaults to min, scale is `Decibel`
2. **`roon_zone_to_bus_zone_percent_scale_value_none_defaults_to_min`**: Tests percentage zone (max > 0) with `value=None` defaults to min, scale is `Percentage`
3. **`roon_zone_to_bus_zone_preserves_actual_value`**: Tests that actual values are preserved when present
4. **`roon_zone_to_bus_zone_no_volume_returns_none`**: Tests that outputs without volume return `volume_control = None`

All tests verify:
- `value` defaults to `min` when `None`
- `scale` is inferred from range (dB if max <= 0, Percentage otherwise)
- `step` defaults to 1.0
- `is_muted` defaults to false
- `output_id` is prefixed with "roon:"

Issue #341, opened 2026-07-30
HQPlayer: reconcile protocol evidence and retire contradictory assumptions


muness:
OH: 80222d6d

Parent epic: #311

Related: #322

## Problem Statement

**Current framing:** Keep the existing HQPlayer protocol notes and correct them when implementation work exposes an error.

**Reframed as:** Maintainers need one evidence-ranked protocol reference that distinguishes observed wire behavior, official client behavior, version-specific findings, and unresolved contradictions, but UHC currently has overlapping documents that disagree on command domains and omit important negative findings.

**The shift:** From prose treated as specification to a provenance-tagged evidence ledger whose disputed claims become executable conformance cases.

### Constraints

- **Hard:** Executable fixtures outrank prose; no reverse engineering of proprietary HQPlayer internals; empirical claims name edition/version/date/playback state; unresolved contradictions remain explicit; official and third-party licenses are preserved.
- **Soft:** Whether the ledger is one document or generated from fixture metadata.

### What this framing enables

Reliable future maintenance, version-specific support claims, faster incident diagnosis, and fewer repeated dead-end investigations.

### What this framing excludes

Treating one hqplayerd 6.0.4 session as universal, copying manual prose, or resolving uncertainty by hardcoding current observations.

## Acceptance criteria

- [ ] Reconcile `.oh/hqplayer-spec.md` and `docs/hqplayer-protocol-reference.md`, including the stale `SetMode` value-vs-index contradiction.
- [ ] Record the three numeric domains separately: native list index, persistent enum ID, and semantic name, including mode-relative behavior.
- [ ] Record negative findings so they are not retried: partial `POST /config` can return HTTP 200 without applying, daemon profile save/load routes are unsafe for a durable preset store, and self-generated 4321 session-authentication keys were rejected in the inspected precedent.
- [ ] Track unresolved `State.active_mode` versus `Status.active_mode` behavior as a versioned conformance question rather than choosing one source globally.
- [ ] Track `SetRate` semantics explicitly: index on the wire, exact runtime Hz pin, Auto behavior, mode/filter/device-dependent lists, and mode switches clearing the pin, with upstream observations clearly marked pending UHC hardware confirmation.
- [ ] Every empirical claim names source repository/ref, HQPlayer edition/version, capture date, and whether playback was active.
- [ ] Link each behavior that affects code to a #322 fixture or a #332 live qualification row.
[...]

Issue #332, opened 2026-07-29
HQPlayer adaptive control: validate the end-to-end compatibility and recovery matrix


muness:
OH: 80222d6d

Parent epic: #313

## Problem Statement

**Current framing:** Ask Michael to test the knob after implementation.

**Reframed as:** Users need confidence that the complete producer-to-surface loop works on real HQPlayer editions, versions, controllers, and failure modes, but unit mocks alone cannot validate undocumented server behavior, restarts, external controllers, firmware limits, or LAN timing.

**The shift:** From informal final testing to an explicit compatibility and recovery qualification gate for the program.

### Constraints
- **Hard:** Do not claim support without evidence; retain CI fixtures for every discovered behavior; test direct and linked roles; destructive persistent tests require backup and controlled environment; no merge solely because CI passes.
- **Soft:** Exact supported version breadth and hardware/device combinations.

### What this framing enables
An honest support matrix, reproducible regression fixtures, release notes grounded in evidence, and a clear manual approval gate.

### What this framing excludes
A vague "works on my machine" claim, undocumented skips, and using production listening configuration as an unrecoverable test target.

## Acceptance criteria

- [ ] Define the matrix across HQPlayer Desktop/Embedded, available v5/v6 versions, direct and linked-DSP roles, single/multi-instance, and supported UHC surfaces.
- [ ] Run protocol discovery, now-playing, transport, seek, decimal-dB volume, live tuning, external-controller convergence, disconnect/restart recovery, and unsupported-capability cases.
- [ ] Embedded rows add backup/restore/restart/readback/rollback testing.
- [ ] ESP32-S3 validation covers zone persistence, metadata, transport, volume, adaptive manifest refresh, and reconnect without a relay PC.
- [ ] Web and MCP validate broader settings from the same producer revisions and report identical terminal outcomes.
- [ ] Every real-server discrepancy becomes a minimized CI fixture before correction.
- [ ] Publish the supported/unsupported matrix, known limitations, recovery instructions, and evidence date.
- [ ] Obtain explicit user approval after hands-on testing before merging/shipping the program.

## Dependencies

Final qualification is blocked by the lifecycle, producer contract, full controls, persistence, surface adoption, and #208 multi-instance work that is claimed for the release.

## HQPTuner evidence rows added to qualification (2026-07-30)

[...]

Issue #329, opened 2026-07-29
HQPlayer: apply live modes and settings with verified pending/error semantics


muness:
OH: 80222d6d

Parent epic: #313

## Problem Statement

**Current framing:** Expose setters for mode, rate, filters, shaper, dither, adaptive output, and other pipeline fields.

**Reframed as:** HQPlayer users need broad live tuning they can trust, but a setter response does not prove the engine applied the requested value, choices can change with mode, and concurrent/external changes can make optimistic client state wrong.

**The shift:** From fire-and-forget setters to serialized, observed state transitions with explicit pending, verified, rejected, superseded, timed-out, and unavailable outcomes.

### Constraints
- **Hard:** Semantic client values; native index/ID conversion stays in the adapter; state readback determines success; constraints are refreshed when dependent state changes; safe ordering; aggregator publishes pending/final truth.
- **Soft:** Batch/staging behavior, retry count, and per-setting timeout.

### What this framing enables
Trustworthy tuning from web/MCP/devices, accurate Applying UX, safe data-driven controls, and meaningful automation.

### What this framing excludes
Immediate optimistic success, cached enum lists surviving mode changes unchecked, uncontrolled simultaneous setters, and advertising unverified commands.

## Acceptance criteria

- [ ] Cover every reliably supported live setting discovered from HQPlayer, including mode, rate, 1x/Nx filters, shaper/dither, adaptive output, and additional tunables established by conformance fixtures.
- [ ] Validate requested semantic value against the latest producer document before sending.
- [ ] Serialize conflicting commands and assign correlation/revision identity.
- [ ] Publish pending state without replacing authoritative current value.
- [ ] Read back until applied, rejected, superseded, disconnected, or timed out.
- [ ] Refresh dependent option lists/constraints after relevant changes.
- [ ] Expose machine-readable outcome and user-facing reason.
- [ ] Staging/batch behavior, if supported, reports live versus restart-required counts.
- [ ] Tests cover delayed apply, explicit error, no-op OK, external overwrite, disconnect mid-apply, and stale choice.

## Dependencies

Blocked by protocol conformance and the HQPlayer declarative producer.
Blocks #222's applying-state UX.
Related: #209.

## HQPTuner beta/dev salvage additions (2026-07-30)

- [ ] A setter's syntactic OK is provisional; compare readback to the exact requested semantic outcome and report rejected/no-op/timeout/superseded separately.
[...]

Issue #325, opened 2026-07-29
HQPlayer: produce declarative controls from discovered state and constraints


muness:
OH: 80222d6d

Parent epic: #312

## Problem Statement

**Current framing:** Return HQPlayer mode/filter/shaper/rate lists to clients.

**Reframed as:** Consumers need a truthful, semantic description of what this HQPlayer instance can do now, but current methods expose partial lists and backend-specific conversion rules without dynamic constraints, availability reasons, apply behavior, or freshness.

**The shift:** From exposing native option lists to making HQPlayer the first complete producer of the generic adaptive-control document.

### Constraints
- **Hard:** Discover from the connected engine; semantic names are client-facing while indices/IDs remain adapter-private; Desktop/Embedded and v5/v6 capability differences are honest; external changes republish state; disabled choices include reasons; no control is advertised without a working execution path.
- **Soft:** Initial advanced-setting breadth and grouping.

### What this framing enables
Generic UI rendering, MCP option discovery, matcher-selected layouts, version-aware behavior, and later producers following a proven contract.

### What this framing excludes
Hardcoded universal HQPlayer enum tables, stale enum caches across mode changes, and advertising persistent controls on unsupported editions.

## Acceptance criteria

- [ ] Produce direct playback, volume, pipeline, advanced setting, profile, and persistence capability groups where supported.
- [ ] Map current values and choices using stable semantic IDs/names, preserving adapter-private native IDs/indexes only for execution.
- [ ] Republish when state, enum choices, constraints, connection, edition, or version capability changes.
- [ ] Describe min/max/step/unit for decimal-dB volume and every discoverable numeric control.
- [ ] Represent mode-dependent filtering/narrowing and unavailable choices with reasons.
- [ ] Distinguish live-only, verified-live, restart-required, and persistent controls.
- [ ] Omit or mark controls unsupported when their command path is not reliable.
- [ ] Snapshot tests cover Desktop/Embedded and representative v5/v6 fixture capabilities.
- [ ] Audit and record provenance for labels, descriptions, constraints, and compatibility metadata: distinguish live-protocol discoveries, official hqp-control evidence, MIT-licensed HQPTuner code, and manual-derived prose/data; do not copy material whose reuse rights are unclear.
- [ ] A generic test consumer can render and invoke every advertised primitive without HQPlayer-specific knowledge.

## Dependencies

[...]

PR #366, opened 2026-07-31
fix(347): verify live setters and scope enumerations to the loaded chain

### LINKED ISSUE #347: HQPlayer: verify live setters and scope enumerations to the loaded chain
OH: 80222d6d

Parent epic: #311

Related: #322 and PR #337

## Problem Statement

**Current framing:** Once a native setter returns result=OK and the adapter refreshes its cached lists after explicit mode changes, the requested setting can be treated as applied.

**Reframed as:** HQPlayer users need live settings to mean exactly what the UI says, but the daemon can acknowledge a setter without applying it, the loaded PCM/SDM chain can change while configured mode remains [source], and one-sided SetFilter calls can overwrite the sibling setting.

**The shift:** From response-based success and mode-scoped caching to readback-verified writes and loaded-chain-scoped enumerations.

### Constraints

- **Hard:** A setter response is not proof of application; verify the authoritative State field. Enumerations are scoped to the loaded chain and invalidated when that chain changes. Runtime list indices are discovered, never accepted as raw user identity. SetFilter always carries both arguments or refuses safely. No public API changes without explicit approval.
- **Soft:** Whether per-family rate intent is held by the adapter in this issue or by the later producer/aggregator contract.

### What this framing enables

Truthful setting outcomes, safe source-following playback, correct filter/shaper selection across chain changes, and a dependable foundation for #325 and #329.

Blind retries after result=OK, guessing a sibling filter from another State field, hardcoded mode positions, numeric-string fallback to native indices, and silently pinning a rate while [source] is configured.

## Acceptance criteria

- [ ] Start from #322 stateful fixtures and add a bite check showing each regression fails against the pre-fix adapter.
- [ ] Every supported live setter checks result semantics and authoritative State readback before reporting applied.
- [ ] Accepted-but-unchanged is a distinct ignored outcome, not success; timeout and retry behavior is bounded and observable.
- [ ] Active chain is resolved independently from configured mode and rate-pin availability; a source-driven PCM/SDM chain transition invalidates and refreshes filter, shaper, and rate lists.
- [ ] SetRate is suppressed or held with an explicit reason while [source] is configured, even when the daemon would answer OK.
- [ ] A mode write is skipped when that mode is already running and, when performed, the cleared rate pin is reflected honestly.
- [ ] SetFilter always sends authoritative value and value1x; if either side cannot be established, the write is refused without mutation.
- [ ] Semantic names resolve only through the current enumeration. Raw integer-string fallback is removed from production control paths.
- [ ] Device-dependent mode lists are handled by semantic prefix/alias matching, including SDM (DSD), without assuming positions.
- [ ] Accumulated native responses have an explicit size limit and oversized/malformed responses fail without wedging later polling.
- [ ] Tests cover external chain changes, silent setter refusal, absent sibling state, no-op mode writes, and device mode omission.
- [ ] No public API route, request, or response contract changes.

## Evidence

HQPTuner refs: main 67557939, beta 6e318324, dev 22dfe5cc. The beta/dev behavior is upstream evidence pending UHC hardware qualification in #332.

## Dependencies

Blocked by #322 as the executable protocol foundation and #341 as the reconciled evidence/authority ledger. This issue blocks the HQPlayer producer and verified live-setting surfaces in #325 and #329.

## Completion signal

A client can issue a semantic setting change during source-following playback and UHC either proves the exact setting applied or returns a bounded, truthful non-applied outcome without corrupting a sibling setting.

## Advertised matrix-profile correctness (direct audit, 2026-07-30)

UHC already advertises matrix-profile selection through legacy and instance-specific HTTP routes, so this is shipped correctness—not optional #346 breadth.

- [ ] Add an executable fixture for `MatrixListProfiles`, `MatrixSetProfile`, and authoritative `State.matrix_profile`, including the empty/default profile identity, setter error, accepted-but-unchanged, and external profile switch.
- [ ] Resolve the requested profile against the fresh native list, send its semantic name, validate the setter result, and read back `State.matrix_profile` before reporting applied.
- [ ] Do not use `MatrixGetProfile` as the current-profile authority unless #341 records versioned evidence that the supported daemon implements it consistently; the observed State field is the default authority.
- [ ] Preserve the existing numeric legacy HTTP request contract unless an API change is explicitly approved, but keep that numeric value at the compatibility boundary and never persist or publish it as adaptive-control identity.
- [ ] Invalidate/refresh the profile list and active selection after external changes and reconnect; a stale list index can never select a different semantic profile silently.
- [ ] #346 owns later profile definition/editing/persistence. This issue owns the truthfulness of the currently exposed live selector.

muness:
Fixes #347

OH: 80222d6d · epic #311 · **stacked on #341**

| | |
|---|---|
| **Base** | `feat/issue-341-hqplayer-evidence-ledger` @ `6d688bdbcb3fa607ca87a1593dd692e7ba00133a` |
| **Head** | `8970fd9cb93d537b6e7e234aef55c6e289a323bb` |
| **Merge order** | #322 → #341 → **#347**. Do not merge before #341 |

---

## Problem

Three things were treated as proof that were not.

1. **`result="OK"` was treated as application.** The daemon can acknowledge a setter and apply
   nothing (HQP-C-028). Worse, the readback #322 added cannot catch the narrow case: requesting Auto
   under `[source]`, where the rate is already Auto and the daemon ignores the pin, compares 0 against
   0 and reports success for a command that did nothing (HQP-C-019).
2. **A mode-keyed list cache was treated as the loaded chain.** Under configured `[source]` the source
   decides which PCM/SDM chain is loaded and `State.mode` never moves (HQP-C-007), so nothing prompted
   a re-read. Observed pre-fix, hermetically, on #347 comment 5125934210: `ok=true` returned while the
   daemon ended on `poly-sinc-lp` — a filter nobody asked for.
3. **A number arriving where a name belongs was treated as a list position.** `resolve_mode_index`,
   `resolve_filter_index` and `resolve_shaper_index` each fell back to `parse::<u32>()` (HQP-C-063).

Plus: one-sided `SetFilter` guessed the absent sibling from the legacy combined `filter` field, which
silently overwrites the setting the caller did not touch (#322 audit, defect C5).

## Solution

**1 · A setter reports what it did.** `SettingOutcome` — `Applied` / `AlreadySet` / `Ignored` /
`Suppressed` / `Ambiguous`, `#[must_use]` — replaces `Result<()>` on the enumerated live setters.
`into_applied_result()` collapses it to today's `{"ok":true}` / `{"error":…}` at the HTTP and MCP
boundaries, so **no response contract moves** and the *compiler*, not a reviewer, is what forces both
surfaces to handle it. `HqpRejected` makes an explicit `result="Error"` a type rather than a message,
because a rejection is terminal while a lost reply is **ambiguous delivery** (HQP-C-029) — a `SetMode`
was once accepted, logged and acted on while the daemon answered nothing — and telling those apart by
matching error text is how the distinction quietly stops holding. A write whose reply is lost is now
read back: if the daemon has it, it applied.

**2 · The loaded chain is the enumeration the daemon serves.** Not the configured mode, and not
`State.active_mode`, whose semantics under `[source]` are *unmeasured* (HQP-C-024) — reading it as an
answer would settle by fiat exactly what #341 refuses to settle. Per-family fingerprints: a freshly
fetched list whose fingerprint differs from the cached one **is** a chain transition, and every
chain-scoped family dies with it. Control paths resolve names against a list fetched now; the read
path probes with `GetRates`, the smallest chain-scoped enumeration, whose two chains were observed
disjoint (HQP-C-020). Connect, disconnect and `mark_disconnected` all invalidate.

**3 · Numbers stop at the boundary.** `legacy_index_to_name` resolves the frozen numeric request
contracts against the current enumeration and passes the **name** inward. A position the current list
does not have is an error there rather than an index forwarded to the wire.

And the consequences the issue names: `SetRate` **suppressed** under `[source]` with the mode quoted
rather than sent and failed on readback arithmetic — the only thing that fixes the Auto case;
`SetMode` **skipped** when the daemon is already in that mode, because it clears the rate pin even
when the mode does not change (HQP-C-017); `SetFilter` carrying **both** sides or refusing without
mutating; modes resolved by semantic family so `"DSD"` reaches the daemon's `"SDM (DSD)"` with no
positional assumption (HQP-C-013).

## RED → GREEN evidence

Every expectation was written first and run against the pre-fix adapter. Thirteen behavioural reds,
commit `4fc93b5`, before any production change:

| Expectation | Failure at `6d688bd` |
|---|---|
| `a_rate_pin_under_source_is_suppressed_before_it_reaches_the_daemon` | `SetRate` count `1`, want `0` |
| `an_auto_rate_request_under_source_is_never_reported_as_applied` | sent, and `Ok` returned |
| `a_no_op_mode_write_is_not_sent_so_the_rate_pin_survives` | `SetMode` count `1`, want `0` |
| `a_mode_alias_resolves_through_the_daemons_own_name` | `Mode 'DSD' not found. Available: [source], PCM, SDM (DSD)` |
| `a_source_driven_chain_change_invalidates_the_cached_filter_list` | stale index 7 accepted, daemon on `SYN-sdm-7` |
| `after_a_chain_change_a_filter_resolves_against_the_newly_loaded_list` | `HQPlayer rejected SetFilter: invalid filter` |
| `a_name_present_in_both_chains_is_sent_with_the_loaded_chains_index` | same |
| `a_chain_change_invalidates_the_shaper_and_rate_lists_the_pipeline_publishes` | PCM shapers still served over a loaded SDM chain |
| `a_reconnect_invalidates_the_cached_enumerations` | `SYN-pcm-*` after reconnect |
| `a_filter_write_is_refused_without_mutation_when_the_sibling_is_unknowable` | sibling guessed and sent |
| `a_numeric_string_is_never_accepted_as_a_setting_name` | `"1"` taken as index 1 |
| `the_current_matrix_profile_is_read_from_state_not_matrix_get_profile` | `MatrixGetProfile` believed over `State` |
| `a_matrix_profile_accepted_but_unchanged_is_not_reported_as_applied` | `OK` reported as applied |
| `a_current_matrix_profile_absent_from_the_fresh_list_...` | published as `MatrixProfile { index: 0, name: "Ghost" }` — index 0 is `Default` |

Further reds were demonstrated by restoring the pre-fix behaviour in place and re-running:

| Expectation | Failure with the pre-fix mechanism restored |
|---|---|
| `a_write_whose_reply_is_lost_after_the_daemon_applied_it_is_reported_as_applied` | `Connection closed mid-document after 1 bytes` |
| `a_write_whose_delivery_cannot_be_established_is_reported_as_ambiguous` | `Response timeout` |
| `a_chain_change_during_a_list_refresh_publishes_nothing_rather_than_a_mixed_set` | `filters=[… "sinc-Lh" …] shapers=[… "NS9" …]` — SDM filters offered beside PCM shapers |
| `a_single_failed_family_publishes_no_part_of_the_refresh` | `Published anyway: ["samplerate"]` |
| `the_legacy_both_sides_filter_write_is_a_single_command` | 2 requests, want 1 |
| `a_both_sides_filter_write_that_is_ignored_reports_both_sides` | `accepted filter1x but filter1x still reads 6` — naming half the write |
| `the_legacy_setting_route_still_refuses_a_name_it_never_accepted` | `{"error":"Connection timeout"}` — `dither` was accepted and reached the adapter |
| `a_chain_change_between_the_state_read_and_the_list_read_is_not_published` | published `""` while the daemon held `poly-sinc-xtr` |
| `a_raw_filter_write_the_daemon_acknowledges_and_drops_is_not_reported_as_success` | `Ok(())` for a write the daemon acknowledged and dropped |
| `a_chain_that_moves_on_both_the_read_and_the_retry_is_reported_rather_than_published` | published `[]` families, `filter1x` selected as `""` out of 0 options, as a successful read |
| `a_single_failed_family_fails_the_read_rather_than_publishing_an_empty_one` | `Ok` with four empty lists, reported as a successful read |
| `a_first_read_whose_required_fill_never_settles_is_not_reported_as_success` | same, reached through the first read after connect |

The static lint's bite was verified the same way: reinstating the fallback in `resolve_shaper` yields
`Offenders: ["resolve_shaper"]`.

### Exact-head review corrections

Four later exact-head CodeRabbit findings plus two independent dissent rounds produced seventeen
additional RED expectations before this head. Each behavioral or structural regression was run
against the prior production shape and failed before its correction:

| Expectation | Failure before the correction |
|---|---|
| `a_profile_load_whose_refresh_fails_never_reuses_pre_profile_enumerations` | after a successful profile POST followed by protocol-level `Error` replies for `GetShapers`, the next read confidently returned the old list containing `NS9` |
| `a_reconnect_during_the_reads_never_publishes_empty_enumerations_as_current` | a reconnect between list and state reads returned all four option families empty and selected `filter1x` as `""` in a successful response |
| `a_final_chain_check_that_cannot_run_twice_is_reported_rather_than_published` | two rejected final `GetRates` checks still published all four families through an unverified cache |
| `a_final_chain_check_that_runs_on_the_retry_publishes_the_verified_read` | one rejected final check was swallowed and the read published without any successful verification |
| `a_lost_volume_range_reply_does_not_publish_the_cache_its_reconnect_emptied` | a post-proof reconnect emptied the cache and the read still returned `Ok` with 0/4 option families |
| `the_pipeline_projection_never_re_reads_the_cache_it_projects` | the AST lint found three direct state-lock acquisitions after/beside proof, leaving a check-then-re-read replacement window |
| `an_old_probe_never_clears_a_newer_complete_cache_with_different_rates` | an old probe erased a newer complete cache instead of yielding as `Replaced` |
| `a_gather_overtaken_by_a_newer_profile_publishes_nothing_and_leaves_the_winner` | a stale off-lock full gather overwrote a newer complete profile cache |
| `an_invalidation_with_no_refill_stops_an_overtaken_gather_as_well` | a gather issued before a reconnect could refill the cache after the reconnect event |
| `a_byte_identical_refill_is_still_a_replacement_for_a_read_that_captured_the_old_cache` | four equal fingerprints made a clear/refill look untouched and joined pre-event lists to post-event `State` |
| `the_generation_counter_never_stalls_at_the_integer_ceiling` | a saturating counter stayed at `u64::MAX`, making every later cache event invisible |
| `a_stale_filter_fetch_overtaken_by_a_complete_winner_is_refused_and_leaves_it_alone` | a stale single-family writer produced a complete mixed cache: three families from the winner and old filters |
| `every_family_refuses_a_fetch_the_cache_has_outrun` | each `fresh_*` path could publish a reply from a session/cache generation that had ended |
| `a_fresh_family_identical_to_the_cached_one_is_not_a_new_filling_of_the_cache` | ordinary identical setter refetches advanced the generation and manufactured retry/failure races |
| `the_pipeline_read_re_reads_the_cache_after_any_refresh_attempt` | `refresh_lists=false` discarded a newer complete winner when this refresh lost the publication race |
| `an_overtaken_fresh_family_that_resolves_like_the_winner_is_still_usable` | the first generation guard returned a 500 after an empty-cache full refresh won with the exact same semantic mapping |
| `a_full_snapshot_fill_fences_an_older_fresh_family_with_a_different_mapping` / `changed_family_metadata_is_published_even_when_the_mapping_fingerprint_is_unchanged` | mutation checks showed the full-snapshot bump and whole-family no-op comparison were not independently pinned |

CodeRabbit then found a separate setter race at exact head `412f20c`: a semantic name was resolved
against one chain, but `AlreadySet`/write/readback compared only the numeric position after the chain
moved. Eleven additional semantic-bracket expectations plus six dissent corrections pin the result:

| Setter proof | Failure before the correction |
|---|---|
| stale filter/shaper no-op | returned `AlreadySet` for a different semantic value at the same position |
| cross-chain filter/shaper write | returned `Applied` after the position selected a different name |
| retry after an earlier write | a later resolve/fetch/decision error erased the wire attempt, or a retry no-op returned `AlreadySet` |
| paired filter readback | split and missing fields both collapsed to ambiguity instead of observed `Ignored` vs unknown `Ambiguous` |
| acknowledged negative readback | leaked numeric positions instead of filter/mode/shaper names or rate Hz (`0` is now `Auto`) |
| rate retry entering `[source]` | source suppression was checked only once, outside the retry decision |
| transition to `[source]` between decision and `SetRate` | Auto was acknowledged/ignored, numeric zero and an unchanged rate list falsely yielded `Applied`; final proof now resolves configured mode and returns wire-aware `Ambiguous` |
| transition after proof `State` before modes fetch | `GetModes` is capability-only; stale `State` still yielded `Applied`; `State` is now re-read after modes and used for source/value proof |
| exhausted bracket | ambiguity did not disclose whether any write reached or may have reached the wire |

`SemanticFamily` now brackets each semantic setter with opening and closing enumerations, resolves and
compares in the semantic domain, and permits one bounded correction. Operation-level bookkeeping
preserves the truth of `Applied`/`Ignored` (sent), `Ambiguous` (attempted), and
`AlreadySet`/`Suppressed` (not sent) across both attempts. Pair selection distinguishes equal, split,
and missing state. The bracket is explicitly not compare-and-set: an out-and-back ABA transition or
another controller between resolution and write remains possible without daemon-side atomic support.

The final read architecture is a value proof rather than a boolean verdict. Reconnect-capable lazy
`VolumeRange` discovery happens first. The read captures a four-family fingerprint plus cache
generation identity, reads `State`/`Status`, takes the live rates probe, and under one state lock
verifies the identity and returns the exact `ChainSnapshot` it proved. Projection consumes only that
snapshot. A partial cache retries; a cache replaced by a newer complete identity retries without
erasing the winner; only a live probe contradicting the same captured identity invalidates it as
moved. The single retry is shared across the whole read.

Writers use the same epoch rule. Full gathers capture `chain_generation` before their first request
and publish all four families only if it is unchanged. Every `fresh_*` captures the generation before
the request and performs transition detection plus publication under one write lock. An overtaken
mapping is refused without disturbing the winner; if the anchored winner has exactly the same
`(index, semantic value)` mapping, the setter may safely resolve through the reply without rewriting
or re-dating the winner. Identical cached contents are a no-op; clear/refill and changed publications
advance the wrapping generation.

The profile test uses ordinary protocol `Error` replies without disconnecting the fake daemon. The
volume regression uses the existing element-scoped one-shot dropped reply; no production sleep,
yield, or test-only hook was introduced. Twenty-one internal identity/publication tests cover all
four family identities, partial/unanchored sets, exact snapshot return, replacement preservation,
stale-probe ordering, full and single-family writer races, equal-winner acceptance, movement,
metadata-only change, byte-identical refill, overflow, and unchanged reads. Two AST tests pin exact
snapshot projection and unconditional winner re-read. Mutation checks killed the prior mechanisms,
including removal of either writer guard, each required generation bump, generation identity,
wrapping overflow, and whole-family equality.

Final exact head: `8970fd9cb93d537b6e7e234aef55c6e289a323bb`. Three consecutive
`hqplayer_conformance` runs passed 271/271. The complete workspace gate passed 675 tests with 12
explicit live tests ignored; CI.s exact `cargo clippy -- -D warnings`, fmt, diff, API contract,
protocol schema/integration, projection lint, and 21 cache-identity tests passed. The final `sg review`
found no blocker; its suggestions were non-blocking helper/readability deduplication and the explicitly
documented request amplification. A final Opus dissent process returned no report and made no edits;
the prior Opus P2/P3 findings were closed by the semantic-bracket tests above and a manual exact-diff
dissent found no remaining P1-P3.

Eight further expectations are **pins**, not reds — they already held and now cannot regress. Each is
labelled `client-red`, `client-pin` or `model-fidelity` in its own doc comment.

## Commands and results

```
cargo fmt --check                                   FMT_EXIT=0
cargo clippy -- -D warnings                         CLIPPY_EXIT=0   (CI's exact invocation, unpiped)
cargo test --workspace --no-fail-fast               TESTS_EXIT=0, 675 passed (12 live tests ignored)
```

| Target | Result |
|---|---|
| `hqplayer_conformance` | 271 passed |
| `hqplayer_pipeline_projection_lint` | 2 passed |
| `hqplayer_ledger_lint` | 40 passed |
| `api_contract` | 2 passed |
| `protocol_schema` | 41 passed |
| `adapter_integration` | 52 passed |
| `client_harness` | 46 passed |
| `protocol_integration` | 33 passed |
| `zones_sha_integration` | 30 passed |
| `architecture_lint` · `aggregator_lint` · `await_in_lock_lint` · `arbitrary_find_lint` · `ignored_send_lint` · `oneshot_leak_lint` · `spawn_cancellation_lint` · `unbounded_channel_lint` · `volume_safety` · `volume_step` | 47 passed |

`cargo check --features web` and `cargo build --bins` both succeed. **`dx build` was not run: `dx` is
not installed in this environment** — `command -v dx` exits 1, and `/Users/muness/.cargo/bin` (which
is on `PATH`) holds `ba`, `cargo-audit`, `repo-native-alignment`, `sg` and `wm` and no `dx`. So the
WASM half of the Dioxus build is unverified here. Nothing
under `src/app/` is touched and no UI file calls the adapter — the pages reach HQPlayer over HTTP —
so the exposure is a build-tooling gap rather than an untested code path, but it is a gap and CI is
what closes it.

## API and appliance boundaries

- **No API change.** No route added, removed or renamed. `tests/fixtures/api_routes.txt` is
  untouched — `git diff` against it is empty — and `cargo test --test api_contract` passes.
- Request schemas unchanged: `HqpSettingRequest { name, value: u32 }` and
  `HqpPipelineRequest { setting, value }` are as they were, numbers included.
- Response schemas unchanged: `{"ok":true}` / `{"error":…}` / `PipelineStatus` /
  `{profiles, current}`. `protocol_schema` pins them.
- **MCP unchanged.** Ten tools, same names, same descriptions, same arguments. No tool was added and
  none was made to do something its description does not promise.
- **No appliance was contacted.** Every test is hermetic against the #322 fake over loopback TCP. No
  credentials were used and `192.168.1.61` was never addressed.

## Dependency and merge order

Blocked by #322 (executable protocol foundation) and #341 (evidence ledger). This PR targets
`feat/issue-341-hqplayer-evidence-ledger` and must merge after it. It unblocks #325 and #329, which
own the producer and the verified live-setting surfaces.

**Ledger changes** (this PR's own rows): HQP-C-019 `open → settled`; HQP-C-026 and HQP-C-063
`open → retired`, both with the anchor rewritten to say what changed; HQP-C-064 `open → settled` (it
went stale the moment a conformance expectation drove `set_shaper`); HQP-C-004 restored to its
stronger wording, which #341 had scoped back because the two paths in HQP-C-063 made it false;
HQP-C-017 gains the client-side proof. Two tripwires added that fail on **regression** rather than on
the happy path — the asymmetry HQP-C-063's own anchor argued a proof must have.

## Residual risks

1. **The refresh bracket detects a chain that *ended* elsewhere than it started.** A change out and
   back inside one gather window would leave both probes agreeing. Nothing observed suggests that is
   reachable — a chain follows source material, not a poll cycle — and the guard is described as
   "did not straddle a visible transition", never as a same-instant snapshot.
2. **`GetRates` as the chain probe rests on HQP-C-020's disjointness.** The rate list also varies with
   the selected filter (HQP-C-021), so a filter change can move the fingerprint without a chain
   change: harmless, costing a refetch. The missed direction — a chain move leaving the rate list
   byte-identical — has no observation behind it, and is stated rather than assumed away.
3. **A numeric index sent as a JSON *string* to `/hqp/pipeline` now fails.** As a JSON *number* it
   still works, resolved at the boundary. The shipped clients send names (the pipeline GET's options
   carry names) or numbers, so no known caller regresses — but a caller sending `"3"` for `filter1x`
   gets an error naming the available filters instead of silently selecting position 3. That is the
   acceptance criterion, not an accident.
4. **`shaper_label` still derives from the configured mode.** Under `[source]` it can read "Shaper"
   over a loaded DSD chain. Changing a response *value* has no acceptance criterion here and was left
   alone deliberately; the loaded chain is now known, so #325/#329 can fix it cheaply.
5. **One extra command per pipeline poll** (the chain probe), and five on a detected transition. The
   alternative is a cache that can name a different filter than the user picked.
6. **The chain check confirms "same chain now as when the lists were published", not "one instant".**
   `get_pipeline_status` settles the lists, reads `State`, then re-checks — but a move out and back
   inside that window leaves the check agreeing. One retry bounds it; only a daemon-side atomic
   snapshot would close it, and the protocol offers none. When the retry is caught by a second
   transition the read **fails** rather than publishing: `GET /hqplayer/pipeline` answers its
   existing 500 error arm, which a flapping-source daemon can therefore reach. That is deliberate —
   the alternative was returning empty option lists as a successful read.
7. **`FilterFieldReporting` is a constructed shape, not evidence.** Every corpus capture carries both
   `filter1x` and `filterNx`; what it models is that UHC's *own parser* admits their absence. It is
   deliberately absent from `armed_upstream_claims`, which names unqualified claims about **daemons**.
8. **Tier-2 provenance is unchanged.** Source-following and the `[source]` rate refusal remain
   `derived-upstream, tier-2-only, pending #332`. This PR builds client behaviour on them; it does not
   promote them.
9. **Semantic proof is bracketed, not daemon-atomic.** Opening/closing enumeration fingerprints now
   catch a visible mapping move, re-resolve once, and prevent numeric-position false success. Another
   controller can still move the daemon between resolution and the wire, or move the mapping out and
   back (ABA) before the closing read. The protocol offers no compare-and-set; #162 retains the broader
   session/provenance boundary rather than this PR claiming atomicity against external controllers.
10. **`VolumeRange` is not yet host/session-versioned.** An old-host fetch can land after `configure`,
    and reconnect does not currently clear the cached range. #162 now owns provenance and invalidation
    for that non-chain cache; the current correction deliberately does not present it as solved.

## How to test

```bash
# The whole thing, hermetic, no daemon:
cargo test --test hqplayer_conformance --test hqplayer_ledger_lint --test api_contract

# The headline hazard on its own — the stale cross-chain filter selection:
cargo test --test hqplayer_conformance a_source_driven_chain_change_invalidates_the_cached_filter_list

# The contract that must not move:
cargo test --test api_contract && cargo test --test protocol_schema

# CI's own gates:
cargo fmt --check && cargo clippy -- -D warnings
```

To see a red rather than take one on trust, restore any pre-fix mechanism in
`src/adapters/hqplayer.rs` and re-run the target — the commit messages name which mechanism each red
depends on.

---

## Summary by CodeRabbit

* **Improvements**
  * HQPlayer setting changes now verify whether values were applied, already set, ignored, or rejected.
  * Invalid, unsupported, ambiguous, and unapplied settings now return clear errors instead of appearing successful.
  * Legacy numeric settings are resolved against the currently loaded HQPlayer options.
  * Filter updates, matrix profiles, samplerates, modes, and shapers now handle setting changes more reliably across reconnects and profile or chain changes.
  * Improved handling of fragmented, unsolicited, malformed, and oversized HQPlayer responses.

### DIFF
--- a/.oh/issue-347-hqplayer-trustworthy-setters.md
+++ b/.oh/issue-347-hqplayer-trustworthy-setters.md
@@ -19,9 +19,9 @@ field.
 **Key constraint:** no public API route/request/response contract changes, no live appliance, and the
 #322 executable corpus — not current Rust behaviour — is the protocol authority.
 
-**Success:** every supported live setter reports `applied`, `already-set`, `ignored`, `suppressed` or
-`ambiguous` from an authoritative `State` readback; enumerations are scoped to the loaded chain and
-invalidated when it moves; no raw integer reaches a control path.
+**Success:** every supported *enumerated* live setter reports `applied`, `already-set`, `ignored`,
+`suppressed` or `ambiguous` from an authoritative `State` readback; enumerations are scoped to the
+loaded chain and invalidated when it moves; no raw integer reaches a control path.
 
 ## Solution Space
 
@@ -71,13 +71,26 @@ compatibility boundary, never publish or persist it as identity.
   a cache that can name a different filter than the user picked.
 - No PCM/SDM label for the loaded chain. `shaper_label` keeps its existing configured-mode heuristic;
   changing a response *value* has no acceptance criterion and would be an unbudgeted risk.
-- `Ambiguous` is reachable only after a write was attempted and its reply lost — the client then does
-  a bounded readback rather than assuming failure (HQP-C-029).
+- `Ambiguous` is reachable after a write was attempted and its reply lost — the client then does a
+  bounded readback rather than assuming failure (HQP-C-029) — and also when the authoritative field is
+  absent, where nothing can be established either way.
+- **Volume is outside the outcome vocabulary and stays `Result<()>`.** Result-checked, never
+  readback-verified: adaptive volume moves the level on its own, so a readback would manufacture
+  failures, and volume is absolute dB rather than a list index (HQP-C-040). Documented on
+  `SettingOutcome` rather than silently implied, so nothing claims "every setter".
+- The refresh bracket detects a chain that **ended** somewhere other than where it started. A change
+  out and back inside one window would leave both probes agreeing. Nothing observed suggests that is
+  reachable, and the guard is described as "did not straddle a visible transition" rather than as a
+  same-instant snapshot, which nothing short of a daemon-side atomic read could give.
 
 ## Execute
 
-Status: complete. See the PR body for RED/GREEN evidence.
+**Status: in progress at the time of writing.** Two commits exist on
+`fix/issue-347-hqplayer-trustworthy-setters` (RED, then GREEN); nothing is pushed and no PR exists.
+An independent execute audit then raised nine items, addressed as forward work before any further
+commit. RED/GREEN evidence goes in the PR body once it is opened.
 
 ## Review / Dissent
 
-Recorded as PR comments 1–6.
+Not yet performed. The six report comments are written and posted after the PR exists, re-anchored to
+the final head.
--- a/.oh/issue-347-hqplayer-trustworthy-setters.md
+++ b/.oh/issue-347-hqplayer-trustworthy-setters.md
@@ -1,0 +1,83 @@
+# Issue #347 — HQPlayer: verify live setters and scope enumerations to the loaded chain
+
+OH: 80222d6d · epic #311 · stacked on #341 (`feat/issue-341-hqplayer-evidence-ledger`, head
+`6d688bdbcb3fa607ca87a1593dd692e7ba00133a`)
+
+## Aim
+
+A client issues a semantic setting change during source-following playback, and UHC either proves the
+exact setting applied or returns a bounded, truthful non-applied outcome without corrupting a sibling
+setting.
+
+## Problem Statement (confirmed from #347)
+
+**The problem:** `result="OK"` and a mode-keyed list cache are both treated as proof. The daemon can
+acknowledge a setter without applying it; the loaded PCM/SDM chain can move while configured mode
+stays `[source]`; and the one-sided `SetFilter` helpers guess the sibling from a different `State`
+field.
+
+**Key constraint:** no public API route/request/response contract changes, no live appliance, and the
+#322 executable corpus — not current Rust behaviour — is the protocol authority.
+
+**Success:** every supported live setter reports `applied`, `already-set`, `ignored`, `suppressed` or
+`ambiguous` from an authoritative `State` readback; enumerations are scoped to the loaded chain and
+invalidated when it moves; no raw integer reaches a control path.
+
+## Solution Space
+
+### Axis 1 — how a setter outcome is represented
+
+| Option | Level | Approach | Trade-off |
+|---|---|---|---|
+| A | Band-Aid | Keep `Result<()>`, put the distinction in error prose | "Ignored" is indistinguishable from "rejected" to any caller; no type stops a future call site from treating OK as applied |
+| B | Local Optimum | `Result<()>` plus a side-channel counter/log | Observable only out-of-band; a boundary cannot map outcomes |
+| C | Reframe | Adapter setters return `Result<SettingOutcome>`, `#[must_use]`, with `Applied / AlreadySet / Ignored / Suppressed / Ambiguous`; HTTP+MCP boundaries collapse it to today's `{ok:true}` / `{error}` shapes | Every call site must be updated; five variants to keep honest |
+| D | Redesign | Move live-setting ownership to a producer/aggregator contract with an outcome event stream | #347 explicitly defers this to #325/#329; would balloon scope |
+
+**Selected: C.** The issue names four outcomes and requires them "observably … using existing
+internal contracts (no public API change)". A type is the only representation a later call site
+cannot silently discard, and `Result<SettingOutcome>` keeps transport/rejection failures where they
+already are. `AlreadySet` is the fifth because *skipping* a write is a distinct fact from *applying*
+one — the no-op mode skip exists precisely so nothing is sent.
+
+### Axis 2 — how the loaded chain is resolved
+
+| Option | Level | Approach | Trade-off |
+|---|---|---|---|
+| A | Band-Aid | `refresh_lists()` before every setter | 4 extra round trips per write; never *detects* a change, so nothing else can react |
+| B | Local Optimum | Keep the mode-keyed cache, add `State.active_mode` as a second key | Settles by fiat exactly what HQP-C-024 records as **unmeasured**. Rejected on evidence grounds |
+| C | Reframe | **The loaded chain is the enumeration the daemon currently serves.** Per-family fingerprints; a fresh fetch whose fingerprint differs from the cached one *is* a chain transition and invalidates every chain-scoped family | Needs one fresh fetch per control operation; no PCM/SDM *label* falls out |
+| D | Redesign | A chain-identity event on the bus that producers subscribe to | #325/#329 own the producer; premature here |
+
+**Selected: C.** It is independent of configured mode and of rate-pin availability *by construction*,
+it invents no threshold and no unmeasured field reading, and HQP-C-008 (`E0-uhc-live`) already
+settles that enumerations are chain-scoped. The deliberate consequence: UHC identifies *which* chain
+is loaded, not what to *call* it. Nothing in #347 needs the label, and inventing a PCM/SDM classifier
+from rate spans would be the exact failure mode this epic exists to end.
+
+### Axis 3 — the legacy numeric HTTP contract
+
+| Option | Level | Approach | Trade-off |
+|---|---|---|---|
+| A | Band-Aid | Keep the adapter's `parse::<u32>()` fallback | The acceptance criterion forbids it; a stale index silently selects a different setting |
+| B | Reframe | Resolve the number to a **name** at the HTTP boundary against the fresh enumeration, then call the semantic setter | Boundary gains a lookup and a failure mode; request/response contracts unchanged |
+
+**Selected: B.** It is what #347 and HQP-C-063 jointly prescribe: keep the number at the
+compatibility boundary, never publish or persist it as identity.
+
+### Accepted trade-offs
+
+- One fresh enumeration fetch per control operation. Correctness over round trips; the alternative is
+  a cache that can name a different filter than the user picked.
+- No PCM/SDM label for the loaded chain. `shaper_label` keeps its existing configured-mode heuristic;
+  changing a response *value* has no acceptance criterion and would be an unbudgeted risk.
+- `Ambiguous` is reachable only after a write was attempted and its reply lost — the client then does
+  a bounded readback rather than assuming failure (HQP-C-029).
+
+## Execute
+
+Status: complete. See the PR body for RED/GREEN evidence.
+
+## Review / Dissent
+
+Recorded as PR comments 1–6.
--- a/docs/hqplayer-evidence-ledger.md
+++ b/docs/hqplayer-evidence-ledger.md
@@ -108,8 +108,8 @@ were **not executed** and are recorded as gaps in that comment.
 | HQP-C-001 | `SetMode value=` carries the **list index**, not the enum ID: SDM is index 2 while its enum ID is 1 | E0-uhc-live | PR #337 comment 5135836825 · read-via-pr · Embedded 6.0.2 / engine 6.0.4 · 2026-07-30 · idle | test:a_delayed_set_mode_still_clamps_indices_into_the_loaded_chain · test:modes_list_distinguishes_list_index_from_enum_id · fixture:tests/fixtures/hqplayer/hqpd-6.0.4-opal/modes.xml | settled | — |
 | HQP-C-002 | Domain 1 — the **live wire** domain: `State.mode/filter/filter1x/filterNx/shaper/rate` and the **enumerated** setters (`SetMode`, `SetFilter`, `SetShaping`, `SetRate`, `SetJunkFilter`) speak a **list index** into the currently loaded enumeration. Volume is **not** in this domain — it is absolute dB (HQP-C-040). Commands evidenced elsewhere: SetShaping (HQP-C-064), SetJunkFilter (HQP-C-051) | E0-uhc-live | PR #337 comment 5135836825 · read-via-pr · Embedded 6.0.2 / engine 6.0.4 · 2026-07-30 · idle | test:a_filter_name_is_sent_as_the_index_the_observed_list_gives_it · test:a_filter_name_is_not_sent_as_its_enum_id · test:a_delayed_set_mode_still_clamps_indices_into_the_loaded_chain · test:a_rate_valid_in_one_chain_is_refused_in_the_other | settled | — |
 | HQP-C-003 | Domain 2 — the **persistent** domain: `hqplayerd.xml` and the 8088 config form store **enum IDs**, which are not list indices and must never be fed to the live lane | E1-upstream-verified | UHC-SALVAGE reports via `.oh/issue-322-hqplayer-protocol-conformance.md` · read-via-report · hqplayerd 6.0.4 (Opal) · 2026-07-29 · idle | test:the_persistent_configuration_lane_stores_enum_ids_not_list_indices · test:feeding_a_persistent_enum_id_to_the_live_lane_is_rejected | settled | — |
-| HQP-C-004 | Domain 3 — the **client** domain **as designed**: clients exchange setting-appropriate values — semantic names for mode/filter/shaper, Hz for rate, decimal dB for volume — and the adapter owns both numeric conversions. **This is the design rule, not a description of every path today**: see HQP-C-063 | E6-documentary | `.oh/hqplayer-spec.md` layer table · direct · n/a · 2026-02-04 · n/a | test:a_filter_name_is_sent_as_the_index_the_observed_list_gives_it | settled | — |
-| HQP-C-063 | Raw indices **do** cross the client boundary today, in two evidenced places: the legacy numeric route `HqpSettingRequest { name: String, value: u32 }` stringifies its number for name-based lookup (`src/api/mod.rs:825-836`), and the adapter's resolvers fall back to parsing a numeric string as a direct index (`resolve_mode_index:2081`, `resolve_filter_index:2186`, `resolve_shaper_index:2247`) | E6-documentary | `src/api/mod.rs` and `src/adapters/hqplayer.rs` at this head · direct · n/a · 2026-07-30 · n/a | none:#347 removing the raw integer-string fallback from production control paths while keeping the legacy numeric HTTP contract at the compatibility boundary | open | #347 |
+| HQP-C-004 | Domain 3 — the **client** domain: clients exchange setting-appropriate values — semantic names for mode/filter/shaper, Hz for rate, decimal dB for volume — and the adapter owns both numeric conversions. The two paths that broke this rule are gone (HQP-C-063, retired); the one frozen numeric request contract is resolved to a name at the HTTP boundary and never travels inward | E6-documentary | `.oh/hqplayer-spec.md` layer table · direct · n/a · 2026-02-04 · n/a | test:a_filter_name_is_sent_as_the_index_the_observed_list_gives_it · test:a_numeric_string_is_never_accepted_as_a_setting_name | settled | — |
+| HQP-C-063 | Raw indices **did** cross the client boundary, in two evidenced places: the legacy numeric route `HqpSettingRequest { name: String, value: u32 }` stringified its number for name-based lookup (`src/api/mod.rs:825-836`), and the adapter's resolvers fell back to parsing a numeric string as a direct index (`resolve_mode_index:2081`, `resolve_filter_index:2186`, `resolve_shaper_index:2247`) | E6-documentary | `src/api/mod.rs` and `src/adapters/hqplayer.rs` at `6d688bd` · direct · n/a · 2026-07-30 · n/a | test:a_numeric_string_is_never_accepted_as_a_setting_name · test:tests/hqplayer_ledger_lint.rs::no_production_control_path_parses_a_setting_name_as_an_index | retired | #347 |
 | HQP-C-064 | **No conformance expectation drives `set_shaper`**, so the shaper setter's behaviour is L1-observed only — `shaper 24→0→24`, `result="OK"`, readback-verified on the live run — and unpinned hermetically | E0-uhc-live | PR #337 comment 5135836825 · read-via-pr · Embedded 6.0.2 / engine 6.0.4 · 2026-07-30 · idle | none:a conformance expectation that drives `set_shaper` and verifies the readback; the fake already applies `SetShaping` (`model.rs`), so only the expectation is missing | open | #329 |
 | HQP-C-005 | `SetMode` does **not** reset the filter or shaper selections; the fake once modelled that and no evidence supports it | E3-derived | `.oh/issue-322-…` amendment row B14 · direct · hqplayerd 6.0.4 (Opal) · 2026-07-30 · unknown | test:set_mode_does_not_reset_the_filter_and_shaper_selections | settled | — |
 | HQP-C-006 | A mode change reloads the chain, so a list index resolved before it can name a **different** setting after it | E1-upstream-verified | UHC-SALVAGE reports via `.oh/issue-322-…` · read-via-report · hqplayerd 6.0.4 (Opal) · 2026-07-29 · idle | test:a_delayed_set_mode_still_clamps_indices_into_the_loaded_chain · test:the_same_filter_index_selects_a_different_filter_per_loaded_chain | settled | — |
@@ -133,9 +133,9 @@ were **not executed** and are recorded as gaps in that comment.
 |---|---|---|---|---|---|---|
 | HQP-C-015 | `SetRate value=` is the **list index** of the wanted rate in the current chain's list; `RatesItem` carries `rate` in Hz and **no** enum ID; `Status.active_rate` reports Hz | E0-uhc-live | PR #337 comment 5135836825 · read-via-pr · Embedded 6.0.2 / engine 6.0.4 · 2026-07-30 · idle | test:rates_list_reports_hz_and_has_no_enum_id · test:a_rate_valid_in_one_chain_is_refused_in_the_other | settled | — |
 | HQP-C-016 | Rate list index **0 is Auto** (`rate="0"`), and a mode change resets the runtime rate to it — L1 observed `rate 3 → 0` | E0-uhc-live | PR #337 comment 5135836825 · read-via-pr · Embedded 6.0.2 / engine 6.0.4 · 2026-07-30 · idle | test:tier1_requires_rate_index_zero_to_be_auto · test:a_same_mode_set_mode_still_clears_the_rate_pin | settled | — |
-| HQP-C-017 | A **same-mode** `SetMode` still reloads the chain and clears the exact-rate pin, so a no-op mode write is not a no-op | E1-upstream-verified | UHC-SALVAGE-BETA-DEV via `.oh/issue-322-…` · read-via-report · hqplayerd 6.0.4 (Opal) · 2026-07-29 · idle | test:a_same_mode_set_mode_still_clears_the_rate_pin | settled | — |
+| HQP-C-017 | A **same-mode** `SetMode` still reloads the chain and clears the exact-rate pin, so a no-op mode write is not a no-op | E1-upstream-verified | UHC-SALVAGE-BETA-DEV via `.oh/issue-322-…` · read-via-report · hqplayerd 6.0.4 (Opal) · 2026-07-29 · idle | test:a_same_mode_set_mode_still_clears_the_rate_pin · test:a_no_op_mode_write_is_not_sent_so_the_rate_pin_survives | settled | — |
 | HQP-C-018 | Under `[source]` a **non-zero** rate pin is accepted and ignored: the setter answers `OK` and the runtime rate does not move | E1-upstream-verified | UHC-SALVAGE-BETA-DEV via `.oh/issue-322-…` · read-via-report · hqplayerd 6.0.4 (Opal) · 2026-07-29 · idle | test:a_nonzero_rate_pin_under_source_is_accepted_and_ignored | settled | — |
-| HQP-C-019 | Under `[source]` an **Auto** (index 0) rate request is ignored and a readback **cannot tell** — comparing 0 against 0 reports success for a command that did nothing | E1-upstream-verified | UHC-SALVAGE-BETA-DEV via `.oh/issue-322-…` · read-via-report · hqplayerd 6.0.4 (Opal) · 2026-07-29 · idle | test:an_auto_rate_request_under_source_is_ignored_and_readback_cannot_tell | open | #347 |
+| HQP-C-019 | Under `[source]` an **Auto** (index 0) rate request is ignored and a readback **cannot tell** — comparing 0 against 0 reports success for a command that did nothing. The client therefore **suppresses** the write with a stated reason instead of verifying it | E1-upstream-verified | UHC-SALVAGE-BETA-DEV via `.oh/issue-322-…` · read-via-report · hqplayerd 6.0.4 (Opal) · 2026-07-29 · idle | test:an_auto_rate_request_under_source_is_ignored_and_readback_cannot_tell · test:an_auto_rate_request_under_source_is_never_reported_as_applied · test:a_rate_pin_under_source_is_suppressed_before_it_reaches_the_daemon | settled | — |
 | HQP-C-020 | The offered rate list varies wholesale **by mode** — PCM 44.1k–768k against SDM 2.8M–24.6M on L1, with no overlap | E0-uhc-live | PR #337 comment 5135836825 · read-via-pr · Embedded 6.0.2 / engine 6.0.4 · 2026-07-30 · idle | test:a_rate_valid_in_one_chain_is_refused_in_the_other | settled | — |
 | HQP-C-021 | The rate list also depends on the **selected filter**, so mode alone is insufficient to resolve a rate | E1-upstream-verified | UHC-SALVAGE-BETA-DEV §4 via #341 [comment 5126438674](https://github.com/open-horizon-labs/unified-hifi-control/issues/341#issuecomment-5126438674) (`livelane.py:33-38`) · read-via-report · hqplayerd 6.0.4 (Opal) · 2026-07-29 · idle | none:capture GetRates before and after a SetFilter on one device and record which entries change | open | #332 |
 | HQP-C-022 | Whether `GetRates` can be **empty** at all is unsupported by any observation in the audited evidence base | E4-unverified | #341 comment 5126438674 · read-via-issue · n/a · 2026-07-30 · n/a | none:read-only GetRates capture on a daemon with no usable output configured, or upstream confirmation that an empty list is reachable | open | #332 |
@@ -147,7 +147,7 @@ were **not executed** and are recorded as gaps in that comment.
 | HQP-C-023 | `Status.active_mode` **echoes the configured mode** under `[source]`, so it cannot resolve the loaded chain | E1-upstream-verified | UHC-SALVAGE reports via `.oh/issue-322-…:1549-1552` and `model.rs:126`/`:146` at `bc9158e` · read-via-report · hqplayerd 6.0.4 (Opal) · 2026-07-29 · active | test:the_fake_does_not_settle_the_independent_state_and_status_active_mode_semantics | settled | — |
 | HQP-C-024 | What `State.active_mode` reports under `[source]` — the loaded chain or the configured mode — is **unmeasured**, upstream and here | E4-unverified | `.oh/issue-322-…:1738-1740` · direct · unmeasured on any daemon · 2026-07-30 · unknown | test:the_fake_does_not_settle_the_independent_state_and_status_active_mode_semantics · #332:Resolve State.active_mode versus Status.active_mode under configured PCM/SDM and [source]/Auto with PCM and DSD sources | open | #332 |
 | HQP-C-025 | `Status` reports active settings as **display names** while `State` reports numbers, so the two are complementary rather than redundant | E3-derived | `tests/fixtures/hqplayer/hqpd-6.0.4-opal/status_playing_with_metadata.xml` · read-via-report · hqplayerd 6.0.4 (Opal) · 2026-07-29 · unknown | test:status_reports_active_settings_as_display_names | settled | — |
-| HQP-C-026 | UHC's own comment at `src/adapters/hqplayer.rs:2473` calls `Status.active_mode` "unreliable" and instructs using `State`'s — a global rule the evidence does not support | E4-unverified | `src/adapters/hqplayer.rs:2473` · direct · n/a · 2026-07-30 · n/a | none:HQP-C-024 settling, after which the comment states a measured fact or is deleted | open | #347 |
+| HQP-C-026 | UHC's own comment at `src/adapters/hqplayer.rs:2473` called `Status.active_mode` "unreliable" and instructed using `State`'s — a global rule the evidence does not support | E4-unverified | `src/adapters/hqplayer.rs:2473` · direct · n/a · 2026-07-30 · n/a | test:tests/hqplayer_ledger_lint.rs::no_production_comment_settles_the_active_mode_question_by_fiat | retired | #347 |
 
 #### HQP-C-023's playback state was corrected, and the wrong value had already travelled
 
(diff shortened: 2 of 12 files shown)


muness:
@coderabbitai review

Draft, stacked on #341. Please review at exact head `2eb699ef5ddb213864db0a8983ef7f352a299812`.

Areas where a second pair of eyes is most valuable:
- `refresh_lists`'s bracketed gather in `src/adapters/hqplayer.rs` — whether the opening/closing `GetRates` probe is sufficient to reject a set that straddles a chain change, and whether the stated residual (out-and-back inside one window) is the only gap.
- The compatibility boundary in `src/api/mod.rs`: `hqp_apply_legacy_setting` / `hqp_apply_named_setting` against the frozen request contracts of `POST /hqplayer/setting` and `POST /hqp/pipeline`. I found and fixed one widening myself (`dither` leaking into the legacy route via the shared applier, `2eb699e`); please look for others.
- `SettingOutcome` and `HqpRejected`: whether any surface can still report a non-applied write as success.

muness:
@coderabbitai review

Both findings from your review of `2eb699e` are fixed at exact head `2349a73191ddce95bce91f78cc4adfe6587c4c5d`. Please re-review at that head.

**1 · `get_pipeline_status` read `State` before the lists.** Confirmed and reproduced before fixing — the published 1x filter was `""` while the daemon held `poly-sinc-xtr`, because index 7 belonged to the 12-entry PCM excerpt, the daemon clamped it to 4 across the transition, and 7 resolves to nothing in the 5-entry SDM list that was then used. Now: settle the lists → read `State` → re-check the chain against the fingerprint the lists were published under, with one bounded retry. `chain_probe` became `chain_still_matches_cache` and returns the answer rather than swallowing it. Test: `a_chain_change_between_the_state_read_and_the_list_read_is_not_published`, mutation-checked.

The remaining race you asked to see documented is stated at the call site and as residual risk 6 in the PR body: the check confirms *same chain now as when the lists were published*, so a move out and back inside that window leaves it agreeing. I could not find a way to close it without a daemon-side atomic snapshot the protocol does not offer — if you see one, that is the most valuable thing you could tell me.

**2 · `set_filter`'s raw-index form returned `Ok(())` without readback.** Agreed, and "escape hatch" was not a good reason. It now goes through `write_setting` like every other write and verifies both sides as one value. Kept public and compatible: the three cross-lane checks that feed a persistent-domain enum ID to the live lane still compile and pass, because a rejection is still an `Err`. Test: `a_raw_filter_write_the_daemon_acknowledges_and_drops_is_not_reported_as_success`, mutation-checked.

While there I also found the `set_filter` doc comment had been orphaned onto `filter_request` by an earlier edit, so the public method carried none.

Still worth your attention at this head:
- The bracketed gather in `refresh_lists` — opening and closing `GetRates` around the four fetches, publishing all four or none.
- `hqp_apply_legacy_setting` / `hqp_apply_named_setting` in `src/api/mod.rs` against the frozen request contracts. I found one widening myself (`dither` leaking into `/hqplayer/setting` via the shared applier) and it now has four boundary tests in `client_harness`; please look for others I missed.

muness:
@coderabbitai review

Supersedes my request at `2349a73` — that head is stale. Please review at exact head `e3a2c088f4e03a5b0b2852dfce4c26de99a71fbd`.

Your two findings on `2eb699e` are fixed, and reviewing my own fix for the first one turned up a **third** hole in the same read path, now also fixed:

**`get_pipeline_status`'s bounded retry was not allowed to fail.** When the chain moved during the `State` read on the *retry* as well, the second-attempt branch set the state and published anyway — but `chain_still_matches_cache()` had already invalidated on its way to returning false, so what got published was a state read against nothing: empty option lists, every selection resolving to `""`, presented as a successful read. A caller cannot tell that from a daemon that genuinely offers no filters. It now returns an error naming what could not be established; `GET /hqplayer/pipeline` has always had an error arm, so no contract moves. Test: `a_chain_that_moves_on_both_the_read_and_the_retry_is_reported_rather_than_published`, RED exit 101 with `published [], filter1x selected as "" out of 0 options`, mutation-checked.

So the read path now has three ordered guarantees, and I would value your eyes on whether they compose:
1. `refresh_lists` publishes four coherent lists or none, bracketed by `GetRates` at both ends.
2. `get_pipeline_status` settles the lists, *then* reads `State`, then re-checks the chain.
3. A second transition during the retry is an error, not a partial answer.

The residual I still cannot close, stated at the call site and as residual risk 6: the check confirms *same chain now as when the lists were published*, so a move out and back inside one window leaves it agreeing. If you can see a way to close that without a daemon-side atomic snapshot the protocol does not offer, that is the most valuable thing you could tell me.

Also still worth attention: `hqp_apply_legacy_setting` / `hqp_apply_named_setting` in `src/api/mod.rs` against the frozen request contracts of `POST /hqplayer/setting` and `POST /hqp/pipeline`. I found one widening myself (`dither` leaking into the legacy route via the shared applier) and it now has four boundary tests in `client_harness`; please look for others I missed.

muness:
@coderabbitai review

Supersedes my request at `e3a2c08`. Please review at exact head `d4acddf7d9a3592dc65b0f16c7639b9b1212261a`.

Your finding is fixed, and it was the sharper version of the one before it — thank you for following the thread down a layer.

**A failed or straddled required refresh returned a successful empty pipeline.** `refresh_lists()` returned nothing, so `get_pipeline_status()` could not tell "four coherent lists" from "published nothing". With an empty cache `chain_still_matches_cache()` has no previous fingerprint to contradict and correctly answers *unchanged* — it is a check about **movement** — so the required fill was the only thing left, and its failure was swallowed.

The part worth calling out: `a_single_failed_family_publishes_no_part_of_the_refresh` **pinned** that response. I wrote a test that made the bad answer the expected one, which is a failure mode I will be watching for now.

`refresh_lists` now returns whether it published a complete coherent snapshot and is `#[must_use]`, so a future caller cannot drop it the way this one did. A required fill that does not settle takes the read's one retry, then fails naming what could not be established. All-or-none publication is unchanged and a valid cache still short-circuits. `load_profile` logs rather than fails — the profile load did succeed.

Tests, each RED at exit 101 against the previous production code and mutation-checked:
- `a_single_failed_family_fails_the_read_rather_than_publishing_an_empty_one`
- `a_first_read_whose_required_fill_never_settles_is_not_reported_as_success`
- `a_chain_change_during_a_list_refresh_publishes_nothing_rather_than_a_mixed_set` (rewritten from the version that pinned `Ok`)

Also finished deliberately while there: the retry loop no longer carries an `Option` and an unreachable re-read of `State` — every arm is terminal, so there is no fall-through case to answer for. Behaviour-preserving, suite green across it.

And a correction to my own evidence, made forward rather than by rewriting: my commit messages at `2349a73` and `e3a2c08` said "487 passed". That was my arithmetic, not cargo's — the real totals were 536 and 538. Per-target numbers were right; only the sum was wrong.

The read path now has four ordered guarantees. I would value your judgement on whether they compose, and on anything that can still reach a caller as a confident answer it should not be:
1. `refresh_lists` publishes four coherent lists or none, bracketed by `GetRates` at both ends, and **says which**.
2. A required fill that does not settle fails the read after one retry.
3. `get_pipeline_status` settles the lists, *then* reads `State`, then re-checks the chain.
4. A second transition during the retry is an error, not a partial answer.

muness:
@coderabbitai review

Supersedes the review at `d4acddf7d9a3592dc65b0f16c7639b9b1212261a`. Please review the exact head `619dee47aff4366db1eff24351511c66b920075d`.

Your prior finding is fixed forward. After either successful profile POST path, every chain-scoped cache family is invalidated before the eager refresh. The regression does not simulate a disconnect or transport timeout: it returns protocol-level `Error` for `GetShapers` once during the eager post-load refresh and twice during the next status read’s bounded fill/retry. Against `d4acddf`, that status read reused the old list containing `NS9`; at this head it cannot return pre-profile enumerations as current.

A separate reconnect regression caught another successful-empty response when a connection change left the lists absent while the old presence check ignored the missing rate fingerprint. The cache-completeness check now includes that fingerprint. Current writers publish and invalidate all fields in lockstep, so this is documented as defensive invariant enforcement.

Please inspect especially:
- both profile-success paths and whether stale enumerations can survive any successful load;
- `chain_lists_incomplete` and the read/retry behavior across reconnects;
- the protocol-fault test and its process-global `UHC_CONFIG_DIR` serialization;
- any remaining path that can return a confident pipeline from incomplete, pre-profile, or mixed enumerations.

Hermetic exact-head result: `hqplayer_conformance` 249/249 in three consecutive runs; the full relevant 18-target gate is 540 passed. No appliance was contacted, no API contract changed, and no merge is authorized.

muness:
@coderabbitai review

Supersedes the review at `619dee47aff4366db1eff24351511c66b920075d`. Please review exact head `b7a7a1a0d8db9996d395e52ada0c71a439466eb1`.

Your finding is fixed forward with two client-red expectations. Against the prior production arm:
- two protocol-level `Error` replies for the final `GetRates` still returned all four option families as a confident pipeline;
- one rejection followed by a healthy daemon consumed only one `GetRates`, proving the rejection was swallowed and no successful verification preceded publication.

The final chain-check error now shares the read path’s existing one-retry bound: the first error re-reads, the second returns a descriptive error. The companion case proves a successful check on the retry publishes and consumes exactly two `GetRates`. The cache, connection, API, and protocol contracts are otherwise unchanged.

Please inspect whether any remaining branch can project `State` through lists without a successful final chain proof, and whether the shared retry flag composes safely with earlier fill/mismatch failures. The intended bound is one retry per whole pipeline read, not one per failure mode; exhaustion must fail rather than publish.

Exact-head hermetic evidence: 251/251 `hqplayer_conformance` in three consecutive runs; full relevant 18-target gate 542 passed; `cargo clippy -- -D warnings`, fmt, diff, API contract, and protocol schema green. `sg review` found no P1–P3 issue and treated the shared retry scope as a deliberate P4 tradeoff. No appliance access and no merge authorization.

muness:
@coderabbitai review

Supersedes the review at `b7a7a1a0d8db9996d395e52ada0c71a439466eb1`. Please review exact head `9b1fdfe533edac22878c5cebc0d453125d55e6ba`.

Your `VolumeRange` finding is fixed at the architectural boundary, together with the broader check-then-re-read race found by independent Opus dissent:
- reconnect-capable lazy `VolumeRange` discovery now precedes list/state coherence work;
- the read captures all four family fingerprints before `State`, not rates alone;
- the final live rates probe and captured identity are verified under one state lock;
- that proof returns the exact `ChainSnapshot` projection uses, and an AST lint forbids `get_pipeline_status` from re-reading shared state;
- `Partial`, `Replaced`, and `Moved` are distinct. A newer complete replacement fails this read and survives. Only a probe contradicting the same captured cache invalidates it.

Please inspect especially the full interleaving from identity capture through snapshot return, the classification order (`Replaced` before `Moved` because the probe predates the lock), all cache writers, the shared one-retry behavior, and whether any reconnect-capable command or direct cache read remains after proof.

New RED evidence against the prior shapes:
- lost first-read `VolumeRange` reply returned `Ok` with 0/4 families;
- projection lint found three state-lock reads in the old check-then-re-read function;
- an old probe versus a newer complete different-rate cache returned `Moved` and erased the winner until classification was reordered.

Exact-head hermetic evidence: 252/252 HQPlayer conformance in three consecutive runs; 19 relevant targets/544 tests; complete cargo test 643 tests; exact clippy, fmt, diff, API contract, protocol schema green. Fresh `sg review` found no requested changes and confirmed the replacement/movement ordering. No API/MCP contract change, no appliance access, and no merge authorization.

muness:
@coderabbitai review\n\nPlease review exact head `412f20cbae0b56e2d6161ebce8b9e32819e62474`. This head addresses the prior stale full-gather publication finding plus the follow-on single-family writer, false invalidation, and winner re-read races. No API or appliance changes.

muness:
@coderabbitai review

Please review exact head `412f20cbae0b56e2d6161ebce8b9e32819e62474`. This head addresses the stale full-gather publication finding and the follow-on single-family writer, false-invalidation, and winner re-read races. No API or appliance changes.

muness:
@coderabbitai review

Please review exact head `78c8e97e09e8470130457aef8e58569f9898bf65` against your semantic-setter finding at `412f20c`.

This head adds an opening/closing enumeration proof with one bounded correction for mode, filter pair/sides, shaper, and rate. It preserves operation-level wire truth across retries; distinguishes equal/split/missing filter state; reports semantic names/Hz; re-checks `[source]` before each rate write; and explicitly documents ABA/non-atomic residuals. The three post-write cross-chain cases now require `Ambiguous` with wire disclosure when the requested value is unavailable on retry.

Local exact-head evidence: `hqplayer_conformance` 269/269 three consecutive runs; workspace 673 passed with 12 live tests ignored; fmt, clippy `-D warnings`, diff, API contract, protocol schema/integration, projection lint, and 21 cache-identity tests passed. `src/api/mod.rs`, `src/mcp/mod.rs`, and `tests/fixtures/api_routes.txt` are unchanged. No appliance was contacted.

Superego exact-diff review: no blocker. It approved the shared semantic-family abstraction, paired outcome/state assertions, and per-attempt source suppression. Its non-blocking suggestions were helper deduplication and extracting the final outcome match for readability; both are deferred as scope-neutral cleanup. A final Opus dissent process stalled without a report or edits; the prior Opus P2/P3 findings are represented by the new tests and a manual exact-diff dissent found no remaining P1-P3.

muness:
@coderabbitai review

Your finding at `78c8e97` reproduced RED exactly: `a_rate_write_that_enters_source_after_the_decision_is_not_applied` returned `Applied` after configured mode changed to `[source]` between the decision `State` reply and `SetRate`.

Fixed at exact head `5f0e2b98fb5bed62c11d807ea795fbd1cf8aaf71`. Rate semantic proof now resolves the configured mode from its final proof state; if a write reached or may have reached the wire and final mode is `[source]`, it returns wire-aware `Ambiguous` before numeric Auto equality can become success. The fake daemon now has a one-shot external-mode interleaving matching its existing chain interleaving.

Exact-head evidence: focused RED then GREEN; `hqplayer_conformance` 270/270 three consecutive runs; workspace 674 passed with 12 live tests ignored; fmt, clippy `-D warnings`, diff, API contract, protocol schema/integration, projection lint, and 21 cache-identity tests passed. API/MCP/routes remain unchanged; no appliance contact.

Superego follow-up review found no blocker. Its only design observation was that the rate-only proof policy is represented by a boolean on the shared family descriptor; retained because the policy belongs beside each family proof and remains exhaustive.

muness:
@coderabbitai review

Your finding at 5f0e2b9 reproduced RED exactly: a_rate_proof_rechecks_state_after_fetching_modes returned Applied when configured mode changed to [source] immediately after the proof State reply.

Fixed at exact head 8970fd9cb93d537b6e7e234aef55c6e289a323bb. Rate proof now fetches modes, then re-reads State and uses that state for both configured-source detection and final semantic selection. A post-write source observation returns wire-aware Ambiguous. The remaining post-State interval is explicitly the non-atomic protocol boundary; HQPlayer exposes no compare-and-set or atomic snapshot.

Exact-content evidence: focused RED then GREEN; hqplayer_conformance 271/271 three consecutive runs; workspace 675 passed with 12 live tests ignored; fmt, clippy -D warnings, diff, API contract, protocol schema/integration, projection lint, and 21 cache-identity tests passed. API/MCP/routes remain unchanged; no appliance contact.

Superego exact-diff review found no blocker and called the fix solid and well-scoped. Its only design note was to state the unavoidable non-atomic boundary, which the production comment and PR residual-risk section do.

muness:
```json
{
  "report": "uhc_workflow_report_v1",
  "report_number": 1,
  "stage": "solution",
  "role": "review",
  "head": "8970fd9cb93d537b6e7e234aef55c6e289a323bb",
  "verdict": "pass",
  "aim": "A semantic HQPlayer setting request either proves the exact value applied or returns a bounded truthful non-applied outcome without corrupting sibling state.",
  "scope_checked": [
    "Issue #347 framing and acceptance criteria",
    "Setter outcome model",
    "Loaded-chain enumeration identity",
    "Filter-pair integrity",
    "Existing HTTP and MCP compatibility boundaries"
  ],
  "out_of_scope": [
    "Adaptive-control producer work owned by #325/#329",
    "Public API or MCP schema changes",
    "Playlist and UI work",
    "Live appliance qualification"
  ],
  "evidence": [
    "Issue #347 problem statement and completion signal",
    "Stack base 6d688bdbcb3fa607ca87a1593dd692e7ba00133a",
    "PR #366 Problem, Solution, RED-to-GREEN, and residual-risk sections"
  ],
  "findings": [
    {
      "severity": "none",
      "summary": "The solution addresses acknowledgement-vs-application, semantic identity, chain movement, source-mode rate refusal, and sibling-filter safety without moving public contracts."
    }
  ],
  "contrary_case": [
    "A simple setter-OK plus one State readback cannot prove Auto under [source] or preserve semantic meaning when enumerations move."
  ],
  "required_actions": [],
  "dispositions": [
    "Use semantic names internally and retain numeric compatibility only at existing boundaries",
    "Bracket semantic decisions with fresh enumerations and one bounded correction",
    "Expose ignored, suppressed, and ambiguous outcomes instead of manufacturing success"
  ],
  "residual_risks": [
    "HQPlayer provides no atomic compare-and-set or state/enumeration snapshot",
    "ABA movement can evade an opening/closing fingerprint bracket"
  ],
  "blockers": [],
  "authorization": "Authorized for implementation and draft PR evidence only; not authorized to merge, mark ready, alter public API/MCP contracts, or contact the live appliance."
}
```

muness:
```json
{
  "report": "uhc_workflow_report_v1",
  "report_number": 2,
  "stage": "solution",
  "role": "dissent",
  "head": "8970fd9cb93d537b6e7e234aef55c6e289a323bb",
  "verdict": "pass_with_explicit_residuals",
  "aim": "Challenge whether the proposed proof model can falsely report a semantic HQPlayer setting as applied.",
  "scope_checked": [
    "Alternative response-based success model",
    "External-controller interleavings",
    "Reconnect and cache replacement",
    "Source-following mode semantics",
    "Operation-level wire truth across retries"
  ],
  "out_of_scope": [
    "Daemon changes requiring Signalyst support",
    "Atomicity the native protocol does not expose",
    "New public endpoints or tools",
    "Live-device evidence promotion"
  ],
  "evidence": [
    "HQP-C-019, HQP-C-020, HQP-C-028, HQP-C-029, and HQP-C-063 references in PR #366",
    "CodeRabbit findings at heads 412f20c, 78c8e97, and 5f0e2b9",
    "Exact-head CodeRabbit clean review at https://github.com/open-horizon-labs/unified-hifi-control/pull/366#issuecomment-5139745082"
  ],
  "findings": [
    {
      "severity": "resolved",
      "summary": "A numeric-position proof could survive a semantic list transition; replaced by SemanticFamily opening/closing enumeration proof."
    },
    {
      "severity": "resolved",
      "summary": "Rate proof could miss entry into [source] before and after GetModes; final modes fetch is followed by a new State read."
    }
  ],
  "contrary_case": [
    "Another controller can still change mode immediately after the last State reply.",
    "An enumeration can move out and back before the closing fingerprint read.",
    "A lost reply means delivery may be ambiguous even when later readback differs."
  ],
  "required_actions": [],
  "dispositions": [
    "Return Ambiguous when a prior write may have reached the wire but final proof is unavailable",
    "Document the non-atomic boundary instead of claiming serializable control",
    "Bound correction to one retry so instability is observable"
  ],
  "residual_risks": [
    "Post-final-State external changes remain outside client-side atomic proof",
    "Device/version behavior still requires separate live qualification"
  ],
  "blockers": [],
  "authorization": "Dissent is closed for this exact draft head only; no merge, ready-for-review transition, live-appliance access, or API/MCP contract change is authorized."
}
```

muness:
```json
{
  "report": "uhc_workflow_report_v1",
  "report_number": 3,
  "stage": "execute",
  "role": "review",
  "head": "8970fd9cb93d537b6e7e234aef55c6e289a323bb",
  "verdict": "pass",
  "aim": "Verify that implementation and tests realize issue #347 without contract drift.",
  "scope_checked": [
    "src/adapters/hqplayer.rs",
    "Stateful HQPlayer fake and conformance regressions",
    "Cache-generation and projection proofs",
    "HTTP and MCP outcome conversion call sites",
    "API route and protocol-schema contracts"
  ],
  "out_of_scope": [
    "Dioxus UI behavior",
    "New adaptive payload schema",
    "Real HQPlayer daemon behavior",
    "Merge and deployment"
  ],
  "evidence": [
    "hqplayer_conformance passed 271 of 271 three consecutive exact-content runs",
    "cargo test --workspace --no-fail-fast passed 675 tests with 12 live tests ignored",
    "cargo fmt --check, cargo clippy -- -D warnings, and git diff --check passed",
    "api_contract, protocol_schema, protocol_integration, projection lint, and 21 cache-identity tests passed",
    "Superego exact-diff review found no blocker"
  ],
  "findings": [
    {
      "severity": "none",
      "summary": "The implementation is test-backed, preserves API and MCP contracts, and keeps semantic proof and cache ownership inside the adapter/aggregator architecture."
    }
  ],
  "contrary_case": [
    "An aggregate gate once returned nonzero with output suppressed; every gate passed immediately when isolated, and the full workspace rerun passed."
  ],
  "required_actions": [],
  "dispositions": [
    "Kept raw numeric set_filter compatibility path unchanged",
    "Added no routes, request fields, response fields, or MCP tools",
    "Used a deterministic mode-switch harness to reproduce both rate races RED before GREEN"
  ],
  "residual_risks": [
    "Superego observed one isolated config_form coverage failure that did not reproduce in the repeated exact-content or workspace runs",
    "No Dioxus dx build was possible because dx is absent, though no UI file changed"
  ],
  "blockers": [],
  "authorization": "Implementation may remain pushed on draft PR #366; no merge, live deployment, appliance contact, or API/MCP contract modification is authorized."
}
```

muness:
```json
{
  "report": "uhc_workflow_report_v1",
  "report_number": 4,
  "stage": "execute",
  "role": "dissent",
  "head": "8970fd9cb93d537b6e7e234aef55c6e289a323bb",
  "verdict": "pass_after_corrections",
  "aim": "Search the exact implementation for stale-state, stale-enumeration, retry, and false-success counterexamples.",
  "scope_checked": [
    "Decision-to-wire and proof ordering",
    "All SemanticFamily setter families",
    "Pair selection equal/split/missing states",
    "Wire-attempt bookkeeping across bounded retries",
    "Generation-fenced full and per-family cache publication"
  ],
  "out_of_scope": [
    "Unobservable daemon-internal timing",
    "Atomic operations absent from HQPlayer protocol",
    "Public contract expansion",
    "Live hardware matrix"
  ],
  "evidence": [
    "CodeRabbit exact-head review sequence 412f20c to 78c8e97 to 5f0e2b9 to 8970fd9",
    "Regression a_rate_write_that_enters_source_after_the_decision_is_not_applied",
    "Regression a_rate_proof_rechecks_state_after_fetching_modes",
    "Final exact-head CodeRabbit statement: no further issue in this follow-up"
  ],
  "findings": [
    {
      "severity": "resolved",
      "summary": "Stale semantic positions could return AlreadySet or Applied across a chain transition."
    },
    {
      "severity": "resolved",
      "summary": "A rate write entering [source] after decision could be acknowledged and falsely reported Applied."
    },
    {
      "severity": "resolved",
      "summary": "Reading State before GetModes left a second stale-state interval; State is now re-read after modes."
    }
  ],
  "contrary_case": [
    "Repeatedly adding another final read cannot eliminate a change immediately after that read.",
    "A closing enumeration verifies semantic namespace stability but cannot make configured mode and rate one atomic observation."
  ],
  "required_actions": [],
  "dispositions": [
    "Accepted the final non-atomic interval as a named residual",
    "Required wire-aware Ambiguous whenever earlier delivery cannot be disproved",
    "Stopped the correction loop at the protocol boundary after exact-head independent review"
  ],
  "residual_risks": [
    "External-controller races after the last State read",
    "ABA enumeration movement",
    "Unqualified HQPlayer versions may differ from hermetic fixtures"
  ],
  "blockers": [],
  "authorization": "Exact-head execution dissent is closed for draft review only; it does not authorize merge, live testing, release, or any public API/MCP change."
}
```

muness:
```json
{
  "report": "uhc_workflow_report_v1",
  "report_number": 5,
  "stage": "ship",
  "role": "review",
  "head": "8970fd9cb93d537b6e7e234aef55c6e289a323bb",
  "verdict": "draft_ready_not_merge_ready",
  "aim": "Assess whether PR #366 has a truthful, reproducible handoff for reviewers and the later beta path.",
  "scope_checked": [
    "Exact SHA and stacked-base declaration",
    "PR problem/solution/evidence narrative",
    "Test commands and totals",
    "Dependency and merge order",
    "API, MCP, appliance, and tooling boundaries"
  ],
  "out_of_scope": [
    "Merging PR #366",
    "Deploying a beta build",
    "Live HQPlayer 6.0.2 qualification",
    "Dioxus WASM verification without dx"
  ],
  "evidence": [
    "Draft PR #366 head matches 8970fd9cb93d537b6e7e234aef55c6e289a323bb",
    "PR body records 271 HQPlayer and 675 workspace tests",
    "CodeRabbit exact-head clean comment https://github.com/open-horizon-labs/unified-hifi-control/pull/366#issuecomment-5139745082",
    "Working tree is clean after commit and push"
  ],
  "findings": [
    {
      "severity": "none",
      "summary": "The draft contains sufficient exact-head evidence for human review and preserves explicit non-authorization boundaries."
    }
  ],
  "contrary_case": [
    "A green hermetic suite is not a substitute for the planned live version matrix or beta-device testing.",
    "The GitHub CodeRabbit status says draft review skipped even though the explicitly requested exact-head comment review completed cleanly."
  ],
  "required_actions": [
    "Merge dependency PR #364/#341 first when separately approved",
    "Obtain explicit user approval before merge",
    "Run the available final CI and beta/live qualification in the later ship phase"
  ],
  "dispositions": [
    "Keep PR #366 draft",
    "Do not claim live qualification",
    "Carry dx absence and protocol atomicity as visible residuals"
  ],
  "residual_risks": [
    "Stacked-base changes can require revalidation",
    "No live appliance or WASM build evidence belongs to this head"
  ],
  "blockers": [
    "PR #366 is stacked on #341/#364",
    "User merge approval has not been given",
    "Live and beta qualification remain later program work"
  ],
  "authorization": "Authorized to publish this draft handoff and reports only; explicitly not authorized to merge, mark ready, deploy, contact the appliance, or change API/MCP contracts."
}
```

muness:
```json
{
  "report": "uhc_workflow_report_v1",
  "report_number": 6,
  "stage": "ship",
  "role": "dissent",
  "head": "8970fd9cb93d537b6e7e234aef55c6e289a323bb",
  "verdict": "withhold_merge",
  "aim": "Try to disprove that the current evidence is sufficient to ship or merge #347.",
  "scope_checked": [
    "Evidence provenance",
    "Draft-review status",
    "Dependency order",
    "Tooling omissions",
    "Live qualification boundary",
    "Rollback and false-success risk"
  ],
  "out_of_scope": [
    "Performing the blocked merge",
    "Changing the live HQPlayer host",
    "Creating a release artifact",
    "Expanding the feature set"
  ],
  "evidence": [
    "AGENTS.md requires CI, testing, user approval, and explicit merge direction",
    "PR #366 remains draft and stacked on #341",
    "dx is unavailable in this environment",
    "All tests used the loopback stateful fake; no appliance was contacted"
  ],
  "findings": [
    {
      "severity": "blocker_to_merge",
      "summary": "The stack dependency and explicit user approval gates are not satisfied."
    },
    {
      "severity": "blocker_to_release",
      "summary": "Live HQPlayer and beta-build qualification are not part of this exact-head evidence."
    }
  ],
  "contrary_case": [
    "The implementation may be correct while packaging, version-specific daemon behavior, or the stacked rebase still fails.",
    "Marking the PR ready now would blur hermetic correctness with operational qualification."
  ],
  "required_actions": [
    "Keep the PR draft",
    "Revalidate after dependency integration",
    "Perform authorized live and beta testing before release",
    "Request explicit user approval for each merge"
  ],
  "dispositions": [
    "No auto-merge",
    "No force push or history rewrite",
    "No appliance mutation",
    "No unsupported success claim"
  ],
  "residual_risks": [
    "Version-specific native-protocol behavior",
    "Dioxus build-tool gap",
    "Non-atomic HQPlayer setting proof under external controllers"
  ],
  "blockers": [
    "Dependency #341/#364",
    "No explicit merge authorization",
    "No live beta evidence"
  ],
  "authorization": "Merge and release authorization is withheld. Only the already-pushed draft branch, PR evidence, and these reports are authorized."
}
```

muness:
Superseded by #428, the single v3-based consolidated HQPlayer direct-control beta PR. Review history and evidence remain here; please use #428 for further review and the QNAP x64 test artifact.

Issue #328, opened 2026-07-29
HQPlayer direct zone: deliver now-playing, transport, seek, and safe volume


muness:
OH: 80222d6d

Parent epic: #313

## Problem Statement

**Current framing:** Route play/pause/next/previous and volume commands to HQPlayer and fill the metadata fields.

**Reframed as:** Direct HQPlayer users need a complete everyday playback zone on UHC surfaces, but current HQPlayer zones are discovered only opportunistically, publish blank metadata/default volume capability, receive no continuous updates, and control-device commands can fall through to Roon.

**The shift:** From isolated HQPlayer command methods to a truthful direct zone whose capabilities, state, routing, and feedback work end to end.

### Constraints
- **Hard:** Explicit backend routing; full metadata from observed status; decimal-dB range/step; allowed controls reflect real state; direct zone coexists with linked DSP enrichment; aggregator is the only surface state source.
- **Soft:** Album-art support and queue visibility in the first slice.

### What this framing enables
Michael's relay-PC-free knob workflow, consistent web/MCP direct control, and a usable first adaptive HQPlayer layout.

### What this framing excludes
Playlist/library management unless needed to make advertised transport truthful, synthesized metadata when HQPlayer does not provide it, and unsafe percent-to-dB assumptions.

## Acceptance criteria

- [ ] Parse and publish title, artist, album, composer, URI, duration/position, playback state, and available transport capabilities when supplied.
- [ ] Route play, pause, stop, previous, next, seek, set/up/down volume, and mute only to the selected HQPlayer instance.
- [ ] Model volume as decimal dB with discovered min/max/step/fixed/adaptive semantics and clamp/validate safely.
- [ ] Current state is observed and republished within the agreed freshness bound, including changes made by another controller.
- [ ] Empty/stopped metadata and capability transitions clear correctly rather than retaining stale data.
- [ ] Knob-compatible now-playing and controls work without backend-specific route bypasses.
- [ ] Linked Roon/LMS zones retain DSP enrichment without duplicate or ambiguous control routing.
- [ ] Tests start from ESP32/iOS/client expectations and fail before implementation changes.
- [ ] Album art is either supported through a safe authenticated/proxied path or explicitly represented unavailable.

## Dependencies

Blocked by the protocol conformance harness, continuous lifecycle #162, and HQPlayer declarative producer.
Related: #208 for instance selection.

## HQPTuner salvage additions (2026-07-30)

[...]

PR #382, opened 2026-07-31
feat(#329): add revision-fenced HQPlayer immediate commands

muness:
Refs #329

OH: 80222d6d

## Outcome

Establish one internal, revision-fenced HQPlayer immediate-command path that later web, MCP, and device consumers can share without calling adapters directly or mistaking request delivery for native application.

This PR implements the command foundation for semantic pipeline controls: mode, filter 1x, filter Nx, shaper, and rate. Direct-zone transport, volume, now-playing, and consumer surfaces remain intentionally stacked in #328 and #331.

## What changed

- advisory preflight against one admitted adaptive snapshot
- producer-coordinator reservation with correlation/fingerprint idempotency
- explicit Pending publication before native I/O
- bounded command actor with graceful drain and caller-cancellation ownership
- exact producer/run/endpoint/transport/generation dispatch fencing
- contract-free typed native receipts and same-session readback evidence
- receipt precedence over racing observations and newer producer epochs
- append-only operation history merged into every observation
- bounded terminal, correlation, unresolved, and retired-instance retention
- explicit Expired/ReplanRequired publication before definitive retirement
- coordinator-owned retry after retirement intent commits; pre-accept failures still roll back
- stopped-manager retirement authority, including unresolved audit publication
- logical producer epoch continuity across same-name remove/re-add
- replacement isolation while an older lifetime is retiring
- explicitly handled coordinator replies plus macro-aware ignored-send enforcement
- ADR 004 updated with the final causality and lifecycle contract

No public route or payload contract changed.

## Review and dissent remediation

Successive independent review/dissent rounds identified and drove regressions for:

- observation erasing an in-flight native receipt
- unbounded retired identities and unresolved audit loss
- producer-epoch advancement racing an exact fenced receipt
- stop/start losing the active adaptive run
- expiration-frontier pruning behind newer terminal history
- manager rollback conflicting with coordinator-owned retirement retry
- stopped removal leaving an adaptive LastKnown ghost
- same-name replacement resetting its producer epoch
- replacement observations inheriting a retiring ledger
- admitted receipt truth being reported as unpublished
- ignored reply sends hidden inside macro bodies

Final review: PROCEED. Final dissent: PROCEED.

## Verification

- `cargo test --all-features`
- HQPlayer producer unit suite: 192 passing in the full library run
- HQPlayer lifecycle integration suite: 45 passing
- ignored-send lint: 6 passing
- API contract: 2 passing
- architecture/adaptive publication lints passing
[...]

Issue #313, opened 2026-07-29
[Epic] Deliver complete HQPlayer controls safely across UHC surfaces


muness:
OH: 80222d6d

Parent program: #310

## Problem Statement

**Current framing:** Add HQPlayer transport, volume, settings, and extra buttons to the knob, web page, and MCP tools.

**Reframed as:** HQPlayer users need direct playback and broad tuning from the UHC surfaces suited to each task, but current surfaces expose inconsistent subsets, can route commands incorrectly, cannot describe valid choices dynamically, and do not share verified apply/pending/failure semantics.

**The shift:** From surface-by-surface HQPlayer features to projections of one authoritative state/control document, with safe execution and explicit persistence behavior.

### Constraints
- **Hard:** Direct transport/volume/metadata must work from the control device; live and persistent settings are distinct; disruptive changes disclose restart behavior; persistent writes require backup/readback/recovery; volume is bounded and unit-aware; commands route by explicit backend identity; every advertised control works as documented.
- **Soft:** Which advanced controls appear on small devices, exact web grouping, MCP tool granularity, and initial version/edition coverage.

### What this framing enables
A simple knob experience for immediate listening, a broad web tuning surface, self-describing MCP control, staged live/restart changes, and reliable recovery without backend-specific behavior being reinvented per client.

### What this framing excludes
Raw protocol fields in firmware, pretending Desktop supports Embedded persistence, shipping controls that cannot be verified, and requiring full feature parity on every surface.

## Outcomes

- HQPlayer direct zones provide current metadata, transport, seek, and safe decimal-dB volume.
- Modes, rates, filters, shapers, dither, adaptive output, and other discoverable live settings use verified apply semantics.
- Embedded persistent settings use backup, restore, restart awareness, readback, and recovery.
- Web, MCP, and device surfaces consume the shared adaptive-control model at appropriate depths.
- Applying, stale, unavailable, disconnected, and failed states are visible and actionable.
- A real-device/version matrix validates end-to-end behavior, including Michael's ESP32-S3 use case.

## Existing work

- Incorporate #209 for MCP discovery and broader controls.
- Incorporate #222 for applying/pending UX.
- Relate #283/#288-290 and PR #291 for manifest/configurator consumption.

## Completion signal

Michael can remove the relay PC for daily direct control, while a power user can complete a listening/tuning session—including recovery from HQPlayer restart—without falling back to the stock UI or an unverified CLI.
## Issue graph

[...]

Issue #338, opened 2026-07-30
CI: restore the v3 lint baseline under Rust 1.97


muness:
OH: 80222d6d

## Problem

The `v3` branch's CI lint job fails under Rust 1.97 on an `unnecessary_sort_by` finding in `src/app/pages/zones.rs:269`. The file is untouched by HQPlayer protocol PR #337, but the inherited failure leaves that PR and subsequent program PRs unable to reach a fully green merge gate.

This is baseline maintenance, not part of #322's HQPlayer protocol scope. Keeping it separate preserves the evidence for #322 and prevents an unrelated source change from being smuggled into that PR.

## Desired outcome

The repository's documented CI clippy invocation passes on an otherwise unchanged `v3` branch, with behavior preserved.

## Acceptance criteria

- A regression check demonstrates the current `v3` lint failure before the source-only fix.
- The CI lint invocation passes with Rust 1.97.
- Zone ordering behavior is unchanged.
- No API routes, request schemas, or response payloads change.
- The fix is delivered independently of #337 and references this issue.
- #337 is rerun or rebased after this lands so its remaining red check can be evaluated honestly.

Related: #310, #322, PR #337

muness:
Follow-up class finding is now tracked separately in #340: reproducible blocking Clippy baseline, advisory latest-stable signal, and the missing CSS preparation in docker.yml lint. This does not broaden #338.

PR #391, opened 2026-07-31
feat(#328): make direct HQPlayer a truthful everyday UHC zone

muness:
## What this is

Makes a direct HQPlayer instance a truthful everyday UHC zone: observed now-playing metadata,
observed playback state, transport + seek, safe decimal-dB volume, and explicit backend routing —
with **no fallback to Roon or LMS**.

Stacked on #382 (`feat/issue-329-hqplayer-immediate-command`). **Do not merge.**

Refs #328 · OH: 80222d6d · Parent epic #313
Session record: [`.oh/hqplayer-direct-zone.md`](../blob/feat/issue-328-direct-hqplayer-zone/.oh/hqplayer-direct-zone.md)

## The defects this closes

Found by reading the code before choosing an approach; each is a specific line, not a category:

1. **Every HQPlayer control command was executed against Roon.** `knob_control_handler` dispatches
   `lms:`/`openhome:`/`upnp:` explicitly then *falls through* to `control_roon` — including for
   `hqplayer:` zone ids, prefix stripped.
2. **The HQPlayer adapter toggle did not filter HQPlayer zones** — the filter tested the prefix
   `hqp:`, zones are published as `hqplayer:`.
3. **Transport capability was asserted, not observed** — `is_seekable`, `is_next_allowed`,
   `is_previous_allowed` hard-coded `true` in every state.
4. **Output-domain facts were published as track metadata** — `TrackMetadata.bit_depth` from
   `Status.active_bits` (the DAC), `format` from `Status.active_mode` (which reads back the literal
   `[source]` under source-following), while the source-domain depth on the `metadata` child was
   dropped.
5. **A lost `VolumeRange` reply erased the volume capability** — the transient-failure fallback has
   `enabled: false`, which the zone projection maps to `volume_control: None`.
6. **Mute was a constant `false`**, though `VolumeMute` is a verified absolute move to the range
   floor and is therefore observable.
7. **A direct zone could be linked to itself as a DSP-enrichment target**, producing two control
   routes to one daemon.
8. **A missing volume value defaulted to `50.0`** — above every possible dB maximum, and invisible
   to the existing safety lint (which scans `src/adapters` only, for a different pattern).

## API contract impact

**None.** No route added, removed or changed; `tests/fixtures/api_routes.txt` untouched; no request
or response payload changes shape. New metadata reaches surfaces only through fields that already
exist on `NowPlaying`/`TrackMetadata`; new internal fields on `HqpStatus` are `#[serde(skip)]`,
following the precedent already set there for `title`/`artist`/`album`.

One acceptance criterion is **blocked** by that constraint rather than silently worked around — see
the Known limitations section of the session record for the exact additive change URI publication
would need, and why it was not made.

## Verification

Hermetic only — the live HQPlayer host was not touched from this branch. Wire/model fixtures from
the existing `tests/mock_servers/hqplayer` harness. Full gate transcript in the Execute-checkpoint
and Review comments below.

### DIFF
--- a/.oh/hqplayer-direct-zone.md
+++ b/.oh/hqplayer-direct-zone.md
@@ -252,6 +252,22 @@ Deliberately **not** touched: `tests/fixtures/api_routes.txt`, `src/api/mod.rs`,
 9. **Mode-switch qualification during playback (issue's beta/dev criterion) is not addressed.**
    It is a #347/#375 setter concern, not a direct-zone projection concern, and nothing here
    changes it.
+10. **`is_muted` has a known false positive at the range floor, and it is inherent to the protocol.**
+    A user who deliberately turns the level down to `min_db` is reported as muted, because
+    `VolumeMute` *is* "set the level to `min_db`" and there is no mute flag to distinguish the two —
+    the two situations produce byte-identical observations. Reporting mute anyway is the better trade
+    (at the floor both readings agree no sound is coming out, and the alternative makes the mute
+    button's effect invisible), but it is a real mislabel for anyone who uses the bottom of the range.
+    Raised by the dissent pass; only a protocol-level mute flag could close it.
+11. **The root-`samplerate`-is-the-source-rate fallback rests on one agreeing sample.** The single
+    corpus `Status` document carries root `samplerate="44100"`, child `samplerate="44100"`,
+    `active_rate="352800"`. Agreement cannot distinguish *"the root is the source rate"* from *"the
+    root happens to equal it on this material"*, and no document shows them diverging. Kept, because
+    it is daemon-supplied data rather than an invention and dropping it loses a fact the daemon sent —
+    but if the root actually follows the output rate, this publishes an output rate as the track's,
+    which is a relocated instance of the defect this issue fixes. **Falsifiable cheaply:** play
+    upsampled material and compare root against child `samplerate`. Left to #332's live qualification;
+    the fallback must be deleted if the check goes the other way.
 
 ---
 
--- a/.oh/hqplayer-direct-zone.md
+++ b/.oh/hqplayer-direct-zone.md
@@ -1,0 +1,273 @@
+# HQPlayer direct zone: now-playing, transport, seek, safe volume (#328)
+
+**OH:** 80222d6d · **Issue:** [#328](https://github.com/open-horizon-labs/unified-hifi-control/issues/328) ·
+**Parent epic:** #313 · **Base branch:** `feat/issue-329-hqplayer-immediate-command` (PR #382)
+
+### Inputs read before choosing an approach
+
+`.oh/adaptive-interaction-plane.md` is the **program session** for epic #313 and is *not* tracked on
+this branch — it lives in the main checkout and `.oh/adaptive-producer-contract.md:5` names it
+"read-only for this session". It was read there. Recorded because a document that is absent from the
+branch it governs is easy to skip, and its constraints are the ones this work is judged against:
+
+* *"Aggregator owns authoritative state"* and *"Devices never call adapters directly"* — hard.
+* *"Existing API changes require explicit approval"* — hard: *"New protocols need approved additive
+  routes/contracts."*
+* *"Provider capabilities differ … Do not manufacture parity"* — hard, and its success criterion is
+  the sentence this issue turns on: **"unsupported operations are never advertised."**
+* *"Project all advertised capabilities into web, HTTP, device UI, and MCP through one semantic
+  execution path; surface-specific subsets are allowed, contradictory semantics are not."*
+* Its chosen Option D (unified adaptive interaction plane) sequences *"HQPlayer proving producer"*
+  ahead of the client-facing plane. #328 is that proving step for the **legacy zone projection**;
+  #331 is where the adaptive surface arrives. This is why Candidate D below is rejected as
+  sequencing rather than direction.
+
+Also read and unedited: [`.oh/adaptive-producer-contract.md`](adaptive-producer-contract.md) (#323
+producer document v1 + compatibility policy), [`.oh/adaptive-publication.md`](adaptive-publication.md)
+(#324 bus + aggregator publication), [`.oh/hqplayer-evidence-ledger.md`](hqplayer-evidence-ledger.md)
+(#341 protocol evidence authority), `docs/ARCHITECTURE.md`, `AGENTS.md`.
+
+---
+
+## Aim
+
+A direct HQPlayer user picks a `hqplayer:` zone on a knob, in the web UI, or through MCP and it
+behaves like every other everyday zone: it says what is playing, it says what it can do, transport
+and volume land on *that* daemon, and what it reports keeps up with the daemon even when somebody
+else is driving.
+
+The behaviour change to look for: a user stops keeping a Roon or LMS zone selected "because the
+HQPlayer one doesn't really work", and stops reaching for the relay PC to change the level.
+
+Not the aim: making HQPlayer a library browser, and not making a zone *look* complete when the
+daemon has not said enough to make it complete.
+
+---
+
+## Problem Statement
+
+**Reframed (from the issue):** direct HQPlayer users need a complete everyday playback zone on UHC
+surfaces. Today the zone exists but is not truthful, and the untruths are individually small and
+jointly disqualifying.
+
+What the code actually does, read before choosing an approach:
+
+| # | Defect | Location |
+|---|--------|----------|
+| 1 | **Every HQPlayer control command is executed against Roon.** `knob_control_handler` dispatches `lms:`, `openhome:`, `upnp:` explicitly and then *falls through* to `control_roon` for everything else — including `hqplayer:`. The Roon adapter is handed `hqplayer:Living Room` with the prefix stripped, and either fails or, worse, matches an unrelated Roon zone id. | `src/knobs/routes.rs:533-554` |
+| 2 | **The HQPlayer adapter toggle does not filter HQPlayer zones.** The zone filter tests the prefix `hqp:`; zones are published as `hqplayer:`. So the test never matches, HQPlayer zones land in the `_ => true` arm, and disabling the adapter in settings leaves its zones listed. | `src/knobs/routes.rs:165` |
+| 3 | **Transport capability is asserted, not observed.** `is_seekable`, `is_next_allowed`, `is_previous_allowed` are hard-coded `true` for every HQPlayer zone in every state, including stopped with nothing loaded. | `src/adapters/hqplayer.rs:7174-7179` |
+| 4 | **Output-domain facts are published as track metadata.** `TrackMetadata.bit_depth` is filled from `Status.active_bits` (the DAC's output depth) and `format` from `Status.active_mode` (the output mode, which under `[source]` reads back the literal string `[source]`). The source-domain depth the daemon *does* supply, on the `metadata` child, is dropped. | `src/adapters/hqplayer.rs:7146-7155` |
+| 5 | **A lost `VolumeRange` reply erases the volume capability.** `ensure_volume_range` answers a transient failure with `VolumeRange::default()`, whose `enabled` is `false`; `hqp_status_to_zone` maps `!enabled` to `volume_control: None`. One dropped reply and a zone that has a working −60…0 dB control reports as fixed-volume. | `src/adapters/hqplayer.rs:3255-3258`, `7120-7133` |
+| 6 | **Mute is published as a constant.** `is_muted: false`, with the comment "HQPlayer doesn't report mute separately". It does not report a *flag*; `VolumeMute` is a verified absolute move to the range floor (HQP, #322 live), so mute is observable — as the level sitting at `min_db`. | `src/adapters/hqplayer.rs:7127` |
+| 7 | **A direct zone can be linked to itself as a DSP enrichment target.** `link_zone` accepts any `zone_id` string, so `hqplayer:A` → instance `A` is storable, after which the zone carries a `dsp` block pointing at its own control path: two routes, one daemon, no disambiguation. | `src/adapters/hqplayer.rs:7783` |
+| 8 | **A missing volume value defaults to 50.** The `vol_abs` arm ends `.unwrap_or(50.0)`. For a dB zone +50 dB is above every possible maximum, and the existing safety lint does not see it — that lint scans `src/adapters` only, and matches the pattern `.value.unwrap_or(50.0)`, not `.as_f64()).unwrap_or(50.0)`. | `src/knobs/routes.rs:610`, `tests/architecture_lint.rs:787-829` |
+
+What is **already** right and must not be re-solved: the 2 s coherent-observation poll loop
+(#162/#369) republishes the zone continuously, so "changes made by another controller" is a
+*verification* obligation here, not an implementation one; `knob_now_playing_handler` already reads
+zone state from `ZoneAggregator` and no surface reads adapter state; `HqpStatus.volume_db` and
+`VolumeRange.{min,max,step}_db` already carry exact decimal dB internally (#322/#347).
+
+### Constraints treated as real
+
+* **No public HTTP/MCP endpoint, request schema, or response schema may change** without explicit
+  user approval, and `tests/fixtures/api_routes.txt` may not be edited. This is the binding
+  constraint on the design and it is what forces the `serde(skip)` internal-field approach below.
+* **No surface may query an adapter for state** (`docs/ARCHITECTURE.md`,
+  `tests/architecture_lint.rs`). Commands to adapters from `_control_handler` are the established
+  and allowed pattern; state reads are not.
+* **No unevidenced protocol claim.** `.oh/hqplayer-evidence-ledger.md` is the authority, and a
+  grep of the whole repository finds **no** evidence for `composer`, `performer`, `albumartist`,
+  `genre`, `date` or `uri` as `Status/metadata` attribute *spellings*. Only `artist`, `album`,
+  `song`, `samplerate`, `bits`, `channels`, `bitrate` are in the corpus.
+* **Hermetic only.** No connection to a live HQPlayer host from this branch.
+
+---
+
+## Solution Space
+
+### Candidate A — Band-aid: add a `hqplayer:` arm to the router
+
+Add the missing `else if req.zone_id.starts_with("hqplayer:")` branch, mapping actions onto the
+adapter methods that already exist.
+
+* Solves the stated problem: **the routing half only** (defect 1). Every other defect is in the
+  *projection* — the zone would still advertise seek on a stopped zone, still publish the DAC's bit
+  depth as the track's, still lose its volume control on a dropped reply.
+* Cost: very low. Second-order: the worst outcome available here, because commands start landing on
+  the right daemon while the capability flags stay false. A knob that now really does send `Seek` to
+  HQPlayer, told by the same server that seek is always allowed, is a *regression* in user-visible
+  correctness even though every individual change is an improvement.
+
+### Candidate B — Local optimum: fix the router and the projection together
+
+Add the `hqplayer:` arm **and** make `hqp_status_to_zone` derive every capability, level and
+metadata field from the observation instead of asserting it: capability from observed state,
+source-domain metadata from the `metadata` child, mute from the floor comparison, volume capability
+retained across a transient read failure.
+
+* Solves the stated problem: **yes**, for every acceptance criterion whose data has somewhere to go
+  in the existing schemas.
+* Cost: medium, concentrated in two files plus a link-service guard.
+* Second-order: the router and the projection have to agree about what "allowed" means, and nothing
+  structural makes them agree — they are two functions in two modules. Mitigated by deriving the
+  router's refusals from the *aggregator's published zone*, so the flag the client was given is
+  literally the flag the command is checked against. That is not a new abstraction; it is the
+  aggregator already being the single source of truth.
+
+### Candidate C — Reframe: a typed direct-zone projection module
+
+Extract an `hqplayer::direct_zone` module owning one function
+`observation -> (Zone, allowed_actions)`, with `AllowedActions` a type the router consumes, so
+"advertised" and "permitted" are one value by construction rather than by agreement.
+
+* Solves the stated problem: yes, and closes B's residual structurally.
+* Cost: high. `AllowedActions` has to cross from `src/adapters` to `src/knobs`, and the only way to
+  carry it is on the `Zone` the aggregator stores — which is `BusEvent::ZoneDiscovered`'s payload,
+  serialised verbatim into `GET /events`. **So the reframe requires a public response-schema change
+  and is therefore not available without approval.** A parallel side-channel keyed by zone id would
+  avoid the schema change and reintroduce exactly the drift the reframe exists to remove, on a path
+  the aggregator does not own.
+* Rejected on the constraint, not on the merits. Recorded as the shape to revisit if #331 opens the
+  adaptive surface, where a typed capability set has a legitimate home.
+
+### Candidate D — Redesign: route direct HQPlayer transport through the #329 immediate-command actor
+
+Model transport and volume as adaptive commands and submit them to
+`HqpImmediateCommandService`, gaining revision fencing, correlation-id dedup, operation records and
+supersession for free.
+
+* Solves the stated problem: yes, and it is where this ends up eventually.
+* Cost: highest. The #329 vocabulary is `Mode`/`Filter1x`/`FilterNx`/`Shaper`/`Rate` — declarative
+  *settings* with a `desired`/`observed` lane and a readback. Transport has no desired lane (there
+  is nothing to stage), and its "readback" is a state transition, not a value comparison. Adding
+  transport to that vocabulary means extending the producer document's control vocabulary, which is
+  the #323 v1 contract under `.oh/adaptive-producer-contract.md` and its compatibility policy.
+* Second-order: it also inherits #329's *pending* work. Doing it here couples a user-visible
+  everyday-zone fix to an unlanded contract extension on a stacked branch. **Rejected as
+  sequencing, not as direction** — #331 is where the adaptive surface arrives, and this is the note
+  to read then.
+
+### Chosen: **B**, with C's invariant enforced by test rather than by type
+
+Level: **local optimum**. The reframe is the better design and the constraint that blocks it is a
+real one, so the honest move is to take B and pay for C's guarantee in verification: the router
+resolves the zone from `ZoneAggregator` and refuses on the *published* flag, and a test asserts the
+two cannot disagree by driving both through the same daemon observation.
+
+Consequences accepted:
+
+1. `Status/metadata` is parsed **generically** — every attribute of the child is read into a map,
+   and known keys are mapped onto typed fields. This makes no claim that any given spelling exists:
+   an attribute is published when the daemon supplies it and absent otherwise. It is the only
+   parser shape that satisfies "parse composer/genre where present" without adding an unevidenced
+   protocol claim to a repository whose ledger lint exists to prevent exactly that.
+2. New metadata reaches surfaces only through fields that **already exist** on `NowPlaying` /
+   `TrackMetadata` (`composer`, `genre`, `sample_rate`, `bit_depth`, `bitrate`). New internal fields
+   on `HqpStatus` are `#[serde(skip)]`, following the precedent already set for `title`/`artist`/
+   `album` on that struct. No response payload changes shape.
+3. **URI cannot be published.** No existing field on any published type can carry it. See Known
+   limitations for the exact approved change it would need.
+
+---
+
+## Execute
+
+**Status:** complete (draft PR open, not merged) · **Updated:** 2026-07-31
+
+Test-first throughout: every behavioural change below has a client-expectation test that was run
+and observed to fail against the unmodified tree before the implementation existed. RED evidence is
+in [Verification evidence](#verification-evidence).
+
+### Client-first failing tests (new file `tests/hqplayer_direct_zone.rs`)
+
+The file is organised by the client whose expectation it encodes, not by the module it exercises.
+
+| Group | Client expectation |
+|-------|--------------------|
+| `routing` | A knob that posts `{zone_id: "hqplayer:…", action: …}` reaches *that* HQPlayer daemon and nothing else — asserted on the mock daemon's received-request log, and by a Roon adapter that would have to have been called and was not. An unknown *prefixed* zone id is refused rather than silently sent to Roon; a legacy unprefixed id still goes to Roon. |
+| `capability` | The flags the client is given match the state the daemon is in: no seek without a duration, no next/previous with nothing loaded, no pause when not playing. |
+| `metadata` | Title/artist/album/composer/genre appear when the daemon supplies them; source-domain rate and depth come from the source, not from the DAC; a malformed `metadata` child loses the metadata and keeps the transport state. |
+| `stale_state` | Stopping clears now-playing, seekability and next/previous; an observed fixed-volume transition clears the volume control; a *transient* read failure does not; zone identity survives a reconnect. |
+| `volume` | Decimal dB survives end to end; a missing or unparseable level is refused rather than defaulted; out-of-range levels clamp to the observed range; the daemon is never sent a volume command when it has no volume control. |
+| `reconciliation` | A change made by another controller is republished to the aggregator without any UHC command being issued. |
+| `dsp_boundary` | A direct zone carries no `dsp` block; linking a `hqplayer:` zone is refused; a linked Roon zone keeps its enrichment. |
+
+### Implementation boundary
+
+Four files. Nothing else is touched.
+
+* `src/adapters/hqplayer.rs` — generic `metadata`-child attribute parse; `HqpStatus` internal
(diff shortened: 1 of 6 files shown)


muness:
## (a) Solution Space decision — #328 direct HQPlayer zone

**Status:** decided, implementation not yet started at this comment.
**Chosen:** Candidate **B** (local optimum) — fix the router *and* the zone projection together —
with Candidate C's invariant enforced by test rather than by type.

### Exact scope

Four files, and the list is exhaustive:

| File | Change |
|------|--------|
| `src/adapters/hqplayer.rs` | generic `Status/metadata` child attribute parse; `#[serde(skip)]` internal fields on `HqpStatus`; `hqp_status_to_zone` derives capability/metadata/mute/volume from the observation; `ensure_volume_range` retains the last observed capability across a transient failure; `HqpZoneLinkService::link_zone` refuses `hqplayer:` ids |
| `src/knobs/routes.rs` | `hqplayer:` routing arm; prefix fallthrough to Roon closed; `hqp:` → `hqplayer:` adapter-filter fix; no `dsp` block on direct zones |
| `tests/hqplayer_direct_zone.rs` | new, client-expectation-first |
| `tests/volume_safety.rs` | lint extension over the direct-zone volume path |

Deliberately untouched: `tests/fixtures/api_routes.txt`, `src/api/mod.rs`, `src/mcp/mod.rs`,
`src/bus/events.rs`, `src/adaptive/**`, `src/producers/**`.

### Candidates and why they lost

**A — Band-aid: add the missing `hqplayer:` router arm only.** Rejected, and rejected as
*actively harmful* rather than merely insufficient. It closes defect 1 and leaves every projection
defect standing, so a knob would start genuinely sending `Seek` to HQPlayer while the same server
keeps telling it seek is always allowed. Each individual change is an improvement and the
combination is a user-visible regression.

**C — Reframe: a typed `direct_zone` projection returning `(Zone, AllowedActions)` so "advertised"
and "permitted" are one value by construction.** The better design. **Rejected on the API
constraint, not on the merits:** the only way to carry `AllowedActions` from `src/adapters` to
`src/knobs` is on the `Zone` the aggregator stores, and that type is `BusEvent::ZoneDiscovered`'s
payload, serialised verbatim into the public `GET /events` SSE stream. So the reframe *requires* a
response-schema change and is unavailable without approval. A side-channel keyed by zone id would
dodge the schema change and reintroduce precisely the drift the reframe exists to remove, on a path
the aggregator does not own. Recorded as the shape to revisit under #331.

**D — Redesign: route transport through #329's immediate-command actor**, gaining revision fencing,
correlation dedup, operation records and supersession. This is where it ends up eventually.
Rejected as **sequencing**: that vocabulary is `Mode`/`Filter1x`/`FilterNx`/`Shaper`/`Rate` —
declarative settings with a `desired`/`observed` lane and a value readback. Transport has no desired
lane (there is nothing to stage) and its readback is a state transition, not a value comparison, so
adding it means extending the #323 producer-document control vocabulary under
`.oh/adaptive-producer-contract.md`'s compatibility policy. That couples a user-visible everyday-zone
fix to an unlanded contract extension on an already-stacked branch.

### The residual B carries, and how it is paid for

B leaves the router and the projection as two functions in two modules with nothing structural
making them agree about "allowed". Mitigation is not a new abstraction: the router resolves the zone
from `ZoneAggregator` and refuses on the **published** flag, so the flag the client was handed is
literally the flag the command is checked against, and a test drives both through one daemon
observation to assert they cannot disagree.

### Evidence / tests

No tests yet — this comment is the decision, and the next one is the RED transcript. The decision
rests on reading `src/knobs/routes.rs`, `src/adapters/hqplayer.rs`,
`src/aggregator.rs`, `src/bus/events.rs`, `tests/architecture_lint.rs`, the
`tests/mock_servers/hqplayer` harness, `.oh/adaptive-interaction-plane.md` (program session, main
checkout — not tracked on this branch), `.oh/adaptive-producer-contract.md`,
`.oh/adaptive-publication.md` and `.oh/hqplayer-evidence-ledger.md`.

### Contract / API impact

None intended, and one AC is expected to be **blocked** by that: `Status/metadata` URI has no field
on any published type. It will be reported with the exact approved change rather than worked around.

### Live-test status

None, by constraint. Hermetic wire/model fixtures only; the 6.0.2 rig is not contacted from this
branch.

### Unresolved risks

1. Attribute *spellings* for `composer`/`genre` have **no** evidence anywhere in this repository —
   a grep for them across `.oh/`, `docs/` and the conformance suite returns nothing. Mitigation: the
   parser reads the child's attributes **generically** and publishes what is present, so it makes no
   protocol claim that any spelling exists. This is the design consequence of the ledger discipline,
   not a workaround around it.
2. `play_pause` and relative volume resolve against aggregator state, which can be one poll interval
   (2 s default) stale. A synchronous daemon pre-read would fix it and is forbidden — surfaces may
   not read adapter state.

### Next action

Write `tests/hqplayer_direct_zone.rs` from ESP32/iOS client expectations, run it against the
unmodified tree, and post the RED transcript with the implementation boundary as comment (b).

muness:
## (b) Execute checkpoint — RED transcript, implementation boundary, gates

**Status:** implementation complete, all gates green. Head `8d54b98`, base `a178b76`.

### Client-first RED, before any behaviour existed

Commit [`74b24d6`](https://github.com/open-horizon-labs/unified-hifi-control/commit/74b24d6) contains
the tests, the two debug-only seams, and the serde-skipped `HqpStatus` fields the tests speak in
terms of — **and no behaviour**. Run against that tree:

```
test result: FAILED. 33 passed; 19 failed; 0 ignored
```

The 19, and the point of separating them from the 33 that passed: the passing ones are the
regression guards (a linked Roon zone keeps its DSP block; a loaded zone advertises transport) which
had to be green from the start or they would prove nothing later.

```
a_direct_zone_cannot_be_its_own_dsp_enrichment_target
a_missing_or_unparseable_level_is_refused_never_defaulted
a_stopped_empty_zone_advertises_no_seek_next_or_previous
a_transient_volume_read_failure_retains_the_last_observed_capability
an_unknown_instance_is_refused_rather_than_redirected
an_unknown_prefix_is_refused_but_a_legacy_unprefixed_id_still_means_roon
decimal_db_survives_the_whole_round_trip
mute_is_routed_and_reported_from_the_observable_floor
out_of_range_levels_clamp_to_the_observed_range
play_pause_resolves_against_the_published_state
seek_is_refused_when_the_published_zone_is_not_seekable
stop_next_previous_and_seek_all_route_to_the_daemon
stopping_clears_the_track_and_the_capabilities_that_depended_on_it
supplied_metadata_attributes_are_published_and_absent_ones_are_not_invented
the_published_track_depth_is_the_source_depth_not_the_output_stage
the_zone_filter_matches_the_prefix_hqplayer_zones_are_actually_published_with
track_metadata_is_the_source_domain_never_the_output_stage
transport_reaches_the_selected_hqplayer_instance_and_not_roon
volume_commands_are_refused_when_the_zone_advertises_no_control
```

Three of those failure messages are worth quoting, because they are the defect stated by the system
itself rather than by me:

```
---- transport_reaches_the_selected_hqplayer_instance_and_not_roon ----
assertion `left == right` failed: pause on a direct HQPlayer zone must succeed;
  body={"error":"Not connected to Roon"}
```
A `hqplayer:` zone id, and Roon answering.

```
---- mute_is_routed_and_reported_from_the_observable_floor ----
body={"error":"Unknown action: mute"}
```

```
---- a_transient_volume_read_failure_retains_the_last_observed_capability ----
a dropped VolumeRange reply is not evidence that the daemon became fixed-volume …
  zone=Zone { … volume_control: None, … }
```

### GREEN

```
cargo test --test hqplayer_direct_zone
test result: ok. 52 passed; 0 failed; 0 ignored
```

### Implementation boundary

Five files, exhaustively:

| File | Δ | What |
|------|---|------|
| `src/adapters/hqplayer.rs` | +327/−48 | generic `metadata`-child parse; derived capability/mute/clamped level; source-vs-output domains; endpoint-scoped volume-capability retention; `link_zone` guard |
| `src/knobs/routes.rs` | +239 | `control_hqplayer`; closed prefix fallthrough; `hqp:`→`hqplayer:`; no `dsp` on direct zones |
| `tests/hqplayer_direct_zone.rs` | new, 52 tests | client expectations |
| `tests/volume_safety.rs` | +111 | direct-zone volume-path lint |
| `tests/hqplayer_conformance.rs` | ±1 test split | see below |

Untouched, deliberately: `tests/fixtures/api_routes.txt`, `src/api/mod.rs`, `src/mcp/mod.rs`,
`src/bus/events.rs`, `src/adaptive/**`, `src/producers/**`.

### One existing test asserted the opposite of an acceptance criterion

`an_unresponsive_volume_range_still_yields_a_fixed_volume_pipeline` (#322) asserted
`pipeline.volume.is_fixed` after a lost `VolumeRange` reply. #328's criterion is *"when volume is
unavailable, retain the control/range … rather than erasing the capability"*. Those are in direct
conflict, so I did not quietly delete it.

Reading it closely, **its setup contradicted its own docstring**. The docstring premise is a daemon
*"never answering `VolumeRange`"*; the setup used `Harness::connected_then_policy`, which connects
successfully first — observing a working −60…0 dB control — and only then silences the element. It
was asserting that losing one reply about an observed capability proves the daemon has none. Its
stated concern, *"rather than invent variable-volume bounds"*, is not violated by retention: the
retained bounds are the observed ones.

So it is **split**, and both halves are pinned:

* `a_never_answered_volume_range_still_yields_a_fixed_volume_pipeline` — the docstring's actual
  premise, with a policy silent from the first request. Keeps `is_fixed`, and keeps the
  CodeRabbit-found transport-generation-fence and 4-families assertions its `client-red` label is
  about.
* `an_observed_volume_capability_survives_a_later_unanswered_volume_range` — the observed-then-lost
  case, asserting retention.

Both were verified red against the pre-change implementation by reverting the fallbacks and
re-running. The two fallbacks turned out to be **independently** covered, which is itself worth
recording: reverting only the handshake copy left the conformance test green (it exercises the poll
copy), and reverting only the poll copy left the direct-zone test green. Fixing one and assuming the
other was the mistake that made my first attempt at this look finished while the defect was fully
intact.

### Gates

| Gate | Command | Result |
|------|---------|--------|
| Targeted | `cargo test --test hqplayer_direct_zone` | **52 passed, 0 failed** |
| Conformance | `cargo test --test hqplayer_conformance` | **290 passed, 0 failed** |
| Client harness | `cargo test --test client_harness` | **passed, 0 failed** |
| Volume safety | `cargo test --test volume_safety` | **16 passed** |
| API contract | `cargo test --test api_contract` | **2 passed** — `api_routes_match_contract` ok |
| Whole suite | `cargo test --all-features` | **~1,340 passed, 0 failed** across 28 targets |
| Clippy | `cargo clippy --all-targets --all-features -- -D warnings` | see note |
| Format | `cargo fmt --all --check` | **clean** |
| Release UI | `dx build --release --platform web --features web` | **Client + Server build completed successfully** |

**Clippy, stated precisely.** The gate is **already red on the base branch**: 397 errors at
`a178b76`, all `unwrap`/`expect`/`panic`-in-test findings that #338 exists to clear. My change
introduced one new finding (a `loop` that should be `while let`), which is fixed. Findings in the
two source files I touched: `src/knobs/routes.rs` **0**; `src/adapters/hqplayer.rs` **3**, all three
pre-existing in its `#[cfg(test)]` module and none on a line I edited. I am not claiming a green
clippy run, because there isn't one to claim.

**Release UI build**, also stated precisely: it initially failed with `Missing rust target
wasm32-unknown-unknown`, which is an environment fault and not a code one — `sccache` corrupts
`dx`'s rustc probe, and `/opt/homebrew/bin/rustc` (1.93.1, no wasm std) shadows the rustup toolchain
that has it. With `RUSTC_WRAPPER=` and the rustup `stable` bin on `PATH` both halves build. Worth
noting either way that **both modules I changed are `#[cfg(feature = "server")]`** and so are absent
from the WASM build entirely.

### Contract / API impact

**None.** No route added, removed or changed; no request or response payload changes shape; every
new `HqpStatus` field is `#[serde(skip)]`, following the precedent already set there for
`title`/`artist`/`album`. `api_routes.txt` is byte-identical.

Two behaviours are additive *within* existing fields and I want them seen rather than buried:
`POST /control`'s `action` string now accepts `mute` and `seek` for `hqplayer:` zones (the field's
type is unchanged; per-zone action vocabularies already differ — OpenHome rejects `vol_up` today),
and refusals use `400`/`404`, both of which that endpoint already returns.

### Live-test status

**None.** Hermetic wire/model fixtures only; the HQPlayer host was not contacted from this branch,
per the task constraint. No claim here is a live claim.

### Unresolved risks

1. `play_pause` and relative volume resolve against aggregator state that can be one poll interval
   (2 s default) stale. Recorded, not mitigated: the fix would be a forbidden adapter state read on
   the command path.
2. `composer`/`genre`/`uri` attribute spellings have **no** evidence in this repository. The parser
   is generic over the child's attributes precisely so this is not a protocol claim — but it is also
   therefore unproven that any daemon populates them.
3. URI cannot be published (no field exists). See the PR body and the session record for the exact
   additive change; not made.

### Next action

Review and dissent passes, actively trying to falsify routing, stale-state, decimal-volume, the
metadata parser and recovery.

muness:
## (c) Review report — falsification of routing, stale-state, decimal volume, metadata parser, recovery

**Verdict:** the change is sound **after four fixes that this pass found**. All four were defects in
code written earlier in this same session, three of them introduced by the #328 fixes themselves.
Head `0a24dfc`, base `a178b76`.

**Scope reviewed:** the complete diff (`66c7ef2..0a24dfc`) — `src/adapters/hqplayer.rs`,
`src/knobs/routes.rs`, `tests/hqplayer_direct_zone.rs`, `tests/volume_safety.rs`, the one
`tests/hqplayer_conformance.rs` test split, `.oh/hqplayer-direct-zone.md`.

Method: adversarial probes appended to the suite, each written to *break* a claim rather than confirm
it, run, then either promoted to a permanent test (a hit) or discarded with the reason recorded (a
miss). Misses are listed too — a review that only reports hits is not evidence of much.

### Findings — all four confirmed, fixed, and pinned

**1 · Decimal volume — the safety floor coarsened a finer reported step.** *Severity: high.*
The floor on the published step was applied as `.max(MIN_PUBLISHED_VOLUME_STEP_DB)` **over the
reported value**, so a daemon offering 0.1 dB was published as offering 0.5 dB. The most
embarrassing of the four: a fix for "never publish an unusable step" had quietly become "never
publish an accurate one", and a fine-step user feels it on the first turn of the knob. It also
*masked* finding 2 — the probe for float noise came back with a suspiciously clean `-23`, which is
what exposed this.
→ floor now applies only when nothing usable was reported ·
`a_finer_reported_step_is_not_coarsened_by_the_safety_floor`

**2 · Decimal volume — float representation error reached the wire.** *Severity: medium.*
`crate::bus::VolumeControl` carries level and step as `f32`. Widening them back to `f64` to compute a
relative target turned `-23.5 + 0.1` into `-23.399999998509884`, and `set_volume_db` formats with
`{}` — so that is the string the daemon would have received where the reference client sends `-23.4`.
→ `quantise_db` rounds to 0.01 dB before the wire: below audibility, below the finest observed step,
so it removes noise without quantising away a real distinction. Not a clamp, and it decides no bound.
· `a_computed_level_reaches_the_wire_without_float_noise`

**3 · Metadata parser — a raw `>` in an attribute value truncated the scan.** *Severity: medium.*
XML forbids `<` and `&` in attribute values but **permits a bare `>`**, so `song="a/>b"` is
well-formed. The plain search for the `metadata` child's `/>` terminator cut inside that value and
silently lost every attribute after it — in the probe, the title, the album *and* the source bit
depth, to a track name containing a slash before a bracket. Note the direction: this was introduced
by the very bound added to stop a malformed child swallowing `</Status>`.
→ quote-aware terminator search; the test pins **both** directions, so the fix cannot regress into
the original hazard · `a_raw_terminator_inside_an_attribute_value_does_not_truncate_the_metadata_scan`

**4 · Routing / stale state — transport accepted for a withdrawn zone.** *Severity: medium.*
`play`, `pause` and `stop` did not consult the published zone at all — only `next`/`previous` did —
so they answered `{"ok":true}` for a zone the aggregator had already withdrawn, purely because the
instance remained in the manager's map. A client that shows no such zone could still command it, and
the whole design premise here is that the published zone is the authority.
→ every arm now requires the published zone; absence is `404` ·
`transport_is_refused_for_a_zone_the_aggregator_has_withdrawn`

### Hypotheses that failed to falsify anything

* **`trim_start_matches("hqplayer:")` mishandling a colon in an instance name.** It does not:
  `hqplayer:a:b` → `a:b`, the correct key — and better than `PrefixedZoneId::raw_id()`, which
  truncates at the first colon. It *would* strip a repeated prefix, but an instance literally named
  `hqplayer:x` is not a reachable configuration.
* **An encoded `&gt;` in an attribute value truncating the scan.** It does not — the raw document
  contains `&gt;`, not `>`. Sharpening this probe to the *unencoded* form is what produced finding 3.
* **Stale-state: a stopped zone retaining capability.** Held under every transition probed
  (playing→stopped, variable→fixed volume, reconnect). The `ZoneDiscovered` full-snapshot replacement
  is what makes it hold, and that predates #328.
* **Recovery: the retained volume capability leaking across an endpoint change.** It does not —
  cleared in the `native_changed` branch of `configure`, verified by reading the path.

### The one place I changed an existing assertion, and why it is not me bending a test

`an_unresponsive_volume_range_still_yields_a_fixed_volume_pipeline` (#322) asserted
`pipeline.volume.is_fixed` after a lost `VolumeRange` reply — the direct opposite of #328's criterion
*"retain the control/range … rather than erasing the capability"*.

I did not delete it. Its **setup contradicted its own docstring**: the docstring premise is a daemon
*"never answering `VolumeRange`"*, but `Harness::connected_then_policy` connects successfully first —
observing a working −60…0 dB control — and only then silences the element. Its stated concern,
*"rather than invent variable-volume bounds"*, is not violated by retention, because the retained
bounds are the **observed** ones. So it was split, and both halves are pinned: the never-answered
case keeps `is_fixed`; the observed-then-lost case asserts retention. Both verified red against the
pre-change implementation, and the CodeRabbit-found transport-generation-fence and 4-families
assertions its `client-red` label is about survive in both.

That exercise also caught something I would otherwise have shipped as finished: **the two
`VolumeRange` fallbacks are independently reachable**. Reverting only the poll copy left the
conformance test green; reverting only the handshake copy left the direct-zone test green. My first
attempt fixed the poll copy alone and looked complete while the defect was fully intact — the
handshake copy is the one that actually fires under a lost reply, because the timeout discards the
socket and the reconnect re-runs the handshake.

### Evidence / tests

| Gate | Result |
|------|--------|
| Targeted | `hqplayer_direct_zone`: **56 passed, 0 failed** (52 + 4 review regressions) |
| Conformance | `hqplayer_conformance`: **290 passed, 0 failed** |
| Volume safety | **16 passed**; both new lints proven non-vacuous by mutation (injecting `unwrap_or(50.0)` fails them) |
| Whole suite | `cargo test --all-features`: **1,250 passed, 0 failed, 13 ignored**, 32 targets |
| API contract | **2 passed**; `api_routes.txt` byte-identical |
| Format | `cargo fmt --all --check` clean |
| Release UI | `dx build --release --platform web --features web`: **Client + Server build completed successfully** |
| Clippy | **not green, and not claimed green** — 397 pre-existing errors at base (#338 backlog). `src/knobs/routes.rs`: 0 findings. `src/adapters/hqplayer.rs`: 3, all pre-existing `#[cfg(test)]`, none on an edited line. One new finding introduced during implementation and fixed. |

### Contract / API impact

**None.** Re-verified after the review fixes: no route, no request or response payload shape change,
`api_routes.txt` untouched, all new `HqpStatus` fields `#[serde(skip)]`. `quantise_db` and
`element_open_tag_end` are private helpers.

### Live-test status

**None.** Hermetic wire/model fixtures only; the HQPlayer host was not contacted. No claim in this PR
is a live claim, and nothing here is user-testable-beta evidence.

### Unresolved risks

1. **One-poll-interval (2 s) staleness** on `play_pause` and relative volume. Confirmed reachable;
   deliberately not fixed — both available fixes are forbidden (a surface reading adapter state, or a
   daemon-side compare-and-set the protocol lacks).
2. **`composer`/`genre`/`uri` spellings are unevidenced.** Unfalsifiable hermetically. The parser is
   generic over the child's attributes precisely so this is not a protocol claim, but it is equally
   unproven that any daemon populates them.
3. **URI is not published** — no field exists (see PR body / Known limitations for the exact additive
   change). AC1 partially not met.
4. **Adaptive-volume mode is not modelled as a mode**, so a level the daemon moved on its own reads
   as a user's level. That issue criterion is **not met** and needs a field.
5. **Three review findings were self-inflicted.** The pattern is worth naming rather than filing
   away: each came from a *fix* that overreached its own justification — a floor that also became a
   ceiling on accuracy, a bound that also became a truncation. Worth weighting the next reviewer's
   attention toward newly-added guards rather than toward the code they guard.

### Next action

Dissent report: argue the strongest case *against* merging this as scoped.

muness:
## (d) Dissent report — the strongest case against merging this as scoped

**Verdict:** `READY_FOR_REVIEW`, **not** ready to close #328, and **not** user-testable beta.
Head `212f555`, base `a178b76`. Two dissent findings produced code-comment and session-record
changes; none produced a behaviour change, and I say why below rather than treating "no change" as
vindication.

**Scope of this pass:** argue against my own conclusions from the Review comment. Where I disagree
with the dissent I say so and give the reason, rather than logging an objection and moving on.

---

### D1 · The flagship metadata fix is verified against a projection function, not against any observed daemon shape · **strongest objection, partly conceded**

Defect 4 (output depth published as the track's) is the fix I have led with. But the wire fake
**cannot express it**: `render_status` derives `Status.active_bits`/`samplerate`/`bitrate` from the
very `metadata` child it derives the source values from, so on that wire the two domains can never
disagree. I had to add a debug seam and test the pure projection instead.

So the claim "the published track depth is the source depth" is proven about a function, and no
fixture in this repository shows a daemon whose source and output depths differ at all.

Worse, and this is the part I concede outright: the **root-`samplerate` fallback** is an inference
from a single sample where root and child *agree* (`44100`/`44100`, with `active_rate="352800"`).
Agreement cannot distinguish *"the root is the source rate"* from *"the root happens to equal it on
this material"*. If the root actually follows the output, that fallback publishes an output rate as
the track's — **a relocated instance of the exact defect this PR claims to fix**.

*Response:* conceded and recorded, not fixed. The fallback stays, because it is daemon-supplied data
rather than an invention and deleting it loses a fact the daemon did send — but the code comment now
states exactly what is and is not evidenced, and Known limitation 11 names the falsifying check (play
upsampled material, compare root against child `samplerate`) and says the fallback must be deleted if
it goes the other way. The honest position is that this is #332's to qualify, not that it is settled.
Extending the fake to separate the domains would make the fixture *more* expressive than any observed
daemon, which in this repository's evidence discipline is its own defect.

### D2 · `is_muted` mislabels a user who turns the volume all the way down · **conceded, unfixable**

I assert confidently in the code that mute is observable. It is — but so is "turned to the floor",
and they are the *same observation*. A user who deliberately sets −60 dB is now reported as muted.

*Response:* conceded; recorded as Known limitation 10. No implementation can close it — `VolumeMute`
*is* "set the level to `min_db`" and there is no wire flag. Reporting mute remains the better trade
(at the floor both readings agree no sound is coming out, and never reporting it makes the mute
button's effect invisible), but the previous version of my own comment presented this as cleanly
observable, which overstated it.

### D3 · Four of nine acceptance-criteria boxes are not fully met, so "owning #328 end-to-end" overstates the delivery · **conceded**

Not met: URI publication (AC1, no field exists); performer/album-artist/date (AC1 addendum, same
cause); adaptive-volume-mode-as-a-mode; mode-switch qualification during playback. That is a
meaningful fraction, and three of the four are blocked by the same API-approval constraint rather
than by effort.

*Response:* conceded, and this is why the readiness line says `READY_FOR_REVIEW` and not "closes
#328". The PR body and Known limitations name each gap and, for the blocked ones, the exact additive
change required. What I do **not** concede is that the constraint should have been worked around: the
instruction was to stop and report rather than silently change a schema, and one additive
`uri: Option<String>` on `NowPlaying` is a `GET /events` payload change however small it looks.

### D4 · The router/projection agreement is a test, not a type · **not conceded, but the residual is real**

Candidate C would have made "advertised" and "permitted" one value. I rejected it on the API
constraint and paid for its guarantee with one test. A future edit to either side can break the
invariant with everything still green except that test.

*Response:* not conceded as a reason to withhold the change — the alternative available today is the
status quo, where the two do not agree *at all* because the router routes to Roon. But the residual is
real and the mitigation is thinner than I would like. #331 is where it gets a type.

### D5 · The new withdrawn-zone `404` can make a recovering zone briefly uncommandable · **not conceded**

Review finding 4 now refuses transport when the aggregator has no zone. During a producer restart the
zone is withdrawn, so a user pressing play inside the recovery window gets `404` where before they
got… a command sent to Roon. Still, this is a *new* refusal on a path that previously "succeeded".

*Response:* not conceded. The prior behaviour was `{"ok":true}` for a command that reached the wrong
backend; a `404` naming the real state is strictly better. And a knob that shows no zone should not be
able to command one — that premise is what the whole capability design rests on.

### D6 · CI will prove nothing about this PR · **conceded, and it constrains what may be claimed**

The branch is feature-branch-based, so no CI workflow runs; green checks on this PR would be vacuous.
The clippy gate is red on the base (397 pre-existing errors, #338's backlog), so "clippy clean" is not
available as a signal either, and the release UI build only worked after routing around an `sccache`
and a shadowed-`rustc` fault on this host.

*Response:* conceded. Every gate result reported here was run locally and is stated with its
qualifications rather than as a pass. Specifically: I am **not** claiming a green clippy run, and I am
**not** claiming live verification.

### D7 · Three of four review findings were self-inflicted, which is evidence about the change's maturity, not just about the fixes · **conceded**

The step floor became a ceiling on accuracy; the terminator bound became a truncation; the capability
check covered two arms out of seven. Each was a fix overreaching its own justification, and each was
found only because a falsification pass was run at all.

*Response:* conceded, and the inference is the useful part: a reviewer should weight attention toward
the **newly added guards** rather than the code they guard. That is where every one of them was.

---

### Status against the required report fields

* **Verdict:** `READY_FOR_REVIEW`. Do not merge; do not close #328.
* **Exact scope:** `src/adapters/hqplayer.rs`, `src/knobs/routes.rs`, `tests/hqplayer_direct_zone.rs`
  (new), `tests/volume_safety.rs`, one test split in `tests/hqplayer_conformance.rs`,
  `.oh/hqplayer-direct-zone.md`. Nothing else.
* **Evidence:** 56 targeted / 290 conformance / **1,250 whole-suite, 0 failed**; RED recorded at
  `74b24d6` as 19 failures; both new lints proven non-vacuous by mutation; `dx build --release
  --platform web` client + server green; `cargo fmt --check` clean.
* **Contract/API impact:** none. `api_routes.txt` byte-identical; no payload shape change; new
  `HqpStatus` fields `#[serde(skip)]`.
* **Live-test status:** none, by constraint. No claim here is a live claim.
* **Unresolved risks/blockers:** D1 (root-samplerate inference, falsifiable in one live check), D2
  (mute at floor, unfixable), D3 (four AC boxes need approved additive fields), D4 (router/projection
  agreement is test-enforced), 2 s staleness on `play_pause` and relative volume, unevidenced
  `composer`/`genre`/`uri` spellings.
* **Next action:** human review. Then either approve the additive `NowPlaying.uri` field (which
  unblocks AC1) or accept it as deferred to #331, and schedule the D1 live check under #332.

muness:
## Opus stacked review

**Verdict: READY_FOR_REVIEW** — one material defect found and fixed on this branch.

### Reviewed range

`git diff feat/issue-329-hqplayer-immediate-command...feat/issue-328-direct-hqplayer-zone`
= `a178b76644e4dd79c73f30e5299d4bc36840b403...212f5551da3b2f83d6c433734b2613816c6ff6d0`
(6 files, +2699 / −44). Reviewed against this PR's declared base only.

Head after this review: **`71c02a6af568e27363f8990949a7512cb4ae07fc`** (was `212f555`). One commit
added; nothing rewritten, so every prior SHA, comment and review anchor still resolves.

Independent adversarial pass — no subagents, no `superego`/`ba`/`wm`, no live host contact.

### Finding: an unorderable `VolumeRange` panicked the polling worker

`min_db` and `max_db` are both `parse_attr_f64(...).unwrap_or_default()`, so nothing upstream
guarantees they are ordered or even finite:

- a daemon that reports `max` and **omits `min`** yields `min_db = 0.0` against a negative `max_db`;
- `"nan".parse::<f64>()` is a parse *success*, so `min="nan"` arrives as NaN.

`f64::clamp` is specified to panic on either shape — and it is an `assert!`, not a
`debug_assert!`, so it fires in release builds too.

This PR added `status.volume_db.clamp(vol_range.min_db, vol_range.max_db)` to `hqp_status_to_zone`,
which runs on the handshake and on **every** status poll, plus two more clamps on the command path
in `dispatch_hqplayer_action`. A daemon in that shape therefore panicked the managed worker before
it ever published: no zone for a surface to show, and no error explaining why. Before this PR the
same line was `status.volume_db as f32`, which could not panic — so the panic is new here, which is
why the fix is here.

A missing `min` is not a hostile hypothetical for this protocol. The verified live 6.0.2 sample
already omits `step`, which is exactly why `step_db` is an `Option` in this file.

**Fixed as a capability question, not as a safer clamp.** A range that cannot bound a level is not a
volume control, exactly as `enabled="0"` is not — publishing a slider whose track runs backwards is
the same class of claim this PR exists to remove. A zero-width range is refused for the adjacent
reason: nothing can move within it, and `is_muted` (`value <= min + tolerance`) would be
unconditionally true, so the zone would report itself permanently muted.

`require_volume_control` re-checks the same precondition. Once the projection refuses, the handler
cannot see an unusable range through the adapter — but it owns the two `.clamp` calls, and a panic
in an HTTP or MCP handler is a worse failure than a 400, so it owns the precondition rather than
trusting its caller's caller.

**Commit:** `71c02a6` — `fix(328): refuse an unorderable volume range instead of panicking`.

### Accepted, not fixed

`quantise_db` can round a clamped level up to **0.005 dB past the published `max`**. With
`max = -20.004`, an absolute write clamps to the `f32` bound and then quantises to `-20.00`, which
is above it. Real, but 0.005 dB is far below audibility and below the finest step any observed
daemon reports, and every fix (quantise toward the interior, or clamp after quantising onto a
non-grid bound) trades a clean decimal on the wire — the thing
`a_computed_level_reaches_the_wire_without_float_noise` was written to protect — for nothing a user
can perceive. Flagging it so it is a decision rather than an oversight; say the word if you want it
closed anyway.

### Tests / red evidence

Two tests added to `tests/hqplayer_direct_zone.rs`, both written before the fix and observed to
fail:

- `an_unorderable_volume_range_is_refused_rather_than_panicking_the_poll` — end to end through the
  wire fake and `POST /control`.
- `a_nan_or_zero_width_volume_range_publishes_no_control` — the NaN, infinite, reversed and
  zero-width shapes the fake cannot express, through the existing `project_direct_zone` seam.

RED output, before the fix:

```
thread '...' panicked at library/core/src/num/f64.rs:1430:9:
min > max, or either was NaN. min = 0.0, max = -10.0
   ... repeated on every poll ...
thread '...' panicked at tests/hqplayer_direct_zone.rs:206:9:
zone never satisfied the condition inside the bounded budget; zones=[]
```

GREEN after the fix, at `71c02a6`:

- `cargo test --all-features` — **1252 passing, 0 failing** across 33 binaries
- `tests/hqplayer_direct_zone.rs` — 58 passing (56 existing + 2 new)
- `cargo fmt --check` clean
- `cargo clippy --lib --all-features -- -D warnings` clean

### API impact

None. No route added, removed or changed; `tests/fixtures/api_routes.txt` untouched;
`cargo test --test api_contract` green. The fix removes a `volume_control` block from a
`GET /knob/zones` / `GET /events` payload in one specific circumstance — a daemon whose range cannot
bound a level — which is a value change within the existing optional field, not a shape change. No
MCP tool name, parameter or schema touched. No adaptive control plane surface added or wired.

### Live status

No live HQPlayer host was contacted at any point in this review. All evidence is hermetic, from the
wire/model fake in `tests/mock_servers/hqplayer`, consistent with this PR's own hermetic-only claim.

### Residual risks

- The 0.005 dB quantisation overshoot above.
- The 2 s capability-staleness window between the published flags and the command that is judged by
  them. Already recorded in `.oh/hqplayer-direct-zone.md`; closing it needs a forbidden adapter read
  or daemon-side compare-and-set.
- Mute's protocol-inherent false positive (a level deliberately turned to the floor reads as muted)
  and the absence of any unmute. Already recorded.
- `TrackMetadata.sample_rate` falls back to the root `samplerate` on the evidence of a single corpus
  document where root and child agree. Already recorded and left to #332's live qualification; still
  the most falsifiable claim in this PR.
- `hqp_status_to_zone` falls back to `hqplayer:<host>` when an adapter has no instance name, and that
  zone id would not resolve back to an instance on the command path. Unreachable today (every
  adapter the manager hands out is named), but a latent unroutable-zone hazard if a bare
  `HqpAdapter` is ever wired to publish.

### Next action

Re-review `71c02a6` and merge after #382. Dependent branch `feat/issue-401-hqplayer-mcp-beta-validation`
(PR #406) has already been rebased onto this head and its full composed suite re-run green.

muness:
Superseded by #428, the single v3-based consolidated HQPlayer direct-control beta PR. Review history and evidence remain here; please use #428 for further review and the QNAP x64 test artifact.

Issue #400, opened 2026-07-31
MCP queue: inspect the queue, play from it, and mutate it where LMS allows


muness:
OH: 80222d6d

Parent epic: #392

## Problem Statement

**Current framing:** MCP has no queue coverage.

**Reframed as:** Enqueue works — `hifi_play action='queue'` maps to `PlayAction::Queue` for Roon and `LmsPlayAction::Queue` for LMS, and the server's `instructions` teach playlist building that way — but the queue is write-only. A client can add ten tracks and never see what it built, cannot tell whether an add landed, cannot skip to the fifth item, and cannot recover from a mistake except by replacing playback. The stated gap was wrong in direction: what is missing is not enqueue, it is **observability**.

**The shift:** From "queue is a play action" to "queue is state you can read and address", exposing only the mutations each provider genuinely supports.

The capability exists and is unused on both sides, and **the two providers are unequal in a direction that matters**:

- **Roon:** the pinned `roon-api` fork (`ohc/main`) exposes `subscribe_queue(zone_or_output_id, max_item_count)` and `play_from_here(zone_or_output_id, queue_item_id)`. Nothing in UHC calls either. Read + jump-to-position, no reorder, no remove.
- **LMS:** materially richer. `status` already returns `playlist_loop`, `playlist_tracks`, and `playlist_cur_index` (the adapter parses all three today for now-playing). Beyond reading, the LMS command surface supports genuine queue *mutation* — reorder, remove-at-index, clear, jump-to-index (the adapter already issues `playlist index +1/-1` for next/previous). And the LMS JSON-RPC transport already exists in-repo — `LmsRpc::execute` (`src/adapters/lms.rs:266`) speaks `slim.request` generically, so no new transport work is needed. **But it is private on a private struct**, so each capability needs a purpose-specific `pub` method on `LmsAdapter`; do not make `execute` public and hand MCP arbitrary CLI construction.

So "advertise only what the provider supports" cuts in LMS's favour here: LMS should end up with a *larger* queue surface than Roon, and a design that levels both down to Roon's read-only shape would be its own kind of dishonesty.

### Constraints

[...]

Issue #399, opened 2026-07-31
MCP browse: provider-neutral navigation contract with paging (Roon implementation)


muness:
OH: 80222d6d

Parent epic: #392

## Problem Statement

**Current framing:** Add a browse tool to MCP.

**Reframed as:** The bridge can navigate a Roon library — `RoonAdapter::browse()` and `load()` back the `/roon/browse` and `/roon/browse/status` HTTP routes, and the web UI uses them — but MCP gets only flat `hifi_search`, capped at 10 results with no paging. A client cannot answer "what genres are in my library", cannot walk Artist → Album → Track, and cannot reach anything search does not surface in its first ten hits.

**The shift:** From one flat search call to navigable hierarchy with paging, using the same adapter path the web UI already exercises.

**Scope split:** this issue defines the browse **tool contract** and implements it for **Roon**. LMS browse — which the LMS command surface fully supports — lands in the follow-up issue against this same contract. The contract must therefore be designed provider-neutral, not fitted to Roon's session-key model.

### Constraints

- **Hard:** The contract is provider-neutral — Roon's stateful `multi_session_key` + `pop_all`/level model is an implementation detail of one provider, and the tool's shape must not bake it in, because LMS browse is stateless index-and-offset and must fit the same contract without contortion. Statefulness never leaks to the client: it gets an opaque navigation handle or refs, never a session key or an `item_key`. Bounded output: paging is mandatory with an enforced cap, because unbounded `load()` against a large library blows past any sane response size. Browse degrades cleanly when Roon Browse is not connected (`is_browse_connected()` — the HTTP handler already guards this; do the same). No HTTP changes. Additive: `hifi_search` is untouched.
- **Hard — say which kind of "no" it is:** where browse is unavailable, the result must distinguish **"this provider cannot browse"** from **"UHC has not wired browse for this provider yet."** For LMS the honest answer until the follow-up lands is *not yet implemented*, **not** unsupported. Reporting LMS browse as a provider limitation would be a false capability claim — LMS supports browsing richly.
- **Soft:** One `hifi_browse` tool with a nav handle vs. separate list/enter tools; page size default and cap; how much list metadata (level, title, total count) is surfaced.

### What this framing enables

Library exploration instead of keyword luck; reaching content beyond the first ten search hits; a client that can present a genre or a discography; a contract LMS browse can land against without renegotiation; and #360's "page/search normalized content" satisfied for Roon now.

### What this framing excludes

[...]

Issue #398, opened 2026-07-31
MCP capabilities: let clients discover per-zone provider truth before acting


muness:
OH: 80222d6d

Parent epic: #392

## Problem Statement

**Current framing:** Document which backends support which MCP features.

**Reframed as:** Provider capability is real and uneven — AGENTS.md's own matrix shows volume, search, and play-by-query as unsupported on OpenHome and UPnP — but the surface does not say so, so a model learns limits only by failing. Worse, `hifi_control` ends its action match with `other => other` and routes any unrecognized zone prefix to Roon, so a typo'd zone id or an unknown action is forwarded to the wrong backend and fails there. #360 names this exact anti-pattern in its exclusions: "generic tools that secretly default unknown zones to Roon."

**The shift:** From capability as documentation to capability as data the client can query, and from silent defaulting to explicit routing.

### Constraints

- **Hard:** Reported capability must be derived from what the code actually does, not from a hand-maintained table that can drift — if a capability is advertised, a test proves the corresponding call path exists for that provider. Additive: existing tools keep their current success behavior for zone ids that work today. Unknown or unprefixed zone ids get an explicit error naming the accepted prefixes instead of a Roon fallback — **and because today's Roon-default is load-bearing for bare (unprefixed) ids in `set_volume` and `hifi_control`, changing it is a behavior change that must be called out, tested, and justified in the PR, not slipped in.** No HTTP changes.
- **Hard — three states, not two.** Every capability resolves to one of: **supported**, **not supported by this provider**, or **not yet implemented in UHC**. Collapsing the last two into "unsupported" produces false claims about providers — the concrete case being LMS, where browse and queue mutation are richly supported by the provider but not yet wired here. A client must be able to tell "this will never work" from "this doesn't work yet."
[...]

Issue #397, opened 2026-07-31
MCP resources: expose readable bridge state as MCP resources alongside tools


muness:
OH: 80222d6d

Parent epic: #392

## Problem Statement

**Current framing:** UHC's MCP has no resources.

**Reframed as:** `/mcp` declares only `ServerCapabilities { tools }`, so `resources/list` returns nothing. This is a protocol *shape* gap rather than a coverage gap — zones, now-playing, bridge status, HQPlayer pipeline and profiles are all readable today through `read_only_hint` tools. But clients that model ambient state as subscribable resources must instead poll tools, and a model spends a tool call to answer "what is playing" — a question that is pure state.

**The shift:** From "every read is a tool invocation" to "state is addressable, tools are for acting" — while keeping every existing read tool exactly as it is.

### Constraints

- **Hard:** Strictly additive — all 10 tools keep working unchanged, including the `read_only_hint` ones this duplicates. No second source of truth: a resource and its equivalent tool must never be able to disagree. Declaring a capability obliges honoring it. HQPlayer resources respect the same settings-based gating that already filters HQPlayer tools out of `tools/list` when the adapter is disabled.
- **Hard — where each resource reads from.** Zone and now-playing resources read from `ZoneAggregator`, which owns that state. Bridge status, HQPlayer pipeline, and HQPlayer profiles read from the **same adapter accessors the equivalent tools already call** (`src/mcp/mod.rs` — `self.state.roon.get_status()`, `self.state.hqplayer.get_status()` / `get_pipeline_status()`). This supersedes an earlier, unsatisfiable version of this constraint that required everything to route through the aggregator: `ZoneAggregator` holds only `zones` and `bus` (`src/aggregator.rs:15-18`) and models none of that state. The AGENTS.md rule being honored is "no second source of truth for zone state" — **do not** expand the aggregator to hold connection or pipeline state just to satisfy a literal reading.
- **Soft:** URI scheme and granularity, and whether resource templates are used.

### SDK findings — already established, do not re-derive

The solution-space gate on this issue resolved the feasibility question against the pinned **rust-mcp-sdk 0.8.3 / rust-mcp-schema 0.9.5** (schema `2025-11-25`). See the gate report in the comments for evidence with line references. Summary:

- **All five resource hooks exist** on the SDK's server handler (list, list-templates, read, subscribe, unsubscribe), as do `notify_resource_list_changed` and `notify_resource_updated`. The SDK is not the constraint.
[...]

PR #404, opened 2026-07-31
feat(mcp): modularize src/mcp and add a contract test harness (#394)

### LINKED ISSUE #394: MCP foundation: modularize the MCP server and add a contract test harness
OH: 80222d6d

Parent epic: #392

## Problem Statement

**Current framing:** Split up `src/mcp/mod.rs` and add some MCP tests.

**Reframed as:** Every MCP change — new tool, new field, new capability — lands in one 855-line module with zero test coverage, so concurrent work self-conflicts and no change can be shown to have preserved the existing surface.

**The shift:** From "refactor later, test later" to "relieve the two constraints that make the rest of the epic unsafe, before touching behavior."

This issue changes **no behavior**. Its output is a module boundary and a harness that pins the current surface exactly as it is today.

### Constraints

- **Hard:** Behavior-preserving — `tools/list` output (names, descriptions, input schemas, hints), `initialize` result (server info, capabilities, instructions, protocol version), and every tool's runtime behavior are byte-for-byte unchanged. Per AGENTS.md, "refactor" means wiring every call site, not adding a module and leaving the old code in place. No HTTP route changes. Base on `v3`.
- **Soft:** Exact module decomposition and whether the harness drives the handler in-process or over HTTP.

### What this framing enables

Five downstream issues editing different files instead of one; a snapshot that fails loudly if any later PR alters an existing tool contract (which is how "additive only" becomes enforceable rather than aspirational); and TDD on MCP work, which AGENTS.md mandates but which is impossible today.

Any new tool, field, capability, or description change. Any change to the HQPlayer-tool filtering behavior in `handle_list_tools_request`. Touching `mcp/index.js` (legacy Node server).

## Acceptance criteria

- [ ] `src/mcp/` is decomposed by concern (suggested: `mod.rs` wiring/transport, `tools/` tool definitions + dispatch, `types.rs` response types, `routing.rs` zone-prefix routing). Every call site moved, nothing left duplicated.
- [ ] Zone-prefix routing (`lms:` / `openhome:` / `upnp:` / Roon-default) exists in exactly one place, called by every tool that routes. **Preserve today's behavior including the Roon default** — correcting that is CAPABILITIES' job, not this issue's.
- [ ] New `tests/mcp_contract.rs`:
  - [ ] snapshot of `tools/list` (10 tools, names + descriptions + input schemas + hints) against a committed fixture, e.g. `tests/fixtures/mcp_tools.json`
  - [ ] snapshot of the `initialize` result including declared capabilities and `instructions`
  - [ ] every tool's input schema is valid JSON Schema and every documented parameter appears in it
  - [ ] a no-orphan-field assertion: every field name a tool *returns* is either consumed by some tool's input or explicitly listed as display-only, encoding the AGENTS.md guardrail as a test
  - [ ] at least one tool-call round-trip per adapter family against `tests/mock_servers/`
- [ ] `cargo test --test mcp_contract` passes; the fixture is regenerable by a documented command.
- [ ] Deliberately breaking one tool description makes the snapshot test fail (demonstrate in the PR, then revert).
- [ ] `cargo clippy -- -D warnings` clean — that is what CI runs. **Do not target `cargo clippy --all-targets`:** it cannot pass, because `src/lib.rs` denies `clippy::unwrap_used` / `expect_used` / `panic` crate-wide (`src/lib.rs:20-22`) and src's own `#[cfg(test)]` modules violate it 76 times, pre-existing and unchanged by this epic; existing suites still pass.

## Dependencies

None. **Blocks every other issue in #392.**

## Notes for the implementer

- Verify the build per AGENTS.md: `dx build --release --platform web --features web`. Run `make css` first in a fresh worktree or nothing compiles; `dx` needs sccache disabled and rustup's `rustc` ahead of Homebrew's on PATH.
- Gate reports (solution / execute / ship, each with review + dissent) go on the PR per #392.
- Do not merge. Do not add labels.

muness:
Fixes #394

**This PR changes no behavior.** It relieves the two constraints that make the rest of #392 unsafe: one 855-line module, and zero MCP test coverage.

---

## CI ~~is not a meaningful gate here~~ — corrected

**Struck through.** I was briefed that feature-branch PRs here run zero CI workflows. That is true only when the PR's *base* is another feature branch. This PR targets `v3`, which matches `pull_request: branches: [master, v3]`, so it gets the full Lint / Test / Build set and **CI is a real gate**.

It immediately earned that: **CI's Lint job caught a `cargo fmt --check` failure that my local verification could not**, because I had been told `cargo clippy -- -D warnings` was the gate and ran exactly that. Fixed in `e877066` as a formatting-only commit. All 26 checks now pass.

The complete gate, for the record — `make css` runs first in both jobs:

| Job | Command |
|---|---|
| Lint | `cargo fmt --check` **and** `cargo clippy -- -D warnings` |
| Test | `cargo test --workspace` |
| Build WASM / Build Linux x64 / Smoke Test Binary / Synology Package Contract | — |

Everything below was also run locally, with output pasted rather than summarized.

## The commit order is the argument

| Commit | Content | `src/mcp/` |
|---|---|---|
| `169551a` | `tests/mcp_contract.rs` + fixtures | **unsplit**, 855 lines |
| `92f79ab` | every dispatch arm, both routing rules | **unsplit**, 855 lines |
| `7bb0470` | **the split** | 11 modules |
| `53d5ef6` | close two execute-gate dissent holes | 11 modules |
| `06821cb` | transport coverage, keyed fixture | 11 modules |
| `e877066` | `cargo fmt` (formatting only) | 11 modules |

The contract tests were written and passing against the original module, then the module was replaced underneath them. They still pass, and **neither fixture needed regenerating**:

```
$ git diff 92f79ab..7bb0470 -- tests/fixtures/
(no output)
```

`tools/list` and `initialize` are byte-identical across the split. That is the behavior-preservation proof, and it only exists because the tests came first.

## Decomposition

Vertical slices by tool family, so a tool's advertised contract (description, input schema) sits next to the behavior that has to match it — #396 changes both for `hifi_play`.

```
src/mcp/mod.rs              transport, session recovery, SDK wiring    248
src/mcp/server.rs           server_details() -> InitializeResult        60  [#397]
src/mcp/handler.rs          ServerHandler: advertise + dispatch         81  [#397]
src/mcp/types.rs            response shapes + result constructors      114  [#395]
src/mcp/routing.rs          zone prefix -> adapter, one place          268  [#398]
src/mcp/tools/mod.rs        tool_box! registry + list_tools()          108  [#399/#400]
src/mcp/tools/zones.rs      hifi_zones, hifi_now_playing                59
src/mcp/tools/transport.rs  hifi_control + volume                     139
src/mcp/tools/library.rs    hifi_search, hifi_play                    193  [#396/#399]
src/mcp/tools/status.rs     hifi_status                                44
src/mcp/tools/hqplayer.rs   the four HQPlayer tools                   139
```

Three decompositions were compared and one more rejected outright; see the solution-space gate report below. **One justification stated at that gate was later withdrawn** in the ship report: the merge-conflict argument for this layout did not survive inspection of the epic's actual waves. The layout stands on review surface, the routing centralization, and the `server_details()` / `list_tools()` seams.

### Routing was open-coded in four places with three different rules

Now `ZoneTarget` in `routing.rs`, called by every site:

```
$ grep -rn 'starts_with("lms:")\|starts_with("openhome:")\|starts_with("upnp:")\|starts_with("roon:")' src/mcp/
src/mcp/routing.rs:64:        if zone_id.starts_with("lms:") {
src/mcp/routing.rs:66:        } else if zone_id.starts_with("openhome:") {
src/mcp/routing.rs:68:        } else if zone_id.starts_with("upnp:") {
src/mcp/routing.rs:70:        } else if zone_id.starts_with("roon:") || !zone_id.contains(':') {
```

The rules are **not** interchangeable, and the difference is observable:

| zone id | transport | volume |
|---|---|---|
| `lms:x` | LMS | LMS |
| `roon:x` / bare `x` | Roon | Roon |
| `openhome:x` / `upnp:x` | own adapter | **refused** |
| `sonos:x` | **Roon** | **refused** |

A single `is_lms()`-style predicate cannot express that. `ZoneTarget::Unknown` is kept distinct from `Roon` even though every call site currently folds it into Roon — that distinction is the seam #398 needs, and collapsing it would erase the only place the silent default is visible.

**Today's defects are frozen, not fixed.** The Roon default for unprefixed ids and the silent Roon fallback for unrecognised prefixes both belong to #398. `routing.rs` says so at the point of the defect, and the tests say a green run means "unchanged", never "correct".

## What I ran locally

```
$ cargo test --test mcp_contract
test result: ok. 48 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

$ cargo test --lib
test result: ok. 95 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

$ cargo fmt --check                      # the gate I had missed
(exit 0)

$ cargo clippy -- -D warnings
Finished `dev` profile [unoptimized + debuginfo] target(s) in 9.12s      (exit 0)

$ cargo clippy --test mcp_contract       # new test file
(no findings in tests/mcp_contract.rs)

$ make css && dx build --release --platform web --features web
#   rustup toolchain ahead of Homebrew's on PATH, RUSTC_WRAPPER unset
  1.91s  INFO Client build completed successfully! 🚀
137.73s  INFO Server build completed successfully! 🚀
```

Every other test target, run individually — all green:

```
adapter_integration    42      protocol_schema        41
zones_sha_integration  20      volume_safety          14
volume_step             9      architecture_lint       8
await_in_lock_lint      4      ignored_send_lint       4
arbitrary_find_lint     3      release_hardening       3
aggregator_lint         2      api_contract            2
arch_packaging          6      oneshot_leak_lint       1
spawn_cancellation_lint 1      unbounded_channel_lint  1
client_harness         41  (+1 pre-existing failure, see below)
protocol_integration   32  (+1 pre-existing failure, see below)
```

## The snapshot bites (AC requirement)

Changed `hifi_zones`'s description in `src/mcp/tools/zones.rs` to `"... UPnP, Sonos)"`:

```
$ cargo test --test mcp_contract tools_list_matches_fixture
test tools_list_matches_fixture ... FAILED

MCP contract drift in tests/fixtures/mcp_tools.json

line 175:
  fixture:     "description": "List all available playback zones (Roon, LMS, OpenHome, UPnP)",
  actual:      "description": "List all available playback zones (Roon, LMS, OpenHome, UPnP, Sonos)",

The MCP surface changed. If that was intentional AND approved, regenerate with:
  UPDATE_MCP_FIXTURES=1 cargo test --test mcp_contract
Otherwise this is the bug: epic #392 is additive-only.
```

Reverted; 48/48 green. Demonstrated both before and after the split.

Two further guardrails were also demonstrated biting, because the dissent showed both had holes:

```
$ # declared protocol version lowered to V2025_06_18
$ cargo test --test mcp_contract declared_protocol_version
assertion `left == right` failed: the declared protocol version changed. The wire
snapshot cannot catch a raised version, so this assertion is the one that matters
  left: "2025-06-18"   right: "2025-11-25"

$ # McpSearchResult::subtitle renamed to caption
$ cargo test --test mcp_contract no_tool_returns_an_unclassified_field
MCP tools return field(s) ["caption"] that are neither consumed by a tool input
nor classified as display-only.
```

Both reverted.

## What the harness pins

`tests/mcp_contract.rs`, 48 tests, two layers:

**Wire level**, through the real Axum `/mcp` route via `tower::oneshot` — so `create_mcp_extension`, the SSE-framed streamable-HTTP transport, session handling and the `ServerHandler` impl are all exercised. This needed **no new seams**, which is why it could be written against the unsplit code.

- `tools/list`: all 10 tools, descriptions, input schemas, annotation hints
- `initialize`: server info, capabilities, instructions, protocol version
- the settings-based HQPlayer filter: with the adapter off, exactly the six non-HQPlayer tools in order
- protocol negotiation for `2024-11-05` / `2025-03-26` / `2025-06-18`
- stale-session auto-recovery, plus GET and DELETE

**Behavior**, because a `tools/list` snapshot guards declarative macro output while the risk lives in the 330-line dispatch match:

- every one of the 10 arms reached, output pinned, error strings included — a model reads `"Radio mode not supported for LMS. Use 'play' or 'queue'."` and retries, so that text is contract
- the full HQPlayer alias table case by case: `mode`, `filter1x`, `filter_1x`, `filterNx`, `filter_nx`, `filternx`, `shaper`, `dither`, `rate`, `samplerate`
- the action map pinned **at the backend** via a new command log on `MockLmsServer`: `playpause` must arrive as LMS `pause`, `previous`/`prev` as `playlist index -1`, `volume_up`/`volume_down` as `+N`/`-N` with a defaulted delta of 5
- `hifi_status` asserted to keep BTreeMap key order, so nobody "tidies" the `json!` literal into a struct and silently reorders the text clients receive
- both routing rules and the fact that they differ
- no orphaned fields (AGENTS.md), inventory derived from the production types

## Deliberate calls I made

- **Harness drives both HTTP and in-process.** #394 left this open. HTTP for the wire snapshots because it covers every layer the split touches and needed no new seams; in-process for schema validity, the filter, orphan fields and routing.
- **Fixture keyed by tool name, not a positional array.** #392 is additive-only, and an array conflates "a tool was added" (permitted) with "an existing tool changed" (forbidden). Keyed, adding a tool adds one key and leaves every existing key byte-identical. #399 and #400 land in parallel and each appends a tool, so the array format would have forced the second to merge into a regeneration. Content equivalence with the array version was verified, not assumed.
- **Order pinned separately** from content, so a reordering and a description change fail different assertions.
- **No JSON Schema validation crate added.** Structural checks plus an explicit `EXPECTED_TOOL_PARAMS` table are dependency-free and make a parameter change a visible edit.
- **`serverInfo.version` redacted in the fixture** and asserted separately, because release builds inject it from the git tag — otherwise the test passes locally and fails every release build.
- **`hifi_status` kept as a `json!` literal.** `serde_json` without `preserve_order` emits BTreeMap order; a struct emits declaration order. Converting it would reorder keys clients see.
- **`SEARCH_LIMIT` split back into two constants.** These were two independent literals before this PR; collapsing them coupled two backends that page and rank differently. A refactor mandated to change no behavior shouldn't introduce couplings either.
- **Resource-method refusal pinned** (raised by the #397 design gate mid-task, verified empirically rather than accepted): `resources/list`, `resources/read` and `resources/templates/list` return `-32603 "does not support resources"` today, because the SDK gates dispatch on the declared capability. Pinned so #397 lands as a visible refusal-to-success diff. #394 does **not** declare the capability, and a test asserts that as a precondition.
- **Two test-fixture bugs fixed.** `MockLmsServer` gained a command log — without it, which backend command an MCP action produced was unobservable. `MockHqpServer` rejected any line starting with `<?xml`, but `HqpAdapter::build_command` sends the declaration and command on **one** line, so the HQPlayer adapter was unreachable from its own mock; it now strips the declaration instead of discarding the line, which is strictly more permissive.

## Deliberately left undone

- **`cargo clippy --all-targets` is not clean and cannot be.** `src/lib.rs` denies `clippy::unwrap_used`, `expect_used` and `panic` crate-wide; the `#[cfg(test)]` modules inside `src/` violate that 76 times — `src/config/mod.rs` (25), `src/adapters/lms_discovery.rs` (17), `src/adapters/lms.rs` (15), others. All pre-existing, count identical before and after this branch. `src/mcp/**` contributes **zero** findings. This AC needs rewording or a separate lint-debt issue.
- **Two `/hqp/discover` tests fail in my sandbox but pass in CI — a local-environment artifact, not repo breakage.** Both do live UDP SSDP broadcast, which my sandbox blocks. CI runs them green:
  ```
  $ gh run view 30674744344 --job 91299539795 --log | grep hqp_discover
  test shared_endpoints::get_hqp_discover ... ok
  test api_endpoints::hqp_discover_returns_json ... ok
  ```
  **Correcting my own earlier characterization:** I called these "pre-existing failures". My evidence (they fail at base commit `c47d36a`, and with this branch's changes stashed) proved only that *my branch does not cause them*, which was the question I needed answered — then I overstated it into "the repo has two failing tests", which CI disproves. Neither path touches `src/mcp`. Original local evidence, for completeness:
  ```
  $ # at c47d36a
  $ cargo test --test protocol_integration api_endpoints::hqp_discover_returns_json
  test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 32 filtered out
  ```
  Neither path touches `src/mcp`.
- **OpenHome and UPnP have no protocol round-trip.** Neither adapter has a `configure()`; they are SSDP-discovery only. Covered: mocks confirmed live, each prefix confirmed to reach its own adapter via that adapter's distinctive refusal. Adding a discovery seam is a behavior change.
- **Roon's success path is untested, and this is a gap in the central claim.** `MockRoonCore` is an in-memory state holder, not a server, so the Roon adapter cannot be driven by it — every Roon assertion here is an error-path assertion. Concretely: `hifi_search`'s Roon branch maps `item.title` to `title` and `item.subtitle` to `subtitle`, and if the split had swapped those two, **nothing in 48 tests would notice**. That mapping was preserved by inspection. A Roon mock server is its own project.
- **`GET /mcp` with an unknown session returns 500 where DELETE returns 404.** Found while writing the transport tests. Arguably wrong — a stale SSE reconnect is a client condition, not a server fault — but it is today's behavior, so it is recorded rather than corrected.
- **The `MockHqpServer` follow-up.** `adapter_integration.rs::hqp_mock_responds_to_getinfo` still hand-rolls the two-line workaround and still carries a comment blaming "a complex TCP protocol" for what was a mock bug. Test-only cleanup, filed separately.

## Rollback

Self-contained: nothing outside `src/mcp/`, `tests/mcp_contract.rs`, two mock servers and two new fixtures is touched.

- **Whole branch:** revert `e877066`, `06821cb`, `53d5ef6`, `7bb0470`, `92f79ab`, `169551a` — or simply don't merge. No migration, no config, no persisted state, no HTTP surface.
- **Split only, keeping the tests:** revert `7bb0470` alone. The tests were written against the unsplit module and pass against it, so that is a supported state rather than a broken one.
- **Runtime:** the MCP endpoint is one Axum route; disabling it is a `src/main.rs` route removal.
- **`tests/fixtures/api_routes.txt` untouched**, so no `api-change-approved` label is needed.

## Residual risk

1. **`UPDATE_MCP_FIXTURES=1` erases the guardrail in one command.** Weaker than I first thought, now that CI is known to run `cargo test --workspace` on PRs targeting `v3` — a red snapshot does block. But CI cannot tell a legitimate regeneration from a lazy one, so the failure message still has to argue its own case to whoever reads it.
2. **Green means "unchanged", never "correct".** The two things this pins hardest are defects: the Roon default, and search results with no addressable identity. Both are labelled at the point of the defect.
3. **Roon's untested success path** — above.
4. **Derive-macro output is not language-guaranteed stable** across compiler or SDK versions. A spurious fixture diff after a toolchain bump would look like a contract break.

## Gate reports

Solution / execute / ship, each with a review and a dissent, in the comments. The dissents blocked three times and each block produced a real fix:

- **solution-space dissent** — snapshots guard the low-risk half; the 330-line dispatch and its three lookup tables were nearly uncovered → `92f79ab`
- **execute dissent** — a *raised* protocol version was invisible to the snapshot; the orphan-field inventory was hand-listed exactly where #396 will edit → `53d5ef6`
- **ship dissent** — three of four public functions in `mod.rs` untested including `auto_recover_session`'s 55 lines; the array fixture conflates additive with modifying changes → `06821cb`

### DIFF
--- a/src/mcp/handler.rs
+++ b/src/mcp/handler.rs
@@ -1,0 +1,81 @@
+//! The MCP `ServerHandler`: advertise tools, and dispatch calls to them.
+//!
+//! Dispatch only. Every arm delegates to a handler in
+//! [`crate::mcp::tools`]; no tool behavior lives here. #397 extends this impl
+//! with the resource methods.
+
+use crate::api::{load_app_settings, AppState};
+use crate::mcp::tools::{self, HifiTools};
+use async_trait::async_trait;
+use rust_mcp_sdk::{
+    mcp_server::ServerHandler,
+    schema::{
+        schema_utils::CallToolError, CallToolRequestParams, CallToolResult, ListToolsResult,
+        PaginatedRequestParams, RpcError,
+    },
+    McpServer,
+};
+use std::sync::Arc;
+
+/// MCP server handler with access to app state.
+pub struct HifiMcpHandler {
+    state: AppState,
+}
+
+impl HifiMcpHandler {
+    pub fn new(state: AppState) -> Self {
+        Self { state }
+    }
+}
+
+#[async_trait]
+impl ServerHandler for HifiMcpHandler {
+    async fn handle_list_tools_request(
+        &self,
+        _params: Option<PaginatedRequestParams>,
+        _runtime: Arc<dyn McpServer>,
+    ) -> Result<ListToolsResult, RpcError> {
+        // Read settings per request: the operator can toggle the HQPlayer
+        // adapter while the server is running, and the advertised tool list has
+        // to follow.
+        let settings = load_app_settings();
+
+        Ok(ListToolsResult {
+            meta: None,
+            next_cursor: None,
+            tools: tools::list_tools(settings.adapters.hqplayer),
+        })
+    }
+
+    async fn handle_call_tool_request(
+        &self,
+        params: CallToolRequestParams,
+        _runtime: Arc<dyn McpServer>,
+    ) -> Result<CallToolResult, CallToolError> {
+        let tool: HifiTools = HifiTools::try_from(params).map_err(CallToolError::new)?;
+        let state = &self.state;
+
+        match tool {
+            HifiTools::HifiZonesTool(_) => tools::zones::handle_zones(state).await,
+            HifiTools::HifiNowPlayingTool(args) => {
+                tools::zones::handle_now_playing(state, args).await
+            }
+            HifiTools::HifiControlTool(args) => {
+                tools::transport::handle_control(state, args).await
+            }
+            HifiTools::HifiSearchTool(args) => tools::library::handle_search(state, args).await,
+            HifiTools::HifiPlayTool(args) => tools::library::handle_play(state, args).await,
+            HifiTools::HifiStatusTool(_) => tools::status::handle_status(state).await,
+            HifiTools::HifiHqplayerStatusTool(_) => tools::hqplayer::handle_status(state).await,
+            HifiTools::HifiHqplayerProfilesTool(_) => {
+                tools::hqplayer::handle_profiles(state).await
+            }
+            HifiTools::HifiHqplayerLoadProfileTool(args) => {
+                tools::hqplayer::handle_load_profile(state, args).await
+            }
+            HifiTools::HifiHqplayerSetPipelineTool(args) => {
+                tools::hqplayer::handle_set_pipeline(state, args).await
+            }
+        }
+    }
+}
--- a/src/mcp/handler.rs
+++ b/src/mcp/handler.rs
@@ -60,16 +60,12 @@ impl ServerHandler for HifiMcpHandler {
             HifiTools::HifiNowPlayingTool(args) => {
                 tools::zones::handle_now_playing(state, args).await
             }
-            HifiTools::HifiControlTool(args) => {
-                tools::transport::handle_control(state, args).await
-            }
+            HifiTools::HifiControlTool(args) => tools::transport::handle_control(state, args).await,
             HifiTools::HifiSearchTool(args) => tools::library::handle_search(state, args).await,
             HifiTools::HifiPlayTool(args) => tools::library::handle_play(state, args).await,
             HifiTools::HifiStatusTool(_) => tools::status::handle_status(state).await,
             HifiTools::HifiHqplayerStatusTool(_) => tools::hqplayer::handle_status(state).await,
-            HifiTools::HifiHqplayerProfilesTool(_) => {
-                tools::hqplayer::handle_profiles(state).await
-            }
+            HifiTools::HifiHqplayerProfilesTool(_) => tools::hqplayer::handle_profiles(state).await,
             HifiTools::HifiHqplayerLoadProfileTool(args) => {
                 tools::hqplayer::handle_load_profile(state, args).await
             }
--- a/src/mcp/mod.rs
+++ b/src/mcp/mod.rs
@@ -1,637 +1,57 @@
-//! MCP (Model Context Protocol) server for AI assistant integration
+//! MCP (Model Context Protocol) server for AI assistant integration.
 //!
-//! Provides HTTP endpoints for MCP clients with both Streamable HTTP and SSE transports.
-//! Routes are integrated into the main Axum app on port 8088 at /mcp endpoint.
+//! Provides HTTP endpoints for MCP clients over the Streamable HTTP transport.
+//! Routes are integrated into the main Axum app on port 8088 at `/mcp`.
+//!
+//! # Layout
+//!
+//! This module is the transport and wiring layer only. Everything else lives in
+//! a submodule, so concurrent work on the MCP surface (epic #392) edits
+//! different files:
+//!
+//! | Module      | Responsibility                                           |
+//! |-------------|----------------------------------------------------------|
+//! | `mod.rs`    | Axum route handlers, session recovery, SDK wiring        |
+//! | [`server`]  | the `initialize` result: identity, capabilities, protocol |
+//! | [`handler`] | the `ServerHandler` impl: advertise tools, dispatch calls |
+//! | [`tools`]   | tool definitions + handlers, by family                   |
+//! | [`types`]   | response shapes and result constructors                  |
+//! | [`routing`] | zone-prefix -> adapter, in exactly one place             |
+//!
+//! # The MCP surface is pinned by tests
+//!
+//! `tests/mcp_contract.rs` snapshots `tools/list` and `initialize` against
+//! committed fixtures and pins every tool's runtime behavior, including the
+//! error strings a model reads. Epic #392 is additive-only, so a fixture diff on
+//! a PR in that epic is a bug report, not a chore. Regenerate deliberately:
+//!
+//! ```sh
+//! UPDATE_MCP_FIXTURES=1 cargo test --test mcp_contract
+//! ```
 
-use crate::api::{load_app_settings, AppState};
-use async_trait::async_trait;
+pub mod handler;
+pub mod routing;
+pub mod server;
+pub mod tools;
+pub mod types;
+
+use crate::api::AppState;
 use axum::http::{HeaderMap, Method, Uri};
 use axum::{body::Body, extract::Extension, response::IntoResponse};
 use rust_mcp_sdk::{
     id_generator::{FastIdGenerator, UuidGenerator},
-    macros::{mcp_tool, JsonSchema},
-    mcp_server::{McpAppState, McpHttpHandler, ServerHandler, ToMcpServerHandler},
-    schema::{
-        schema_utils::CallToolError, CallToolRequestParams, CallToolResult, Implementation,
-        InitializeResult, ListToolsResult, PaginatedRequestParams, ProtocolVersion, RpcError,
-        ServerCapabilities, ServerCapabilitiesTools, TextContent,
-    },
+    mcp_server::{McpAppState, McpHttpHandler, ToMcpServerHandler},
     session_store::InMemorySessionStore,
-    tool_box, McpServer, TransportOptions,
+    TransportOptions,
 };
-use serde::{Deserialize, Serialize};
 use std::{sync::Arc, time::Duration};
+
+pub use handler::HifiMcpHandler;
+pub use server::server_details;
+pub use tools::HifiTools;
 
 /// MCP session header name
 const MCP_SESSION_ID_HEADER: &str = "mcp-session-id";
-
-// ============================================================================
-// Tool Definitions
-// ============================================================================
-
-/// List all available playback zones
-#[mcp_tool(
-    name = "hifi_zones",
-    description = "List all available playback zones (Roon, LMS, OpenHome, UPnP)",
-    read_only_hint = true
-)]
-#[derive(Debug, Deserialize, Serialize, JsonSchema)]
-pub struct HifiZonesTool {}
-
-/// Get current playback state for a zone
-#[mcp_tool(
-    name = "hifi_now_playing",
-    description = "Get current playback state for a zone (track, artist, album, play state, volume)",
-    read_only_hint = true
-)]
-#[derive(Debug, Deserialize, Serialize, JsonSchema)]
-pub struct HifiNowPlayingTool {
-    /// The zone ID to query (get from hifi_zones)
-    pub zone_id: String,
-}
-
-/// Control playback
-#[mcp_tool(
-    name = "hifi_control",
-    description = "Control playback: play, pause, playpause (toggle), next, previous, or adjust volume"
-)]
-#[derive(Debug, Deserialize, Serialize, JsonSchema)]
-pub struct HifiControlTool {
-    /// The zone ID to control
-    pub zone_id: String,
-    /// Action: play, pause, playpause, next, previous, volume_set, volume_up, volume_down
-    pub action: String,
-    /// For volume actions: the level (0-100 for volume_set) or amount to change
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub value: Option<f64>,
-}
-
-/// Search for music
-#[mcp_tool(
-    name = "hifi_search",
-    description = "Search for tracks, albums, or artists. Roon: searches Library, TIDAL, or Qobuz (use source param). LMS: searches all installed providers including streaming plugins (zone_id recommended as different players may have different sources configured).",
-    read_only_hint = true
-)]
-#[derive(Debug, Deserialize, Serialize, JsonSchema)]
-pub struct HifiSearchTool {
-    /// Search query (e.g., "Hotel California", "Eagles", "jazz piano")
-    pub query: String,
-    /// Zone ID for context-aware results. Recommended for LMS (different players may have different sources).
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub zone_id: Option<String>,
-    /// Where to search: "library" (default), "tidal", or "qobuz". Roon only; LMS searches all providers.
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub source: Option<String>,
-}
-
-/// Search and play music in one command
-#[mcp_tool(
-    name = "hifi_play",
-    description = "Search and play music. Searches and plays, queues, or starts radio from the first matching result. Use action='queue' to add to queue. action='radio' and source param are Roon-only; LMS searches all providers."
-)]
-#[derive(Debug, Deserialize, Serialize, JsonSchema)]
-pub struct HifiPlayTool {
-    /// What to play (e.g., "early Michael Jackson", "Dark Side of the Moon")
-    pub query: String,
-    /// Zone ID to play on (get from hifi_zones)
-    pub zone_id: String,
-    /// Where to search: "library" (default), "tidal", or "qobuz". Roon only; LMS searches all providers.
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub source: Option<String>,
-    /// What to do: "play" (default), "queue", or "radio". radio is Roon-only.
-    #[serde(skip_serializing_if = "Option::is_none")]
-    pub action: Option<String>,
-}
-
-/// Get overall bridge status
-#[mcp_tool(
-    name = "hifi_status",
-    description = "Get overall bridge status (Roon connection, HQPlayer config)",
-    read_only_hint = true
-)]
-#[derive(Debug, Deserialize, Serialize, JsonSchema)]
-pub struct HifiStatusTool {}
-
-/// Get HQPlayer status
-#[mcp_tool(
-    name = "hifi_hqplayer_status",
-    description = "Get HQPlayer Embedded status and current pipeline settings",
-    read_only_hint = true
-)]
-#[derive(Debug, Deserialize, Serialize, JsonSchema)]
-pub struct HifiHqplayerStatusTool {}
-
-/// List HQPlayer profiles
-#[mcp_tool(
-    name = "hifi_hqplayer_profiles",
-    description = "List available HQPlayer Embedded configurations",
-    read_only_hint = true
-)]
-#[derive(Debug, Deserialize, Serialize, JsonSchema)]
-pub struct HifiHqplayerProfilesTool {}
-
-/// Load an HQPlayer profile
-#[mcp_tool(
-    name = "hifi_hqplayer_load_profile",
-    description = "Load an HQPlayer Embedded configuration (will restart HQPlayer)",
-    destructive_hint = true
-)]
-#[derive(Debug, Deserialize, Serialize, JsonSchema)]
-pub struct HifiHqplayerLoadProfileTool {
-    /// Configuration name to load (get from hifi_hqplayer_profiles)
-    pub profile: String,
-}
-
-/// Change an HQPlayer pipeline setting
-#[mcp_tool(
-    name = "hifi_hqplayer_set_pipeline",
-    description = "Change an HQPlayer pipeline setting (mode, samplerate, filter1x, filterNx, shaper, dither)"
-)]
-#[derive(Debug, Deserialize, Serialize, JsonSchema)]
-pub struct HifiHqplayerSetPipelineTool {
-    /// Setting to change: mode, samplerate, filter1x, filterNx, shaper, dither
-    pub setting: String,
-    /// New value for the setting
-    pub value: String,
-}
-
-// Generate toolbox enum with all tools
-tool_box!(
-    HifiTools,
-    [
-        HifiZonesTool,
-        HifiNowPlayingTool,
-        HifiControlTool,
-        HifiSearchTool,
-        HifiPlayTool,
-        HifiStatusTool,
-        HifiHqplayerStatusTool,
-        HifiHqplayerProfilesTool,
-        HifiHqplayerLoadProfileTool,
-        HifiHqplayerSetPipelineTool
-    ]
-);
-
-// ============================================================================
-// Response Types (for JSON serialization)
-// ============================================================================
-
-#[derive(Debug, Serialize)]
-struct McpZone {
-    zone_id: String,
-    zone_name: String,
-    state: String,
-    volume: Option<f64>,
-    is_muted: Option<bool>,
-}
-
-#[derive(Debug, Serialize)]
-struct McpNowPlaying {
-    zone_id: String,
-    zone_name: String,
-    state: String,
-    title: Option<String>,
-    artist: Option<String>,
-    album: Option<String>,
-    volume: Option<f64>,
-    is_muted: Option<bool>,
-}
-
-#[derive(Debug, Serialize)]
-struct McpSearchResult {
-    title: String,
-    subtitle: Option<String>,
-}
-
-#[derive(Debug, Serialize)]
-struct McpHqpStatus {
-    connected: bool,
-    host: Option<String>,
-    pipeline: Option<McpPipelineStatus>,
-}
-
-#[derive(Debug, Serialize)]
-struct McpPipelineStatus {
-    state: String,
-    filter: String,
-    shaper: String,
-    rate: u32,
-}
-
-// ============================================================================
-// Server Handler
-// ============================================================================
-
-/// MCP server handler with access to app state
-pub struct HifiMcpHandler {
-    state: AppState,
-}
-
-impl HifiMcpHandler {
-    pub fn new(state: AppState) -> Self {
-        Self { state }
-    }
-
-    fn text_result(text: String) -> CallToolResult {
-        CallToolResult::text_content(vec![TextContent::from(text)])
-    }
-
-    fn error_result(msg: String) -> Result<CallToolResult, CallToolError> {
-        Ok(CallToolResult::text_content(vec![TextContent::from(
-            format!("Error: {}", msg),
-        )]))
-    }
-
-    fn json_result<T: Serialize>(data: &T) -> CallToolResult {
-        let json = serde_json::to_string_pretty(data).unwrap_or_else(|_| "{}".to_string());
-        Self::text_result(json)
-    }
-
-    // Helper method for volume control
-    async fn set_volume(
-        &self,
-        zone_id: &str,
-        value: f64,
-        relative: bool,
-    ) -> Result<CallToolResult, CallToolError> {
-        let result = if zone_id.starts_with("lms:") {
-            self.state
-                .lms
-                .change_volume(zone_id, value as f32, relative)
-                .await
-        } else if zone_id.starts_with("roon:") || !zone_id.contains(':') {
-            self.state
-                .roon
-                .change_volume(zone_id, value as f32, relative)
-                .await
-        } else {
-            return Self::error_result("Volume control not supported for this zone type".into());
-        };
-
-        match result {
-            Ok(()) => Ok(Self::text_result(format!(
-                "Volume {}",
-                if relative { "adjusted" } else { "set" }
-            ))),
-            Err(e) => Self::error_result(format!("Volume error: {}", e)),
-        }
-    }
-}
-
-#[async_trait]
-impl ServerHandler for HifiMcpHandler {
-    async fn handle_list_tools_request(
-        &self,
-        _params: Option<PaginatedRequestParams>,
-        _runtime: Arc<dyn McpServer>,
-    ) -> Result<ListToolsResult, RpcError> {
-        let mut tools = HifiTools::tools();
-
-        // Filter out HQPlayer tools if adapter is disabled in settings
-        let settings = load_app_settings();
-        if !settings.adapters.hqplayer {
-            tools.retain(|t| !t.name.starts_with("hifi_hqplayer"));
-        }
-
-        Ok(ListToolsResult {
-            meta: None,
-            next_cursor: None,
-            tools,
-        })
-    }
-
-    async fn handle_call_tool_request(
-        &self,
-        params: CallToolRequestParams,
-        _runtime: Arc<dyn McpServer>,
-    ) -> Result<CallToolResult, CallToolError> {
-        let tool: HifiTools = HifiTools::try_from(params).map_err(CallToolError::new)?;
-
-        match tool {
-            HifiTools::HifiZonesTool(_) => {
-                let zones = self.state.aggregator.get_zones().await;
-                let mcp_zones: Vec<McpZone> = zones
-                    .into_iter()
-                    .map(|z| McpZone {
-                        zone_id: z.zone_id,
-                        zone_name: z.zone_name,
-                        state: z.state.to_string(),
-                        volume: z.volume_control.as_ref().map(|v| v.value as f64),
-                        is_muted: z.volume_control.as_ref().map(|v| v.is_muted),
-                    })
-                    .collect();
-                Ok(Self::json_result(&mcp_zones))
-            }
-
-            HifiTools::HifiNowPlayingTool(args) => {
-                match self.state.aggregator.get_zone(&args.zone_id).await {
-                    Some(z) => {
-                        let np = McpNowPlaying {
-                            zone_id: z.zone_id,
-                            zone_name: z.zone_name,
-                            state: z.state.to_string(),
-                            title: z.now_playing.as_ref().map(|n| n.title.clone()),
-                            artist: z.now_playing.as_ref().map(|n| n.artist.clone()),
-                            album: z.now_playing.as_ref().map(|n| n.album.clone()),
-                            volume: z.volume_control.as_ref().map(|v| v.value as f64),
-                            is_muted: z.volume_control.as_ref().map(|v| v.is_muted),
-                        };
-                        Ok(Self::json_result(&np))
-                    }
-                    None => Self::error_result(format!("Zone not found: {}", args.zone_id)),
-                }
-            }
-
-            HifiTools::HifiControlTool(args) => {
-                // Map MCP actions to backend actions
-                let backend_action = match args.action.as_str() {
-                    "play" => "play",
-                    "pause" => "pause",
-                    "playpause" => "play_pause",
-                    "next" => "next",
-                    "previous" | "prev" => "previous",
-                    "volume_set" => {
-                        if let Some(v) = args.value {
-                            return self.set_volume(&args.zone_id, v, false).await;
-                        }
-                        return Self::error_result("volume_set requires a value (0-100)".into());
-                    }
-                    "volume_up" => {
-                        let delta = args.value.unwrap_or(5.0);
-                        return self.set_volume(&args.zone_id, delta, true).await;
-                    }
-                    "volume_down" => {
-                        let delta = args.value.unwrap_or(5.0);
-                        return self.set_volume(&args.zone_id, -delta, true).await;
-                    }
-                    other => other,
-                };
-
-                // Determine which adapter to use based on zone_id prefix
-                let result = if args.zone_id.starts_with("lms:") {
-                    self.state
-                        .lms
-                        .control(&args.zone_id, backend_action, None)
-                        .await
-                } else if args.zone_id.starts_with("openhome:") {
-                    self.state
-                        .openhome
-                        .control(&args.zone_id, backend_action, None)
-                        .await
-                } else if args.zone_id.starts_with("upnp:") {
-                    self.state
-                        .upnp
-                        .control(&args.zone_id, backend_action, None)
-                        .await
-                } else {
-                    // Default to Roon
-                    self.state.roon.control(&args.zone_id, backend_action).await
-                };
-
-                match result {
-                    Ok(()) => {
-                        // Return updated state
-                        if let Some(zone) = self.state.aggregator.get_zone(&args.zone_id).await {
-                            let np = McpNowPlaying {
-                                zone_id: zone.zone_id,
-                                zone_name: zone.zone_name,
-                                state: zone.state.to_string(),
-                                title: zone.now_playing.as_ref().map(|n| n.title.clone()),
-                                artist: zone.now_playing.as_ref().map(|n| n.artist.clone()),
-                                album: zone.now_playing.as_ref().map(|n| n.album.clone()),
-                                volume: zone.volume_control.as_ref().map(|v| v.value as f64),
-                                is_muted: zone.volume_control.as_ref().map(|v| v.is_muted),
-                            };
-                            let json = serde_json::to_string_pretty(&np)
-                                .unwrap_or_else(|_| "{}".to_string());
-                            Ok(Self::text_result(format!(
-                                "Action '{}' executed.\n\nCurrent state:\n{}",
-                                args.action, json
-                            )))
-                        } else {
-                            Ok(Self::text_result(format!(
-                                "Action '{}' executed.",
-                                args.action
-                            )))
-                        }
-                    }
-                    Err(e) => Self::error_result(format!("Control error: {}", e)),
-                }
-            }
-
-            HifiTools::HifiSearchTool(args) => {
-                // Route based on zone_id prefix
-                if args.zone_id.as_ref().is_some_and(|z| z.starts_with("lms:")) {
-                    // LMS search - uses globalsearch for all providers (library, TIDAL, Qobuz, etc.)
-                    match self
-                        .state
-                        .lms
-                        .search(&args.query, args.zone_id.as_deref(), Some(10))
-                        .await
-                    {
-                        Ok(results) => {
-                            let mcp_results: Vec<McpSearchResult> = results
-                                .into_iter()
-                                .map(|item| {
-                                    let subtitle = match item.result_type {
-                                        crate::adapters::lms::LmsSearchResultType::Album => {
-                                            item.artist.map(|a| format!("Album by {}", a))
-                                        }
-                                        crate::adapters::lms::LmsSearchResultType::Artist => {
-                                            Some("Artist".to_string())
-                                        }
-                                        crate::adapters::lms::LmsSearchResultType::Track => {
-                                            match (&item.artist, &item.album) {
-                                                (Some(a), Some(al)) => {
-                                                    Some(format!("{} - {}", a, al))
-                                                }
-                                                (Some(a), None) => Some(a.clone()),
-                                                _ => None,
-                                            }
-                                        }
-                                    };
-                                    McpSearchResult {
-                                        title: item.title,
-                                        subtitle,
-                                    }
-                                })
-                                .collect();
-                            Ok(Self::json_result(&mcp_results))
-                        }
-                        Err(e) => Self::error_result(format!("Search error: {}", e)),
-                    }
-                } else {
-                    // Roon search (default)
-                    use crate::adapters::roon::SearchSource;
-
-                    let source = match args.source.as_deref() {
-                        Some("tidal") => SearchSource::Tidal,
-                        Some("qobuz") => SearchSource::Qobuz,
-                        _ => SearchSource::Library,
-                    };
-                    let zone_id = args.zone_id.as_deref();
-
-                    match self
-                        .state
-                        .roon
-                        .search(&args.query, zone_id, Some(10), source)
-                        .await
-                    {
-                        Ok(results) => {
-                            let mcp_results: Vec<McpSearchResult> = results
-                                .into_iter()
-                                .map(|item| McpSearchResult {
-                                    title: item.title,
-                                    subtitle: item.subtitle,
-                                })
-                                .collect();
-                            Ok(Self::json_result(&mcp_results))
-                        }
-                        Err(e) => Self::error_result(format!("Search error: {}", e)),
-                    }
-                }
-            }
-
-            HifiTools::HifiPlayTool(args) => {
-                // Route based on zone_id prefix
-                if args.zone_id.starts_with("lms:") {
-                    use crate::adapters::lms::LmsPlayAction;
-
-                    // LMS: source param ignored (library only), radio not supported
-                    if args.action.as_deref() == Some("radio") {
-                        return Self::error_result(
-                            "Radio mode not supported for LMS. Use 'play' or 'queue'.".into(),
-                        );
-                    }
-
-                    let action = LmsPlayAction::parse(args.action.as_deref());
-
-                    match self
-                        .state
-                        .lms
-                        .search_and_play(&args.query, &args.zone_id, action)
-                        .await
-                    {
-                        Ok(message) => Ok(Self::text_result(message)),
-                        Err(e) => Self::error_result(format!("Play error: {}", e)),
-                    }
-                } else {
-                    // Roon play (default)
-                    use crate::adapters::roon::{PlayAction, SearchSource};
-
-                    let source = match args.source.as_deref() {
-                        Some("tidal") => SearchSource::Tidal,
-                        Some("qobuz") => SearchSource::Qobuz,
-                        _ => SearchSource::Library,
-                    };
-                    let action = PlayAction::parse(args.action.as_deref().unwrap_or("play"));
-
-                    match self
-                        .state
-                        .roon
-                        .search_and_play(&args.query, &args.zone_id, source, action)
-                        .await
-                    {
-                        Ok(message) => Ok(Self::text_result(message)),
-                        Err(e) => Self::error_result(format!("Play error: {}", e)),
-                    }
-                }
-            }
-
-            HifiTools::HifiStatusTool(_) => {
-                let roon_status = self.state.roon.get_status().await;
-                let hqp_status = self.state.hqplayer.get_status().await;
-
-                let status = serde_json::json!({
-                    "roon": {
-                        "connected": roon_status.connected,
-                        "core_name": roon_status.core_name,
-                    },
-                    "hqplayer": {
-                        "connected": hqp_status.connected,
-                        "host": hqp_status.host,
-                    }
-                });
-                Ok(Self::json_result(&status))
-            }
-
-            HifiTools::HifiHqplayerStatusTool(_) => {
-                let status = self.state.hqplayer.get_status().await;
-                let pipeline = self.state.hqplayer.get_pipeline_status().await.ok();
-
-                let mcp_status = McpHqpStatus {
-                    connected: status.connected,
-                    host: status.host,
-                    pipeline: pipeline.map(|p| McpPipelineStatus {
-                        state: p.status.state,
-                        filter: p.status.active_filter,
-                        shaper: p.status.active_shaper,
-                        rate: p.status.active_rate,
-                    }),
-                };
-                Ok(Self::json_result(&mcp_status))
-            }
-
-            HifiTools::HifiHqplayerProfilesTool(_) => {
-                let profiles = self.state.hqplayer.get_cached_profiles().await;
-                let profile_names: Vec<String> = profiles.into_iter().map(|p| p.title).collect();
-                Ok(Self::json_result(&profile_names))
-            }
-
-            HifiTools::HifiHqplayerLoadProfileTool(args) => {
-                match self.state.hqplayer.load_profile(&args.profile).await {
-                    Ok(()) => Ok(Self::text_result(format!(
-                        "Loaded profile: {}",
-                        args.profile
-                    ))),
-                    Err(e) => Self::error_result(format!("Failed to load profile: {}", e)),
-                }
-            }
-
-            HifiTools::HifiHqplayerSetPipelineTool(args) => {
-                // All settings now use name-based lookups - adapter handles conversion
-                // Only samplerate needs numeric parsing (Hz value)
-                let result = match args.setting.as_str() {
-                    "mode" => {
-                        // Accepts name like "PCM", "DSD", "[source]"
-                        self.state.hqplayer.set_mode(&args.value).await
-                    }
-                    "filter1x" | "filter_1x" => {
-                        self.state.hqplayer.set_filter_1x(&args.value).await
-                    }
-                    "filterNx" | "filter_nx" | "filternx" => {
-                        self.state.hqplayer.set_filter_nx(&args.value).await
-                    }
-                    "shaper" | "dither" => self.state.hqplayer.set_shaper(&args.value).await,
-                    "rate" | "samplerate" => {
-                        // Samplerate uses Hz value (e.g., "48000", "96000")
-                        if let Ok(v) = args.value.parse::<u32>() {
-                            self.state.hqplayer.set_rate(v).await
-                        } else {
-                            return Self::error_result(
-                                "Invalid rate value (expected Hz like 48000, 96000)".into(),
-                            );
-                        }
-                    }
-                    _ => {
-                        return Self::error_result(format!(
-                            "Unknown setting: {}. Valid: mode, samplerate, filter1x, filterNx, shaper, dither",
-                            args.setting
-                        ));
-                    }
-                };
-
-                match result {
-                    Ok(()) => Ok(Self::text_result(format!(
-                        "Set {} to {}",
-                        args.setting, args.value
-                    ))),
-                    Err(e) => Self::error_result(format!("Failed to set {}: {}", args.setting, e)),
-                }
-            }
-        }
-    }
-}
 
 // ============================================================================
 // MCP State Container (for Extension layer)
@@ -796,33 +216,6 @@ pub async fn handle_mcp_delete(
 ///
 /// Call this to get the extension layer, then add MCP routes and the layer to your router.
 pub fn create_mcp_extension(state: AppState) -> axum::Extension<McpExtState> {
-    let server_details = InitializeResult {
-        server_info: Implementation {
-            name: "unified-hifi-control".into(),
-            version: env!("CARGO_PKG_VERSION").into(),
-            title: Some("Unified Hi-Fi Control".into()),
-            description: Some("Control your music system via MCP".into()),
-            icons: vec![],
-            website_url: Some("https://github.com/open-horizon-labs/unified-hifi-control".into()),
-        },
-        capabilities: ServerCapabilities {
-            tools: Some(ServerCapabilitiesTools { list_changed: None }),
-            ..Default::default()
-        },
-        meta: None,
-        instructions: Some(
-            "Unified Hi-Fi Control MCP Server - Control Your Music System\n\n\
-            Use hifi_zones to list available zones, hifi_now_playing to see what's playing, \
-            hifi_control for playback control, hifi_search to find music, and hifi_play to play it.\n\n\
-            Note: hifi_search and hifi_play currently work with Roon and LMS zones only. \
-            Transport controls (play/pause/next/volume) work with all zones (Roon, LMS, OpenHome, UPnP).\n\n\
-            To build a playlist: call hifi_play multiple times with action='queue'. The first track \
-            can use action='play' to start playback, then subsequent tracks use action='queue' to add to the queue."
-                .into(),
-        ),
-        protocol_version: ProtocolVersion::V2025_11_25.into(),
-    };
-
     let handler = HifiMcpHandler::new(state);
 
     // Create MCP app state (mirrors what HyperServer does internally)
@@ -830,7 +223,7 @@ pub fn create_mcp_extension(state: AppState) -> axum::Extension<McpExtState> {
         session_store: Arc::new(InMemorySessionStore::new()),
         id_generator: Arc::new(UuidGenerator {}),
         stream_id_gen: Arc::new(FastIdGenerator::new(Some("s_"))),
-        server_details: Arc::new(server_details),
+        server_details: Arc::new(server::server_details()),
         handler: handler.to_mcp_server_handler(),
         ping_interval: Duration::from_secs(12),
         transport_options: Arc::new(TransportOptions::default()),
--- a/src/mcp/routing.rs
+++ b/src/mcp/routing.rs
@@ -1,0 +1,267 @@
+//! Zone-prefix routing for MCP tools.
+//!
+//! Every MCP tool that talks to a backend decides which adapter to use from the
+//! zone id's prefix. Before issue #394 that decision was open-coded in four
+//! places with three *different* rules; this module is the single place it lives.
+//!
+//! # This preserves today's behavior exactly, including its defects
+//!
+//! Two things here are known to be wrong, and #394 deliberately does not fix
+//! either — correcting them is #398's job:
+//!
+//! 1. An unprefixed zone id is treated as Roon.
+//! 2. An *unrecognised* prefix (`sonos:foo`) is also treated as Roon by
+//!    transport, search and play — silently, with no capability check.
+//!
+//! [`ZoneTarget::Unknown`] exists so #398 has something to change: today every
+//! call site folds `Unknown` into Roon, and the fix is to stop doing that. Do
+//! not collapse `Unknown` into [`ZoneTarget::Roon`] — that would erase the seam.
+//!
+//! # The three rules are not interchangeable
+//!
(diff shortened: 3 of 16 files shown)


muness:
## Gate report — SOLUTION SPACE / review

**Gate/role:** Solution-space gate, reviewer

### Aim and scope checked

Aim: make later #392 work parallelizable and provable, without changing behavior. Scope checked: choice of decomposition for `src/mcp/mod.rs` (855 lines) and choice of harness driver (in-process vs HTTP) — the two "soft" constraints #394 leaves to the implementer.

### Options compared

The terrain first. `src/mcp/mod.rs` has seven distinguishable concerns:

| Lines | Concern |
|---|---|
| 28–173 | 10 `#[mcp_tool]` structs + `tool_box!` registry |
| 175–219 | 5 `Mcp*` response structs |
| 221–279 | `HifiMcpHandler`, result constructors, `set_volume` (contains routing) |
| 281–301 | `handle_list_tools_request` (settings-based HQPlayer filter) |
| 303–633 | `handle_call_tool_request` — one 330-line match, 10 arms |
| 636–789 | Axum transport: 3 route handlers + `auto_recover_session` |
| 791–855 | `create_mcp_extension`: `InitializeResult` + SDK app-state wiring |

**Option A — the issue's literal suggestion (4 files):** `mod.rs` (transport) / `tools.rs` / `types.rs` / `routing.rs`.
Smallest diff, lowest risk. **Rejected:** `tools.rs` still carries the 330-line dispatch match, so #395, #396, #397, #399 and #400 all still edit one file. That is the exact constraint #394 exists to relieve — A satisfies the acceptance criteria as written while failing the stated purpose.

**Option B — split tool *definitions* from tool *handlers*** (`tools/defs.rs` holding all 10 `#[mcp_tool]` structs, one handler file per family).
**Rejected:** relocates the bottleneck rather than removing it. Every downstream issue that adds or changes a tool edits `defs.rs`. It also separates a tool's contract (description + input schema, which is exactly what the snapshot pins) from its behavior, so #396 — which changes both for `hifi_play` — would straddle two files for no gain.

**Option C — vertical slices by tool family (selected).**
```
src/mcp/mod.rs              transport, create_mcp_extension, re-exports
src/mcp/server.rs           server_details() -> InitializeResult        [#397]
src/mcp/handler.rs          HifiMcpHandler + ServerHandler dispatch     [#395/#397]
src/mcp/types.rs            Mcp* response structs + result constructors [#395]
src/mcp/routing.rs          ZoneTarget classification + unit tests      [#398]
src/mcp/tools/mod.rs        tool_box! registry + list_tools()           [#399/#400]
src/mcp/tools/{zones,transport,library,status,hqplayer}.rs
```
Each downstream issue gets a distinct primary file. Tool contract and tool behavior stay colocated. Remaining shared edit points are two one-line additive spots (`tool_box!` list, dispatch arm) — unavoidable without Option D, and trivially resolvable.

**Option D — replace `tool_box!` with a `Vec<Box<dyn ToolHandler>>` registry.**
Would reduce shared edit points to zero. **Rejected on the hard constraint:** `tool_box!` generates both `tools()` (which produces the exact `tools/list` bytes via the `JsonSchema` derive) and `TryFrom<CallToolRequestParams>`. Hand-rolling either risks changing the advertised schema. High risk against the one thing #394 must not do, for a marginal conflict-surface gain.

### Harness driver: HTTP, plus in-process

#394 leaves this open. Chosen: **both**, and the reason is TDD-shaped rather than aesthetic. Snapshotting `HifiTools::tools()` in-process would have worked pre-split, but would not have covered `create_mcp_extension`, the SSE-framed streamable-HTTP transport, session handling, or the settings filter — all of which the split touches. Driving the real `/mcp` route through `tower::oneshot` covers every layer a refactor could disturb, and needed **no new seams**, so it works against the unsplit code. In-process assertions cover what a snapshot cannot: schema validity, parameter inventory, orphan fields, routing.

### Evidence inspected

- `src/mcp/mod.rs` in full; four routing sites at lines 257–269, 371–389, 424, 497
- `rust-mcp-sdk-0.8.3/src/mcp_macros/tool_box.rs` — confirms `tools()` and `TryFrom` are macro-generated (basis for rejecting D)
- `rust-mcp-sdk-0.8.3/src/utils.rs:136` — `enforce_compatible_protocol_version` negotiates *down* to the client's version, which is why the harness requests the server's declared `2025-11-25`
- `tests/architecture_lint.rs:134` — `analyze_file` only scans `api/` and `knobs/`, so new `src/mcp/*` files need no allowlist entry
- `src/api/mod.rs:2059` `load_app_settings()` reads disk; `src/config/mod.rs:74` honours `UHC_CONFIG_DIR` — the seam that makes the filter test deterministic
- `grep` for `create_mcp_extension|handle_mcp_|mcp::` across `src/`: 4 call sites, all in `src/main.rs`, all preserved by re-export

### Findings

**P2 — routing has four shapes, not one, and they are not reducible to a single boolean.** `set_volume` (l.257) treats `roon:`-or-bare as Roon and *everything else* as unsupported; `hifi_control` (l.371) treats *everything else* as Roon. So `foo:bar` is "unsupported" for volume but "Roon" for control. A naive `is_lms()` helper would silently change behavior. Required: a 5-variant classification (`Lms`/`OpenHome`/`Upnp`/`Roon`/`Unknown`) where `Roon` and `Unknown` stay distinguishable, each call site mapping explicitly. This also hands #398 the exact seam it needs: today's silent default is `Unknown → Roon`.

**P2 — `serverInfo.version` is build-dependent.** `Cargo.toml` pins `version = "0.0.0"` with the real value injected from the git tag at release. An un-redacted snapshot would pass locally and fail every release build. Redacted in the fixture, asserted separately against `CARGO_PKG_VERSION`.

**P3 — `hifi_status` must keep its `serde_json::json!` literal.** `serde_json` without `preserve_order` serializes maps as `BTreeMap` (alphabetical); a struct serializes in declaration order. Converting that one response to a typed struct would reorder keys in the returned text — a real output change. It stays a `json!` literal, and its field inventory is listed literally in the orphan-field table.

**P3 — the `initialize` snapshot pins the *negotiated* version, not the declared one.** First fixture capture recorded `2025-06-18` because the test client asked for it. Fixed by having the client request the server's declared `2025-11-25`; a downgrade of the declared version now becomes an outright error for that request rather than a silent change. Confirmed by regenerating the fixture.

### Required actions

1. Implement routing as the 5-variant classification above; unit-test each of the four call-site shapes against today's behavior. Do not collapse `Roon` and `Unknown`.
2. Keep `hifi_status` as a `json!` literal.
3. Extract `server_details()` and `list_tools(hqplayer_enabled: bool)` as public seams — both are needed by later issues and both improve testability without changing output.
4. Byte-diff `git show` of the moved dispatch arms against the originals before declaring the split behavior-preserving. Reading them is not enough at this size.

### Disposition

Option C selected. Actions 1–4 carried into the execute gate.

### Residual risk

The `#[mcp_tool]` and `JsonSchema` derives generate the input schema from struct definition order and doc comments. Moving a struct between files should not change generated output, but *nothing in the language guarantees that*. The `tools/list` snapshot is the control, and it was committed before the move specifically so this risk is measured rather than assumed.

### Next gate authorization

Authorized to proceed to execute gate.

muness:
## Gate report — SOLUTION SPACE / dissent

**Gate/role:** Solution-space gate, dissent
**Verdict:** block

Blocking on P1. The plan as written would let a behavior-changing refactor ship with a green test suite.

### Contrary case

**P1 — The snapshot pins the *advertised* surface. The refactor's risk is in the *unadvertised* surface, and the plan leaves it almost entirely uncovered.**

`tools/list` and `initialize` are declarative data produced by derive macros. Moving a struct between files is very unlikely to perturb them — which means the two snapshots the plan leads with are guarding the *low-risk* half of the module.

The high-risk half is the 330-line `handle_call_tool_request` match, which is being redistributed across five files by hand. It contains at least three pure lookup tables that `tools/list` cannot see and that a careless move breaks silently:

- action mapping (l.347–367): `playpause` → `play_pause`, `previous|prev` → `previous`, and `volume_down` → `set_volume(-delta)`. **Invert that sign and every snapshot in this PR still passes.**
- default deltas: `args.value.unwrap_or(5.0)` on `volume_up`/`volume_down`
- HQPlayer setting aliases (l.594–621): `filter1x|filter_1x`, `filterNx|filter_nx|filternx`, `shaper|dither`, `rate|samplerate`. Drop one alias while moving the arm and a documented input starts returning "Unknown setting" — the precise AGENTS.md "no broken tools" failure, with a green suite.

The acceptance criteria ask for "at least one tool-call round-trip per adapter family." Read literally that is roughly four calls covering perhaps three of ten tools. The review accepted that framing without noticing it inverts the risk profile.

**Required (P1):** before the split is called behavior-preserving, every arm of the dispatch must be reached by a test that pins its observable output, and the three lookup tables above must be pinned case by case. Error-path output counts: `Error: Zone not found: {}`, `Error: Volume control not supported for this zone type`, `Error: Radio mode not supported for LMS. Use 'play' or 'queue'.` are strings an AI client reads and acts on, so they are contract too. Disconnected adapters are sufficient to reach most arms — coverage here does not require more mocks, only more calls.

**P2 — The parallelism argument for Option C is weaker than the review claims, and the review did not check it.**

I traced the epic's own waves against Option A rather than against "one file":

| Wave | Issues | Files under Option A | Conflict? |
|---|---|---|---|
| 2 | #395 alone | `tools.rs` | none — nothing parallel |
| 3 | #396, #397, #398 | `tools.rs`, `mod.rs`, `routing.rs` | **none** |
| 4 | #399, #400 | `tools.rs`, `tools.rs` | yes |

Option A produces exactly one conflicting pair, in wave 4, and Option C does not remove it — #399 and #400 still both edit `tools/mod.rs` and the dispatch. The review's "five downstream issues editing different files instead of one" is inherited from the issue text, not established by inspection.

What survives is a weaker and different claim: #395 rewrites result construction across all ten arms, and reviewing that as ten hunks in five small family files beats reviewing it as one 330-line hunk. That is a review-surface argument, not a merge-conflict argument. Option C should be selected on that basis and the overstatement withdrawn — otherwise the next person reads this PR as evidence that 11 files were needed to avoid conflicts that A would not have produced.

Not blocking: C is still defensible, and the routing centralization and the `server_details()` / `list_tools()` seams are worth having on their own.

**P2 — The orphan-field assertion, as specified, will bless the defect the epic exists to fix.**

`McpSearchResult` returns `{title, subtitle}` and no key. That *is* #392's keystone finding: the only route from "found it" to "playing it" is handing a title back to `hifi_play`, which re-searches. Encoding today's truth means `title` and `subtitle` land on the display-only list — and a bare display-only list reads as "reviewed, fine."

The instruction to encode today's truth is right; the representation matters. Every display-only entry must carry a reason string, and the two search fields must carry an explicit "not a stable identifier — #396" note, so the next implementer finds a labelled defect rather than an approval.

**P3 — `std::env::set_var` in a parallel test binary is a latent flake.** All six env-touching tests carry `#[serial_test::serial(uhc_config_dir)]` today. Nothing enforces that the seventh will. The requirement must be stated at the definition of the fixture, not left to the reader.

**P3 — The regeneration escape hatch can erase this PR's entire value, and nothing here prevents it.** `UPDATE_MCP_FIXTURES=1` rewrites the guardrail in place. A downstream implementer facing a red snapshot has a one-command way to make it green, and CI is not a gate on this repo's feature-branch PRs, so nobody is forced to look. This is not fixable inside #394 — but it must be named as residual risk rather than left implicit, because "the snapshot makes additive-only enforceable" is the PR's central claim and it holds only as long as humans read the failure.

**P4 — The fixture pins tool *order*.** Appending a tool to `tool_box!` leaves the fixture's first N entries intact; inserting one mid-list shifts everything. Both are additive changes, but only one produces a small diff. Worth one sentence in the test's docs so a downstream implementer appends rather than regenerating in confusion.

**P4 — `server.rs` holding one function is thin** and only justified by #397's future need. Acceptable; noted so it isn't mistaken for a pattern.

### Where the review is right

Rejecting Option D is correct and the reasoning is sound: `tool_box!` generates the code that produces the pinned bytes, and hand-rolling it to win a one-line conflict would trade the hard constraint for a soft one.

The 5-variant routing finding is the best thing in the review. `foo:bar` being "unsupported" for volume and "Roon" for control is a genuine trap that a plausible `is_lms()` helper walks straight into, and it was found by reading the code rather than by trusting the issue's four-way description.

Driving the harness over HTTP is the right call for a reason the review undersells: it required no new seams, so the pre-split snapshot is a measurement of the real code rather than a measurement of scaffolding built to be measured.

### Required actions

1. **(P1, blocking)** Reach every dispatch arm with an output-pinning test. Pin the action map, the volume defaults and sign, and every HQPlayer setting alias, case by case. Pin the error strings.
2. **(P2)** Withdraw the merge-conflict justification for Option C; restate it as review surface. Keep Option C.
3. **(P2)** Give every display-only field a reason; flag `title`/`subtitle` against #396.
4. **(P3)** State the `serial_test` requirement at `SettingsFixture`.
5. **(P3)** Name the `UPDATE_MCP_FIXTURES` escape hatch as residual risk in the PR body.
6. **(P4)** Document the ordering sensitivity in the test.

### Disposition

Block until action 1 is planned into the execute gate. Actions 2–6 are pass-with-actions.

### Residual risk after actions

Even with every arm reached, these tests pin *today's* behavior including its bugs — the Roon default for unprefixed ids most of all. #394 is right to freeze it rather than fix it, but the freeze is load-bearing in a way that is easy to misread later: a green suite will mean "unchanged," never "correct." That distinction has to survive in the test's own comments, because it will not survive in anyone's memory.

### Next gate authorization

Not authorized until action 1 is incorporated. Re-review at the execute gate.

muness:
## Gate report — EXECUTE / review

**Gate/role:** Execute gate, reviewer

### Aim and scope checked

Aim: decompose `src/mcp/` and add a contract harness, changing no behavior. Scope checked: the three commits, the resulting module boundaries, and whether the evidence actually supports the behavior-preservation claim.

### The TDD shape, and why it is the whole argument

| Commit | Content | State of `src/mcp/` |
|---|---|---|
| `169551a` | `tests/mcp_contract.rs` + 2 fixtures | **unsplit**, 855 lines |
| `92f79ab` | dispatch-arm and routing coverage | **unsplit**, 855 lines |
| `7bb0470` | the split | 11 modules, 1448 lines |

44 tests were written and passed against the original module, then the module was replaced underneath them. They still pass, and **neither fixture needed regenerating** — `tools/list` and `initialize` are byte-identical. That is the proof, and it only works because the tests came first.

```
$ cargo test --test mcp_contract      # at 92f79ab, before the split
test result: ok. 44 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

$ cargo test --test mcp_contract      # at 7bb0470, after the split
test result: ok. 44 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

$ git diff 92f79ab..7bb0470 -- tests/fixtures/
(no output)
```

### Evidence inspected

**Dissent P1 (blocking) — discharged.** The dissent was right that snapshots guard the low-risk half. Coverage added in `92f79ab`:

- all ten dispatch arms reached, output pinned, error strings included
- the HQPlayer alias table, each of `mode` / `filter1x` / `filter_1x` / `filterNx` / `filter_nx` / `filternx` / `shaper` / `dither` / `rate` / `samplerate` checked individually
- the action map pinned *at the backend*, via a new command log on `MockLmsServer`: `playpause` must arrive as LMS `pause`, `previous` and `prev` both as `playlist index -1`
- the volume sign and default delta pinned the same way: `mixer volume +7` / `-7`, and `+5` / `-5` when no value is given

That last one is the dissent's own example. Inverting `volume_down`'s sign now fails `lms_round_trip_pins_the_action_map_and_the_volume_sign`, where before it would have passed everything.

**Routing.** Four open-coded sites, three different rules, now one `ZoneTarget`. Verified centralized:

```
$ grep -rn 'starts_with("lms:")\|starts_with("openhome:")\|starts_with("upnp:")\|starts_with("roon:")' src/mcp/
src/mcp/routing.rs:64:        if zone_id.starts_with("lms:") {
src/mcp/routing.rs:66:        } else if zone_id.starts_with("openhome:") {
src/mcp/routing.rs:68:        } else if zone_id.starts_with("upnp:") {
src/mcp/routing.rs:70:        } else if zone_id.starts_with("roon:") || !zone_id.contains(':') {
```

The solution gate's P2 held up under implementation. `ZoneTarget::Unknown` is kept distinct from `Roon` even though every call site folds it into Roon, because that distinction is the only place the silent default is visible — it is #398's seam. `routing.rs` says so at the point of the defect.

**The routing tests found a real error — in themselves.** `volume_refuses_where_transport_defaults_to_roon` initially asserted `for_transport() != Roon` for `sonos:x`, which is exactly backwards: `sonos:x` routing to Roon for transport *is* the frozen behavior. The test failed, the assertion was wrong, and the production code was right. A test that can only pass is not evidence; this one demonstrated it can fail for the right reason before it was corrected.

**Settings filter.** `list_tools(hqplayer_enabled: bool)` is now a pure function with unit tests asserting the filter removes exactly four tools and preserves the relative order of the other six. Settings are still read per request in `handle_list_tools_request`, so runtime toggling still works.

**Byte-identical output.** Two things were nearly broken and deliberately preserved:
- `hifi_status` stays a `serde_json::json!` literal. `serde_json` without `preserve_order` emits `BTreeMap` order (`hqplayer` before `roon`); a struct would emit declaration order. `hifi_status_shape_is_pinned` asserts the ordering, not just the content.
- `McpZone` / `McpNowPlaying` / `McpSearchResult` field order preserved verbatim, since these serialize in declaration order.

**Scope discipline.** No HTTP route touched, `tests/fixtures/api_routes.txt` untouched, `mcp/index.js` untouched, no tool description or input schema edited. `src/main.rs`'s four call sites work unchanged through re-exports.

**Coordinator's mid-task addition (resource refusal).** Verified empirically rather than accepted: `resources/list`, `resources/read` and `resources/templates/list` all return `-32603` with "does not support resources". Pinned, with the test asserting as a *precondition* that `capabilities.resources` is absent — so #394 cannot accidentally declare the capability it is only supposed to observe.

**Test suite.** Every target run individually. All green except two pre-existing failures — see findings.

### Findings

**P2 — Two test failures exist, and neither is caused by this branch.** `client_harness::shared_endpoints::get_hqp_discover` and `protocol_integration::api_endpoints::hqp_discover_returns_json` both fail with 500 instead of 200. Both hit `/hqp/discover`, which does live UDP SSDP broadcast (`hqp_discover_handler` -> `discover_hqplayers`) and fails in this sandbox. Proven pre-existing two ways: stashing this branch's changes still failed, and checking out the base commit `c47d36a` still failed:

```
$ git checkout c47d36a && cargo test --test protocol_integration api_endpoints::hqp_discover_returns_json
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 32 filtered out
```

Nothing in either path touches `src/mcp`. Not fixed here — that is someone else's issue.

**P2 — `cargo clippy --all-targets` cannot be clean, and the AC asking for it is unmeetable as written.** `src/lib.rs` sets `#![deny(clippy::unwrap_used)]`, `expect_used` and `panic` crate-wide, which the `#[cfg(test)]` modules inside `src/` violate 76 times — all pre-existing, in `src/config/mod.rs` (25), `src/adapters/lms_discovery.rs` (17), `src/adapters/lms.rs` (15) and others. What *is* verifiable, and was:
- `cargo clippy -- -D warnings` (what CI runs) exits 0
- the error count is unchanged from the pre-split baseline, 76 before and after
- `src/mcp/**` contributes **zero** clippy findings of any severity

**P3 — Two mock-server fixture fixes were needed, and one is a latent bug affecting more than this issue.** `MockHqpServer` rejected any line starting with `<?xml`, but `HqpAdapter::build_command` emits the declaration and the command on one line, so the mock answered nothing and the adapter timed out — the HQPlayer adapter was **unreachable from its own mock**. `adapter_integration.rs` had worked around this by hand-writing the two parts on separate lines and commenting "because HQP adapter uses a complex TCP protocol"; the actual cause was mock fidelity. Now fixed by stripping the declaration rather than discarding the line, which is strictly more permissive, so the existing test is unaffected. Anyone who tries to integration-test HQPlayer benefits.

**P3 — OpenHome and UPnP have no real round-trip, and the AC's "per adapter family" is only partly met.** Neither adapter has a `configure()`; they are SSDP-discovery only, so a mock cannot be injected without a discovery seam. What is covered: the mocks are confirmed live, and each prefix is confirmed to reach *its own* adapter via that adapter's distinctive refusal (`Device not found:` for OpenHome, `Renderer not found:` for UPnP). That is the routing edge a refactor could break. The transport gap is documented in the test body rather than glossed. Adding a discovery seam is a behavior change and out of scope for a no-behavior-change issue.

**P4 — 11 files, 1448 lines, up from 1 file and 855.** Roughly 400 of the increase is doc comments recording *why* things are the way they are — the routing asymmetry, the `BTreeMap` ordering trap, which downstream issue owns which file. Given that this issue's entire output is a boundary other people will work inside, that seems like the right place to spend lines. Worth watching that it does not become a habit.

### Required actions

1. **(P2)** State both pre-existing failures explicitly in the PR body with the base-commit evidence. Do not let "44 passed" imply the whole suite is green.
2. **(P2)** State plainly that `cargo clippy --all-targets` is not clean and cannot be, with the numbers, rather than quietly substituting a narrower command.
3. **(P3)** Flag the `MockHqpServer` fidelity bug for follow-up so `adapter_integration.rs`'s workaround can be removed and its stale comment corrected.
4. **(P3)** Demonstrate the snapshot biting in the PR body, per the AC.
5. Complete the `dx build --release --platform web --features web` verification before the ship gate.

### Disposition

Pass with actions 1–5 carried to the ship gate. The dissent's blocking P1 is discharged; its P2 (overstated parallelism claim) is addressed in the ship report.

### Residual risk

The harness pins today's behavior *including its defects*, and the defects it freezes are load-bearing: the Roon default, and search results with no stable identifier. A green suite means "unchanged", never "correct". That distinction lives in comments in `routing.rs`, `types.rs` and `mcp_contract.rs` because it will not live in anyone's memory.

### Next gate authorization

Authorized to proceed to ship gate once action 5 completes.

muness:
## Gate report — EXECUTE / dissent

**Gate/role:** Execute gate, dissent
**Verdict:** block

The split is sound and the dissent's earlier P1 is genuinely discharged. Blocking on two holes in the *harness*, both concrete, both fixable, and both in guarantees the review states as though they hold.

### Contrary case

**P1 — The `initialize` snapshot does not detect a raised protocol version. The seam to fix it was built in this very PR and then not used.**

The SDK negotiates *down* to the client's request (`utils.rs:136`). The test client requests `2025-11-25`. So:

| declared version | client asks | result | snapshot |
|---|---|---|---|
| `2025-11-25` (today) | `2025-11-25` | `2025-11-25` | passes |
| `2025-06-18` (lowered) | `2025-11-25` | error | **fails** ✓ |
| `2026-06-18` (raised) | `2025-11-25` | `2025-11-25` | **passes** ✗ |

Raising the declared protocol version is invisible. The review's claim that the snapshot "pins the *declared* version" is half true — it pins a lower bound, not the value.

Worse, `SERVER_PROTOCOL_VERSION` in the test is a hand-copied duplicate of `ProtocolVersion::V2025_11_25` in `server.rs`, with nothing connecting them. Two constants that must agree and no assertion that they do is exactly the pattern this PR exists to eliminate.

This is embarrassing rather than hard, because **`server_details()` was extracted in this PR specifically to be inspectable without a transport**, and the test does not call it. The whole justification for that seam was testability, and the one test that needs it goes through HTTP instead.

**Required:** assert the declared version directly off `server_details()`, so the test reads production's value instead of restating it. Keep the wire-level snapshot as well — it covers negotiation, which the direct assertion does not.

**P1 — The no-orphan-field guardrail has a hole exactly where the epic's keystone defect lives.**

`no_tool_returns_an_unclassified_field` reads returned fields from real tool responses — for four tools. For the fields it cannot reach without a live music library it does this:

```rust
// `hifi_search`'s result fields are unreachable without a live music library,
// so they are listed explicitly. Keep in sync with `McpSearchResult`.
for field in ["title", "subtitle"] {
    returned.insert(field.to_string());
}
```

"Keep in sync with" is folklore, and the test's own `stale` check exists to stamp out folklore. Consider renaming `McpSearchResult::subtitle` to `caption`:

- `subtitle` is still hand-inserted into `returned`, and is still in `FIELD_ROLES`, so the `stale` check passes
- `caption` is never inserted, so the unclassified check never sees it
- **the test passes, and a brand-new unclassified field ships**

The hole is in `McpSearchResult` and `McpPipelineStatus` — and `McpSearchResult` is the struct at the centre of #392's keystone finding, the one #396 exists to change. The guardrail is weakest at precisely the field set most likely to be edited next.

Again the fix is available and unused: the split made these types `pub`, so the test can construct a sample instance, serialize it, and collect the real keys. That is a strictly stronger inventory than the hand-written list, and it is what the four reachable tools already do.

**Required:** derive the inventory from the types, not from a comment.

**P3 — `LmsHarness::stop()` aborts the aggregator but never stops the LMS adapter.** `adapter_integration.rs` calls `adapter.stop()`; this harness does not. Two consequences: the polling task outlives the test for the rest of the process, and it may write `lms-config.json` into a `TempDir` that `SettingsFixture` has already dropped. `serial_test` prevents cross-test corruption today, so this is untidiness rather than a bug — but it is untidiness in the fixture other issues will build on, which is the worst place for it.

**P3 — Roon's success path is genuinely untested and the review lists this only as residual risk.** `hifi_search`'s Roon branch maps `item.title` -> `title` and `item.subtitle` -> `subtitle`. If the split had swapped those two, nothing in 44 tests would notice: `MockRoonCore` is an in-memory state holder, not a server, so the Roon adapter cannot be driven by it. Every Roon assertion in this PR is an error-path assertion. That is not fixable within #394 — building a Roon mock server is a project — but it should be stated as a *coverage gap in this PR's central claim*, not filed under general residual risk. "Byte-identical output" is proven for `tools/list`, `initialize`, and the paths the tests reach. Roon's success path is asserted by inspection only.

**P4 — `SEARCH_LIMIT` couples two previously independent values.** The original had two separate `Some(10)` literals, one for LMS and one for Roon. They are now one constant, so a future change to LMS's limit silently changes Roon's. No behavior change today, and arguably the constant is better; but it is a coupling that did not exist, introduced by a refactor whose mandate was to introduce none.

**P4 — The gate-1 dissent's P2 has not been actioned.** It asked that the merge-conflict justification for Option C be withdrawn and restated as a review-surface argument, because tracing the epic's waves against Option A produces exactly one conflicting pair. The execute review acknowledges the solution gate's P2 (routing) but says nothing about this one. It is a claim in the PR record that inspection does not support, and the ship report is the last chance to correct it.

### Where the review is right

The three-commit ordering is the real thing here, not a ceremony. `git diff 92f79ab..7bb0470 -- tests/fixtures/` returning empty is a stronger statement than any prose in this PR, and it is only available because the fixtures were captured against the 855-line module.

Pinning the action map at the *backend* rather than at the MCP boundary is the right instinct and required real work — the `MockLmsServer` command log did not exist, and without it `playpause` -> `play_pause` -> LMS `pause` was simply unobservable. That is the difference between testing that a call returns Ok and testing that it did the right thing.

Proving the two pre-existing failures twice — once by stashing, once by checking out `c47d36a` — is the correct standard. "It was already broken" is the most abused claim in software and it was actually demonstrated here.

And the routing test that failed because the *test* was wrong about `sonos:x` is worth more than the tests that passed first try. It showed the assertion could fail.

### Required actions

1. **(P1, blocking)** Assert the declared protocol version off `server_details()` directly. Keep the wire snapshot for negotiation.
2. **(P1, blocking)** Derive the orphan-field inventory from `McpSearchResult` / `McpPipelineStatus` instead of a hand-written list.
3. **(P3)** Stop the LMS adapter in `LmsHarness::stop()`.
4. **(P3)** State Roon's untested success path as a coverage gap in the PR body, not as generic residual risk.
5. **(P4)** Either justify `SEARCH_LIMIT`'s coupling or restore two independent constants.
6. **(P4)** Withdraw the merge-conflict justification for Option C in the ship report.

### Disposition

Block until 1 and 2 land. Both are narrow, both use seams this PR already built, and leaving them means shipping a guardrail with a hole under the header "additive-only is now enforceable".

### Residual risk after actions

Even fully fixed, this harness pins what the code does, and the two things it most firmly pins are defects: unprefixed zone ids routing to Roon, and search results with no addressable identity. The PR is right to freeze both. The danger is a future reader taking a green `mcp_contract` run as a statement about correctness. Comments in `routing.rs` and `types.rs` are the only defence, and comments lose.

### Next gate authorization

Not authorized. Re-review at the ship gate after actions 1 and 2.

muness:
## Gate report — SHIP / review

**Gate/role:** Ship gate, reviewer
**Verdict:** pass

### Aim and scope checked

Aim: `src/mcp/` decomposed by concern, a contract harness that pins today's surface, no behavior change. Scope checked: the four commits, all acceptance criteria, and whether the dissent's two blocking findings actually landed.

### Dissent P1s — both discharged, both demonstrated

**Raised protocol version.** Was invisible: the SDK negotiates down, so with the client asking for `2025-11-25`, a declared version of `2026-06-18` still yields `2025-11-25` and the snapshot passes. Fixed by reading the value off `mcp::server_details()` directly — the seam extracted earlier in this PR and, as the dissent pointed out, not used by the one test that needed it. Demonstrated biting:

```
$ # declared version lowered to V2025_06_18
$ cargo test --test mcp_contract declared_protocol_version
assertion `left == right` failed: the declared protocol version changed. The wire
snapshot cannot catch a raised version, so this assertion is the one that matters
  left: "2025-06-18"
 right: "2025-11-25"
```

The raise direction cannot be exercised today — rust-mcp-sdk 0.8.3 has no variant above `V2025_11_25` — which is exactly the situation the guard exists for: the next SDK bump. The test also now pins the declared *capability set* (tools present; resources, prompts, logging absent), so #394 cannot drift into declaring what #397 owns.

**Orphan-field hole.** Fields unreachable over the wire were hand-listed under "keep in sync with `McpSearchResult`". Renaming `subtitle` would have left `subtitle` classified and the new name unnoticed — on the exact struct #396 exists to change. The inventory is now collected by serializing the production types. Demonstrated biting:

```
$ # McpSearchResult::subtitle renamed to caption
$ cargo test --test mcp_contract no_tool_returns_an_unclassified_field
MCP tools return field(s) ["caption"] that are neither consumed by a tool input
nor classified as display-only.
```

Both demonstrations were fully reverted; the working tree is clean and `src/mcp/server.rs` / `src/mcp/types.rs` are byte-identical to `7bb0470`.

### The snapshot bites (AC requirement), demonstrated after the split

```
$ # hifi_zones description changed to "... UPnP, Sonos)" in src/mcp/tools/zones.rs
$ cargo test --test mcp_contract tools_list_matches_fixture
MCP contract drift in tests/fixtures/mcp_tools.json

line 6:
  fixture:     "description": "List all available playback zones (Roon, LMS, OpenHome, UPnP)",
  actual:      "description": "List all available playback zones (Roon, LMS, OpenHome, UPnP, Sonos)",

The MCP surface changed. If that was intentional AND approved, regenerate with:
  UPDATE_MCP_FIXTURES=1 cargo test --test mcp_contract
Otherwise this is the bug: epic #392 is additive-only.
```

Reverted; 45/45 green after.

### Withdrawing an overstated claim (gate-1 dissent P2)

The solution-gate review justified Option C partly on merge conflicts, inheriting "five downstream issues editing different files instead of one" from the issue text without checking it. The dissent traced the epic's actual waves against Option A and found exactly one conflicting pair (wave 4: #399 and #400, both adding tools), which Option C does not remove either.

**That justification is withdrawn.** The claim that survives inspection is narrower: #395 rewrites result construction across all ten tool arms, and reviewing that as ten hunks in five small family files beats reviewing one 330-line hunk. Option C stays selected on review-surface grounds, plus the routing centralization and the `server_details()` / `list_tools()` seams, which are worth having regardless. Anyone reading this PR later should not take it as evidence that 11 files were needed to avoid conflicts Option A would not have produced.

### Acceptance criteria

| AC | Status |
|---|---|
| `src/mcp/` decomposed, every call site moved, nothing duplicated | ✅ 11 modules; 4 call sites in `src/main.rs` unchanged via re-exports |
| Zone routing in exactly one place, Roon default preserved | ✅ verified by grep; `ZoneTarget` |
| `tools/list` snapshot vs committed fixture | ✅ `tests/fixtures/mcp_tools.json` |
| `initialize` snapshot incl. capabilities + instructions | ✅ `tests/fixtures/mcp_initialize.json` + direct assertion |
| Input schemas valid JSON Schema, documented params present | ✅ structural checks + `EXPECTED_TOOL_PARAMS` |
| No-orphan-field assertion | ✅ derived from types, reasons mandatory |
| Tool-call round-trip per adapter family | ⚠️ **partial** — see below |
| `cargo test --test mcp_contract` passes; fixture regenerable | ✅ 45/45; `UPDATE_MCP_FIXTURES=1` |
| Breaking a description fails the snapshot | ✅ demonstrated above |
| `cargo clippy --all-targets` clean | ❌ **not achievable** — see below |
| Existing suites still pass | ⚠️ two pre-existing failures, proven |

### Verification evidence

```
$ cargo test --test mcp_contract
test result: ok. 45 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

$ cargo test --lib
test result: ok. 95 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

$ cargo clippy -- -D warnings          # what CI runs
Finished `dev` profile [unoptimized + debuginfo] target(s) in 8.79s   (exit 0)

$ dx build --release --platform web --features web    # rustup toolchain, sccache off
  1.91s  INFO Client build completed successfully! 🚀
137.73s  INFO Server build completed successfully! 🚀
```

All other test targets green: `adapter_integration` 42, `protocol_schema` 41, `client_harness` 41 (minus 1 pre-existing), `protocol_integration` 32 (minus 1 pre-existing), `zones_sha_integration` 20, `volume_safety` 14, `volume_step` 9, `architecture_lint` 8, plus every lint target.

Behavior preservation, the load-bearing evidence — diffing the fixtures across the split commit produces no output:

```
$ git diff 92f79ab..7bb0470 -- tests/fixtures/
(no output)
```

45 tests written against the 855-line module; the module was replaced underneath them; neither fixture moved.

### Honest gaps

**`cargo clippy --all-targets` cannot be clean.** `src/lib.rs` sets `#![deny(clippy::unwrap_used)]`, `expect_used` and `panic` crate-wide, which the `#[cfg(test)]` modules inside `src/` violate 76 times — `src/config/mod.rs` (25), `src/adapters/lms_discovery.rs` (17), `src/adapters/lms.rs` (15), and others. All pre-existing; the count is identical before and after this branch. `src/mcp/**` contributes **zero** clippy findings at any severity. This AC needs rewording in a future issue or a separate lint-debt cleanup; it is not met and cannot be met here.

**Two pre-existing test failures.** `client_harness::shared_endpoints::get_hqp_discover` and `protocol_integration::api_endpoints::hqp_discover_returns_json`, both 500 instead of 200. Both hit `/hqp/discover`, which does live UDP SSDP broadcast and fails in this sandbox. Proven two independent ways — stashing this branch's changes, and re-running at base commit `c47d36a`:

```
$ # at base commit c47d36a
$ cargo test --test protocol_integration api_endpoints::hqp_discover_returns_json
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 32 filtered out
```

Neither path touches `src/mcp`.

**OpenHome and UPnP have no protocol round-trip.** Neither adapter has a `configure()`; they are SSDP-discovery only. Covered: mocks confirmed live, and each prefix confirmed to reach its own adapter via that adapter's distinctive refusal. Not covered: SOAP transport through the mock. Adding a discovery seam is a behavior change and out of scope here.

**Roon's success path is untested, and this is a gap in the central claim rather than generic residual risk** (dissent P3, accepted). `MockRoonCore` is an in-memory state holder, not a server, so the Roon adapter cannot be driven by it. Every Roon assertion in this PR is an error-path assertion. Concretely: `hifi_search`'s Roon branch maps `item.title` to `title` and `item.subtitle` to `subtitle`, and if the split had swapped those two, **nothing in 45 tests would notice**. That mapping was preserved by inspection, not by test. Building a Roon mock server is its own project.

### Rollback

Fully self-contained; nothing outside `src/mcp/`, `tests/mcp_contract.rs`, two mock servers, and two new fixtures is touched.

- **Whole branch:** revert the four commits `53d5ef6`, `7bb0470`, `92f79ab`, `169551a` — or simply do not merge. No migration, no config, no persisted state, no HTTP surface.
- **Split only, keeping the tests:** revert `7bb0470` alone. The tests were written against the unsplit module and pass against it, so this is a supported state, not a broken one. That is the practical value of the commit ordering.
- **Runtime risk if a defect is found post-merge:** the MCP endpoint is one Axum route (`/mcp`); disabling it is a `src/main.rs` route removal. No client depends on the split.
- **`tests/fixtures/api_routes.txt` untouched**, so no `api-change-approved` label is needed and the API contract test is unaffected.

### Follow-up filed

The `MockHqpServer` fidelity bug means `adapter_integration.rs::hqp_mock_responds_to_getinfo` still hand-rolls the two-line workaround and still carries a comment blaming "a complex TCP protocol" for what was a mock bug. Flagged for a separate test-only change.

### Disposition

Pass. Not merged, no labels added, no force-push. Awaiting operator approval per AGENTS.md and #392.

### Residual risk

Ranked by how likely each is to actually bite:

1. **`UPDATE_MCP_FIXTURES=1` erases the guardrail in one command**, and feature-branch PRs here run no CI, so nothing forces a human to look at a red snapshot. The failure message argues its own case ("Otherwise this is the bug: epic #392 is additive-only"), which is the only defence available inside this issue.
2. **Green means "unchanged", never "correct".** The two behaviors this most firmly freezes are defects: unprefixed zone ids routing to Roon, and search results with no addressable identity. Both are labelled at the point of the defect in `routing.rs`, `types.rs` and `FIELD_ROLES`. Comments lose over time.
3. **Roon's success path** — above.
4. **Derive-macro output is not language-guaranteed stable** across compiler or SDK versions. The fixtures are the control; a spurious diff after a toolchain bump is possible and would look like a contract break.

muness:
## Gate report — SHIP / dissent

**Gate/role:** Ship gate, dissent
**Verdict:** block

The split is right, the routing centralization is right, and the two execute-gate P1s were genuinely fixed and genuinely demonstrated. Blocking on one hole the review did not look for, and raising one design flaw that will bite the first time this guardrail is used in anger.

### Contrary case

**P1 — Three of the four public functions in `src/mcp/mod.rs` have zero test coverage, including 55 lines of real logic. The review inventoried acceptance criteria and never asked what in the refactored module is untested.**

The harness drives `/mcp` through `tower::oneshot`, which is good. But it only ever issues POSTs. Never exercised:

| Function | Lines | Coverage |
|---|---|---|
| `handle_mcp_post` | ~30 | ✅ every test |
| `handle_mcp_get` | ~25 | ❌ **none** |
| `handle_mcp_delete` | ~23 | ❌ **none** |
| `auto_recover_session` | ~55 | ❌ **none** |

`auto_recover_session` is not glue. It reads the session header, checks the store, synthesizes an entire `initialize` request with a hardcoded `2024-11-05` payload, extracts the new session id from the response headers, and rewrites the client's headers. It exists because real MCP clients hold stale sessions across a server restart — which is to say it exists because of a bug someone hit in production.

The split moved all of it verbatim. **If it had been broken in the move, nothing in 45 tests would notice.** The PR's headline claim is behavior preservation for `src/mcp/`, and the single most intricate function in the module is asserted by nothing but my having read it. That is precisely the standard the execute-gate dissent rejected for the dispatch match, and it is being accepted here one layer up.

It is also cheap to fix, which makes it worse to ship. A POST carrying a bogus `mcp-session-id` should still succeed, because recovery should transparently mint a new session. That is roughly fifteen lines. GET (the SSE stream) and DELETE (session teardown) are similarly cheap to smoke.

**Required:** cover session recovery, and smoke GET and DELETE.

**P2 — The fixture format conflates "a tool was added" with "an existing tool changed", and epic #392's wave 4 walks straight into it.**

`tests/fixtures/mcp_tools.json` is a JSON **array**. `tools/mod.rs` tells implementers to append rather than insert, so that existing entries stay put. That advice fails exactly when it is needed:

#399 (browse) and #400 (queue) run **in parallel** by the epic's own plan, and each appends a tool. Whichever merges second gets a fixture conflict, and then has to regenerate a fixture that must now contain the other's tool as well. So the first real exercise of this guardrail is two people, both with entirely legitimate reasons, both running `UPDATE_MCP_FIXTURES=1`. That is how reflexive regeneration becomes a habit — and the review lists exactly that as residual risk #1 without noticing the PR's own design guarantees it happens in wave 4.

The deeper problem is that one array blob cannot distinguish the two things the epic cares most about telling apart:

- a tool was **added** — additive, fine, should be a purely additive diff
- an existing tool's description or schema **changed** — forbidden, should be a loud diff

Key the fixture by tool name instead of positionally, and adding a tool adds a key while every existing key stays byte-identical. Then pin the advertised **order** in a separate, tiny assertion. Additive changes produce additive diffs; modifications produce modification diffs; and the two stop being the same event.

This is the right moment to change it — nothing depends on the fixture format yet, and every downstream issue inherits whatever ships here. Doing it after #399 and #400 have both regenerated is doing it too late to matter.

**P4 — The `DisplayOnly` reason check measures length, not content.** `reason.len() > 15` is a proxy, and proxies get gamed. In its defence it *did* catch a real case: "Roon core label" is exactly 15 characters, the test failed, and the reasons were rewritten to say something. So it is doing work. But a future implementer facing a red test will reach for padding before insight, and the check cannot tell the difference.

**P4 — `LmsHarness::start()` polls for up to 10 seconds.** Two tests use it; the suite runs in 12s locally. On a loaded machine the adapter-connect-to-aggregator propagation could exceed the budget and flake. Not worth restructuring, worth knowing.

### Where the review is right

Withdrawing the merge-conflict justification for Option C is the right call and unusual — a review that had already shipped a claim went back and narrowed it because inspection did not support it. The narrower claim (review surface, not conflict avoidance) is defensible and is what the record should say.

Reporting `cargo clippy --all-targets` as **not met and unmeetable**, with the 76-error breakdown and the observation that the count is identical before and after, is the correct handling of an acceptance criterion that cannot be satisfied. Substituting a narrower command and calling it clean would have been the easy path.

Proving the two pre-existing failures twice is right, and stating Roon's untested success path as a gap in the central claim rather than filing it under general residual risk is the correct demotion of the PR's own guarantee. The concrete version — swap `title` and `subtitle` in Roon's branch and 45 tests stay green — is more useful than any amount of hedging.

The rollback note is genuinely actionable, and the observation that reverting the split alone leaves a supported state (because the tests predate it) is the commit ordering paying off a second time.

### Required actions

1. **(P1, blocking)** Cover `auto_recover_session`; smoke `handle_mcp_get` and `handle_mcp_delete`.
2. **(P2, blocking)** Key `mcp_tools.json` by tool name and pin the advertised order separately, so additive changes yield additive diffs before #399 and #400 inherit the array format.
3. **(P4)** Note the length-proxy weakness of the `DisplayOnly` check where the check lives.
4. **(P4)** Note the polling budget in `LmsHarness`.

### Disposition

Block on 1 and 2. Neither is large; both are in the harness rather than in production code, so neither risks the behavior-preservation claim; and both get materially harder to do after other issues build on this foundation.

### Residual risk after actions

Unchanged in character: this pins what the code does, and two of the things it pins hardest are defects. What actions 1 and 2 buy is narrower and worth having anyway — that the module's least-tested function stops being its most intricate one, and that the guardrail can tell the difference between the change the epic permits and the change it forbids.

muness:
## Gate report — SHIP / review, addendum after `06821cb`

**Gate/role:** Ship gate, reviewer (re-review)
**Verdict:** pass

The ship-gate dissent blocked on two findings. Both landed in `06821cb`, which touches **no production code**:

```
$ git diff 53d5ef6..06821cb --name-only
tests/fixtures/mcp_tools.json
tests/mcp_contract.rs
```

### P1 — transport coverage

The dissent was right that the review inventoried acceptance criteria and never asked what in the refactored module was untested. Three of four public functions in `src/mcp/mod.rs` had none, including `auto_recover_session`'s 55 lines of stale-session logic — the most intricate code in the module, moved verbatim by the split.

Now covered:

- `a_stale_session_id_is_transparently_recovered` — a POST carrying a session id the server has never seen still returns the full 10-tool list, with no error surfaced to the client
- `get_and_delete_are_wired_and_session_aware` — GET and DELETE refuse a missing session; DELETE with a live session succeeds; `PATCH` returns 405 as the control case, which is what proves the other statuses came from the handlers rather than from axum's router

Writing it surfaced a real asymmetry the review had not seen: **with an unknown session, DELETE returns 404 but GET returns 500.** GET is arguably wrong — a stale SSE reconnect is a client-side condition, not a server fault — but it is today's behavior, so it is recorded rather than corrected, with a note so a later fix reads as intentional.

That is the third time in this PR that writing a test for something believed uninteresting turned up behavior nobody had characterized.

### P2 — fixture keyed by name

`mcp_tools.json` is now a map keyed by tool name. The dissent's argument holds and the review had missed it: an array conflates "a tool was added" (permitted by #392) with "an existing tool changed" (forbidden), and the epic's wave 4 lands #399 and #400 in parallel with each appending a tool — so the array format guaranteed that the guardrail's first real use would be two legitimate regenerations. Keyed, adding a tool adds one key and every existing key stays byte-identical. Order is pinned separately in `tools_list_order_is_pinned`, so a reordering and a description change fail different assertions.

Content equivalence was verified rather than assumed — rebuilding the keyed map from `7bb0470`'s array is an exact match, 10 entries either way:

```
array entries: 10 | keyed entries: 10
CONTENT IDENTICAL: True
```

The reformatted fixture was confirmed to still bite on a changed description (`line 175`), and the tool name is dropped from each value since it is the key, so renaming a tool cannot masquerade as one added key plus one removed key with identical bodies.

P4s actioned: the `DisplayOnly` length check now says at its own site that it measures length rather than content and that the fix for a failure is a better reason, not a longer one; `LmsHarness` documents its 10s wall-clock propagation budget.

### Final verification on the shipped tree

```
$ cargo test --test mcp_contract
test result: ok. 48 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

$ cargo test --lib
test result: ok. 95 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

$ cargo clippy -- -D warnings
Finished `dev` profile [unoptimized + debuginfo] target(s) in 10.83s     (exit 0)

$ cargo clippy --test mcp_contract
(no findings in tests/mcp_contract.rs)

$ dx build --release --platform web --features web
  1.64s  INFO Client build completed successfully! 🚀
121.08s  INFO Server build completed successfully! 🚀
```

Unchanged and still true: `cargo clippy --all-targets` is not clean and cannot be (76 pre-existing errors in `src/`'s own test modules under the crate-wide `#![deny]`s, identical count before and after this branch, zero from `src/mcp/**`); and two `/hqp/discover` UDP-SSDP tests fail pre-existing, proven at base commit `c47d36a`.

### Disposition

Pass. All three dissents' blocking findings are discharged. Not merged, no labels, no force-push.

### Note for the record

The dissents blocked three times and each block produced a fix that mattered:

| Gate | Blocking finding | Fixed in |
|---|---|---|
| solution | snapshots guard the low-risk half; the 330-line dispatch and its three lookup tables were nearly uncovered | `92f79ab` |
| execute | a *raised* protocol version was invisible; orphan-field inventory hand-listed exactly where #396 will edit | `53d5ef6` |
| ship | three of four public functions in `mod.rs` untested; array fixture conflates additive with modifying changes | `06821cb` |

Every one was a hole in a guarantee the review had already stated as holding. That is the point of running the two roles separately, and it is worth noting that in each case the fix used a seam this PR had already built and then failed to use.

muness:
Reviewed the report. Three things resolved at the epic level rather than in this PR:

**The `--all-targets` AC was wrong, and I have corrected it in all twelve issues.** You are right that it is unmeetable: `src/lib.rs:20-22` denies `unwrap_used`/`expect_used`/`panic` crate-wide and src's own test modules violate it 76 times, pre-existing. Every issue now specifies `cargo clippy -- -D warnings` — what CI actually runs — and explicitly says not to chase `--all-targets`. This was my error propagated across the epic; you were the first to hit it and eight more agents would have.

**The withdrawn merge-conflict justification is the right call.** A layout defended on a reason that does not hold is a layout that will be re-litigated later. Standing it on review surface, routing centralization, and the `server_details()`/`list_tools()` seams is more durable — and the dissent tracing the actual epic waves to check the claim is exactly what that role is for.

**The untested Roon success path is now #408.** Your framing of it as a gap in the central claim rather than generic risk is what made it actionable: a swapped `title`/`subtitle` passing 48 tests is a specific, demonstrable hole. #396's design gate independently hit the same wall from the browse side. #408 requires a regression test that catches precisely that swap, and must emit `Parsed::Error(RoonApiError::BrowseInvalidItemKey)` on demand so #405 can test error correlation.

Your two out-of-scope findings are #409 (the stale `adapter_integration.rs` workaround with its misattributing comment, and `GET /mcp` returning 500 where `DELETE` returns 404). Recording rather than fixing them was correct for a behavior-preserving PR.

The evidence that matters here is the commit order — 44 tests written and passing against the unsplit module, then the module replaced underneath them, with `git diff 92f79ab..7bb0470 -- tests/fixtures/` empty. That is a real proof of preservation rather than an assertion of one.

Not merging; that is the operator's call per AGENTS.md. #395 is now running against this branch.

muness:
## CI caught what local verification could not

`e877066` is `cargo fmt`, nothing else. Recording why it was needed, because the reason is more interesting than the fix.

### What I got wrong

I verified `cargo clippy -- -D warnings` and treated it as the lint gate. The Lint job runs **two** commands (`.github/workflows/build.yml:365-369`):

```yaml
      - name: Check formatting
        run: cargo fmt --check

      - name: Run clippy
        run: cargo clippy -- -D warnings
```

`cargo fmt --check` was never in my recipe, so I never ran it, and clippy does not subsume it — they check disjoint things. CI failed on diffs in `src/mcp/handler.rs`, `src/mcp/routing.rs`, `src/mcp/tools/transport.rs` and ~20 sites in `tests/mcp_contract.rs`.

I also asserted in the original PR body that feature-branch PRs here run zero CI. That holds only when the *base* is another feature branch. This PR targets `v3`, which matches `pull_request: branches: [master, v3]`, so it gets the full set. **CI is a real gate on this PR and I said otherwise.** Both claims are now corrected in the PR body.

The complete gate, for whoever picks up the next issue in #392:

| Job | Commands (after `make css`) |
|---|---|
| **Lint** | `cargo fmt --check`, `cargo clippy -- -D warnings` |
| **Test** | `cargo test --workspace` |
| **Build** | WASM assets, Linux x64, smoke test, Synology package contract |

### Re-proving the preservation claim rather than assuming it

`cargo fmt` should not be able to change serialized output. But the whole value of this PR is the claim that `tools/list` and `initialize` are byte-identical across the split, so I re-proved it rather than reasoning about it.

`cargo fmt` touched exactly the four files this branch added or changed, and nothing pre-existing — the base was already fmt-clean, so this is entirely my formatting:

```
$ git show --stat e877066
 src/mcp/handler.rs         |   8 +-
 src/mcp/routing.rs         |   5 +-
 src/mcp/tools/transport.rs |  14 +++-
 tests/mcp_contract.rs      | 188 +++++++++++++++++++++++++++++----------------

$ git status --short tests/fixtures/     # after cargo fmt
(empty — no fixture touched)
```

Fixtures still match what the **unsplit** 855-line module emitted at `169551a`:

```
$ git diff 169551a -- tests/fixtures/mcp_initialize.json
(empty)

$ # tools/list: rebuild 169551a's array into today's keyed form and compare
pre-split array entries: 10 | current keyed entries: 10
TOOLS/LIST CONTENT IDENTICAL TO PRE-SPLIT: True
```

And the full local sweep on the formatted tree:

```
$ cargo fmt --check
(exit 0)

$ cargo clippy -- -D warnings
Finished `dev` profile [unoptimized + debuginfo] target(s) in 9.12s      (exit 0)

$ cargo test --test mcp_contract
test result: ok. 48 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

$ cargo test --lib
test result: ok. 95 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

### All 26 checks green

```
Lint                       pass    49s
Test                       pass    1m48s
Build WASM Assets          pass    4m23s
Build Linux x64            pass    3m48s
Smoke Test Binary          pass    5s
Synology Package Contract  pass    10s
Build Summary              pass    7s
Post Artifact Links        pass    4s
Plan                       pass    3s
CodeRabbit                 pass    (review rate limited)
+ 16 release jobs skipped (tag-gated)
```

### A second correction, from CI's Test job

I reported two "pre-existing test failures" on `/hqp/discover`. **CI runs both green:**

```
$ gh run view 30674744344 --job 91299539795 --log | grep hqp_discover
test shared_endpoints::get_hqp_discover ... ok
test api_endpoints::hqp_discover_returns_json ... ok
```

They do live UDP SSDP broadcast, which my sandbox blocks. My evidence — they fail at base commit `c47d36a` and with this branch stashed — was sound for the question I was actually asking, *does my branch cause these?* I then generalized it into "the repo has two failing tests", which does not follow: a test that fails on every commit in one environment is evidence about the environment, not the repo. Corrected in the PR body.

Both mistakes have the same shape: I verified a claim against the evidence available to me and then stated it more broadly than that evidence supported. Worth flagging because this PR's entire product is a claim about preservation backed by evidence, and the standard it sets for itself is the standard its own prose should meet.

### Residual risk, revised

Residual risk #1 said `UPDATE_MCP_FIXTURES=1` can erase the guardrail with nothing forcing a human to look, *because there is no CI*. That premise was wrong — `cargo test --workspace` runs on PRs targeting `v3`, so a red snapshot does block a PR. The risk is real but narrower than stated: CI cannot distinguish a legitimate regeneration from a lazy one, so the failure message still has to make its own case to the person reading it.

Unchanged: not merged, no labels, no force-push. #395 can build on `e877066`.

Issue #408, opened 2026-07-31
Test infrastructure: MockRoonCore is not a server, so no Roon success path is asserted


muness:
OH: 80222d6d

Parent epic: #392
Found by: the #394 execute gate and the #396 solution-space gate, independently

## Problem Statement

**Current framing:** Add more Roon tests.

**Reframed as:** `MockRoonCore` is not a server. It holds state, but nothing speaks Roon's protocol to it, so **no test exercises a Roon success path end to end.** The #394 gate put this precisely: if the module split had swapped `title` and `subtitle` in Roon's search mapping, none of its 48 contract tests would have noticed. The #396 gate hit the same wall from the other side — there is no browse mock, so the ref design cannot state an honest Roon test strategy.

**The shift:** From "Roon behavior is covered because Roon code is called" to "Roon's wire behavior is asserted against something that answers like Roon."

Two agents found this independently while doing unrelated work, which is the signal that it is infrastructure and not a per-issue concern.

### Constraints

- **Hard:** Must support the paths the epic actually depends on: search (including the `multi_session_key` minting/consuming sequence), browse and load with levels and `pop_all`, `play_item`, and error emission — specifically `Parsed::Error(RoonApiError::BrowseInvalidItemKey)`, which #405 needs to test correlation against. Must not require a real Roon Core or network. Must not become a second implementation of Roon's semantics that can drift from the pinned fork — where behavior is uncertain, record real responses rather than invent them. No production-code changes beyond what testability genuinely requires, and any such change is called out separately in the PR.
- **Soft:** Whether this extends `MockRoonCore` or introduces a separate protocol-level fake; whether responses are hand-built or recorded from a live Core; how much of the Roon transport surface is covered in the first pass.

### What this framing enables

Honest Roon tests for #396 (refs), #399 (browse), and #405 (browse errors) — all three of which currently have to choose between unit-scoping their logic or asserting nothing. It also retires the standing caveat that Roon's success path is unverified.

### What this framing excludes

Reimplementing Roon Core. Testing against the operator's live Core in CI. Changing adapter behavior — this issue builds the instrument, it does not fix what the instrument finds. Anything it does find gets its own issue.

## Acceptance criteria

- [ ] A test-only Roon fake that answers the adapter's actual protocol, sufficient for: a search returning known items, a browse/load sequence across at least two levels including `pop_all`, and `play_item` resolving to a playable action.
[...]

Issue #396, opened 2026-07-31
MCP refs: opaque content references that a play tool actually accepts


muness:
OH: 80222d6d

Parent epic: #392

## Problem Statement

**Current framing:** Return `item_key` from search so tools can play a specific result.

**Reframed as:** `hifi_search` returns `{title, subtitle}` and nothing else, so the only path from "found it" to "playing it" is handing a title back to `hifi_play`, which re-runs the search and takes the first match. The client cannot address the third result, cannot distinguish two same-titled albums, and cannot tell whether what played is what it picked.

**The shift:** From title round-tripping to opaque, server-resolved references — the client holds a token it cannot misinterpret, and exactly one tool consumes it.

This is the keystone of #392. It is also where the repo's history matters: AGENTS.md forbids orphaned fields *and names `item_key` as the offender*. That guardrail stands. The reason it can be satisfied now is that `RoonAdapter::play_item(item_key, zone_id, action)` exists and is already exposed over HTTP at `/roon/play_item` — so a consuming tool is available. **A ref must not ship without the tool that accepts it, in the same PR.**

### Lifetime is per target kind, not per provider

The central design tension, and the thing two earlier drafts of this issue got wrong in opposite directions. Neither "Roon is ephemeral, LMS is durable" nor "LMS has no stable ids" is true.

`LmsSearchResult` (`src/adapters/lms.rs:549-564`) carries three addressing paths, and `parse_single_item` (`:1205-1252`) **refuses to emit a result without at least one playback handle** — so LMS addressability is already an invariant. `search_and_play` (`:1376-1430`) already branches across all three; it simply discards `results[1..]`. But they differ in durability:

- `Library { kind, id }` — durable across UHC *and* LMS restarts (`playlistcontrol`); dies only on a rescan that reassigns ids.
- `Url { url }` — mostly durable (`playlist load/add`).
- `GlobalSearchItem { item_id }` — **not durable.** The play call (`:1384-1394`) omits the `search:` term, so LMS resolves it against server-side cached feed state left by the preceding `globalsearch items` call.

And `search()` prefers globalsearch, falling back to `search_library` only when globalsearch returns nothing (`:1096-1098`) — so **in the common case, LMS results reaching MCP are the ephemeral kind.** The durable-id story describes the fallback path. Roon's keys are ephemeral too, scoped to a `multi_session_key`.

[...]

Issue #395, opened 2026-07-31
MCP results: one structured, capability-honest result envelope for every tool


muness:
OH: 80222d6d

Parent epic: #392
Realizes: #221

## Problem Statement

**Current framing:** MCP responses are human-readable strings; make them JSON.

**Reframed as:** Tools answer with prose like `"Play Now: Superstition"` or `"Volume adjusted"`, which states neither outcome nor scope nor observed state — so a model cannot distinguish "accepted" from "possibly failed" and retries. #221 recorded ChatGPT looping on `hifi_play` for exactly this reason. Failures are worse: an unsupported operation and a genuine error are both just text, so a model cannot learn a limit, only re-hit it.

**The shift:** From per-tool ad-hoc strings to one result envelope every tool fills, carrying outcome, scope, observed state, and — when refused — *why*, in a form a model can act on.

### Constraints

- **Hard:** Additive — every tool keeps returning its existing human-readable text content; structured data is *added* alongside, not substituted, so existing clients see no change. Refusals distinguish "this provider cannot do this" from "this failed" from "your input was invalid." Observed state, where reported, is read back from the aggregator rather than assumed from a successful command. No HTTP changes.
- **Soft:** Envelope field names, whether structured output rides `structuredContent` or a JSON text block, and how much observed state each tool returns.

### What this framing enables

Models that stop retrying accepted commands; capability limits that teach instead of just failing; a single place for later issues (REFS, BROWSE, QUEUE, CAPABILITIES) to attach their payloads; and the vocabulary #360 will need for pending/verified/failed outcomes.

### What this framing excludes

Changing or removing any existing text response. Inventing verification the adapters cannot do — if a backend does not confirm, say `accepted`, do not claim `verified`. Reworking error handling outside `src/mcp/`.

## Acceptance criteria

- [ ] One envelope type used by all tools, carrying at least: outcome (e.g. `ok` / `accepted` / `unsupported` / `invalid` / `error`), the scope acted on (zone id and resolved zone name, provider), the action and its parameters as understood by the server, and observed state where available.
- [ ] `unsupported` results name the provider and the operation, and say what *is* available instead — e.g. volume on an OpenHome zone points at what that zone does support.
- [ ] `invalid` results name the offending parameter and its accepted values.
- [ ] Every one of the 10 existing tools is migrated; the human-readable text of each is unchanged (proven by the #392 FOUNDATION (#394) snapshot still passing).
[...]

PR #412, opened 2026-07-31
fix(roon): stop discarding Core browse errors, so a stale item key is not a 10s ambiguous timeout (#405)

### LINKED ISSUE #405: Roon adapter: stop discarding browse errors, which turns a stale item key into a 10s ambiguous timeout
OH: 80222d6d

Parent epic: #392
Discovered by: the solution-space gate on #396

## Problem Statement

**Current framing:** MCP needs a distinct "your reference expired" outcome, so add one in the MCP layer.

**Reframed as:** When Roon Core rejects a browse item key, the adapter silently discards the rejection and lets the caller block until a 10-second timeout, which then reports "Browse request timed out" — the same message an unreachable Core produces. A recoverable, instantly-known condition is presented as an ambiguous failure, ten seconds late. No MCP-layer design can fix this: a ref table knows when *UHC* forgot a reference, not when *Roon Core* did.

**The shift:** From "expiry is a message the MCP layer composes" to "the adapter must not throw away the error the protocol already gave it."

### Evidence

- The pinned fork (`open-horizon-labs/rust-roon-api@ohc/main`, checkout `06dd807`) reports this precisely: `src/browse.rs:174-188` yields `Parsed::Error(RoonApiError::BrowseInvalidItemKey((req_id, multi_session_key)))`.
- UHC never handles it. `grep -rn "Parsed::Error\|RoonApiError" src/` returns **nothing**; the event loop's catch-all `_ => {}` at `src/adapters/roon.rs:1911` swallows it.
- The pending oneshot is therefore never resolved, so the caller waits out `BROWSE_TIMEOUT` (10s, `src/adapters/roon.rs:39`, awaited at `:614` and `:660`) and receives a generic timeout.
- This is **not** an MCP-only defect. It degrades `/roon/browse` and `/roon/play_item` today, for the web UI as much as for any future MCP client.

### Constraints

- **Hard:** Correlation is the crux and it is concurrency-sensitive. `pending_browses`/`pending_loads` are keyed by `req_id` but scanned by session key (`src/adapters/roon.rs:1872-1877`), and the error payload carries **both** — so the error arm must resolve the right waiter without racing the success arms or leaving an entry stranded. Every request that can now hang must either be resolved or time out as it does today; no new leak, no double-resolve. The repo's cancellation, channel, and lock lint suites must pass (`spawn_cancellation_lint`, `unbounded_channel_lint`, `await_in_lock_lint`). Behavior-additive: a valid browse or load is unaffected. No HTTP route changes.
- **Hard:** Failure classes must be distinguishable to callers: item key rejected by Core (recoverable, instant) vs. Core unreachable or slow (the existing timeout) vs. Browse service not connected (already handled). Collapsing these back together defeats the issue.
- **Soft:** Whether other `RoonApiError` variants are routed in the same pass; the error type exposed to callers; whether a browse mock is added here or the tests are unit-scoped.

### What this framing enables

#396's expiry criterion becomes achievable; `/roon/browse` stops taking ten seconds to say "stale key"; and the first handling of `Parsed::Error` in this codebase, which every later Roon feature can extend.

MCP-layer changes. Ref tables and TTLs — that is #396. Redesigning the browse session model. Handling error variants that no caller can act on.

## Acceptance criteria

- [ ] `Parsed::Error(RoonApiError::BrowseInvalidItemKey(..))` resolves the correct pending browse/load waiter with a typed, recoverable error, instead of being dropped.
- [ ] A rejected item key returns **promptly** — no `BROWSE_TIMEOUT` wait — and is distinguishable by callers from an unreachable Core and from Browse-not-connected.
- [ ] Correlation is correct under concurrency: a test with several in-flight browse/load requests, where one is rejected, proves the right waiter is resolved and the others complete normally.
- [ ] No stranded map entries and no double-resolve; the cancellation/channel/lock lint suites pass.
- [ ] `/roon/browse` and `/roon/play_item` surface the new condition meaningfully rather than as a timeout.
- [ ] Test strategy stated explicitly in the PR: either the Roon mock is extended to emit this error, or the correlation logic is unit-scoped — say which and why.
- [ ] **Answer the empirical question this uncovered:** does an `item_key` minted in one `multi_session_key` resolve inside another? `play_item()` (`src/adapters/roon.rs:1094-1113`) mints a *fresh, unrelated* session key and browses the caller's key inside it, so the repo already assumes portability — but `/roon/play_item` has **zero in-repo callers and no test** (`src/main.rs:364`), so it may never have worked. Verify against a live Core and record the finding in the PR. If keys are not portable, say so loudly: it changes #396's design and means `/roon/play_item` is broken.

## Dependencies

**Blocks #396** (its expiry criterion depends on this). Independent of #394/#395 — this is adapter work and can proceed in parallel with the MCP foundation.

## Notes for the implementer

- Read `src/adapters/roon.rs:1860-1915` (the event loop and its pending maps) and the fork's `src/browse.rs:150-200` together before designing; the mismatch between `req_id` keying and session-key scanning is the whole problem.
- Requires the operator's Roon Core for the portability question. Script it, run what you can, and report precisely what needs their rig — do not declare untested behavior working.

muness:
Fixes #405. Parent epic #392. Unblocks #396.

## What was wrong

Roon Core answers an unusable browse or load with a *named error* instead of a result. The pinned fork turns that into `Parsed::Error(RoonApiError::Browse*)`, carrying the originating `req_id` **and** the request's `multi_session_key` (fork `src/browse.rs:174-188`).

UHC never handled `Parsed::Error`. The event loop's catch-all `_ => {}` swallowed it, the pending oneshot was never resolved, and the caller sat out the full 10s `BROWSE_TIMEOUT` before receiving `"Browse request timed out"` — the same message an unreachable Core produces. A condition the Core reported instantly, and which the caller can recover from, arrived ten seconds late disguised as an outage.

The red commit on this branch shows it as a wall-clock fact: `99b3779` runs the new suite in **10.01s**, because the promptness test's waiter genuinely waits out the timeout. After the fix: **0.00s**.

## Correlation approach — `req_id` primary, session key as the guard

`RoonState::route_browse_rejection` (`src/adapters/roon.rs`) keys on the `req_id` in the rejection and cross-checks the session key before resolving.

**Why `req_id` is authoritative.** `Browse::browse`/`Browse::load` return the request id that UHC stores as its map key (fork `src/browse.rs:126-154`); `Moo::send_req` allocates ids from a single monotonic per-connection counter (fork `src/moo.rs:124-135`); `parse_msg` reads that same id back off the wire (`:157`) and hands it to the error variant. So a `req_id` names exactly one submitted request, and `pending_browses` / `pending_loads` cannot both hold it. Lookup is exact and O(1).

**Why not scan by session key, as the success arms do.** The success arms have no choice — `Parsed::BrowseResult` carries no `req_id`. But a session key is shared by *every* request in a browse session: `search()` reuses one across six requests, and `/roon/browse` lets the caller supply its own. Scanning cannot say which request in a session was refused, or even whether it was a browse or a load. The test `load_rejection_resolves_the_load_not_a_browse_sharing_its_session` pins exactly that case — a scan of `pending_browses` would have resolved the wrong waiter.

**Why the session key is still checked.** `Moo::new` restarts the counter at 0 on every reconnect (fork `src/moo.rs:92`). A rejection arriving on a fresh connection can therefore carry a `req_id` that a stale, not-yet-timed-out entry still occupies. On mismatch the waiter is left alone and falls back to `BROWSE_TIMEOUT` — the pre-#405 behavior — because resolving the wrong caller is worse than resolving them late. `ErrorRouting::KeyMismatch` is logged at `warn`.

**Exactly-once, no strand.** Every delivery path `remove`s the entry before sending, so a repeated rejection finds nothing (`a_second_rejection_for_the_same_req_id_cannot_double_resolve`) and no waiter can be resolved twice. A dropped receiver still clears the entry (`rejection_with_a_dropped_receiver_still_clears_the_entry`). `KeyMismatch` deliberately leaves the entry for the caller's own timeout cleanup — the one path that does not remove, and the only correct one that does not.

**Both maps are searched for a full match before any mismatch is reported** (`b24ae86`, found by this branch's own dissent pass). A stale entry with a colliding `req_id` in `pending_browses` must not short-circuit a matching entry in `pending_loads`; that would cost a legitimate waiter its prompt answer for no safety gain. Pinned by `a_mismatched_entry_in_one_map_does_not_hide_a_match_in_the_other`.

## Failure classes, kept apart

| Condition | Before | After |
|---|---|---|
| Item key refused by Core | 500 `Browse request timed out`, after 10s | **404**, immediately, `Roon Core rejected the item key: it is no longer valid in this browse session - search or browse again to get a fresh key (browse session '…')` |
| Zone unknown to Core | 500 timeout, after 10s | **404**, immediately |
| Bad browse level | 500 timeout, after 10s | 500, immediately, with the Core's reason |
| Core-side unexpected error | 500 timeout, after 10s | 500, immediately, with the Core's reason |
| Core unreachable or slow | 500 timeout, after 10s | unchanged |
| Browse not connected | 503, immediately | unchanged |

Callers classify with `RoonBrowseError::from_error(&err) -> Option<&RoonBrowseError>` — no message-string matching. `RoonBrowseErrorKind::is_recoverable()` marks the two kinds a caller fixes by re-acquiring its reference and retrying. That boundary is a judgment, not a fact: `ZoneNotFound` is also caller-fixable, by naming a different zone. The line drawn here is "retry after re-acquiring your reference" versus "fix your input", and #395's envelope may reasonably want them as one category. The error travels inside the `anyhow::Error` the adapter methods already return, so `browse`, `load`, `search`, `play_item` and `search_and_play` keep their signatures and it survives `?` unchanged.

`RoonBrowseErrorKind::explain()` never contains the word "timeout", and a test asserts it.

## Who this actually reaches today

Worth being precise, because the headline surface is not the one with consumers:

- **`/roon/browse` and `/roon/play_item` — the routes #405 is about — have zero in-repo callers.** The Dioxus UI has no browse page. Their consumers, if any, are external: the iOS app, the knob, the operator's own tooling. `/roon/play_item` may never have been exercised by anything, which is the same evidence that makes its key-portability suspect.
- **`search()` is the path with known callers.** It calls `browse()`/`load()` six times, and MCP's `hifi_search` and `hifi_play` go through it. A rejection anywhere in that sequence used to cost 10s and now returns immediately. This is the change most likely to be noticed.
- **`/roon/image` is used by the web UI.** Cover art the Core refuses now fails fast instead of blocking for its own 10s timeout.

So the fix is worth having on all three counts, but no user-facing timeout is being removed from a route this repo is known to call — the routes named in the issue are external surface.

## Scope decisions

**All four browse rejections are routed, not only `InvalidItemKey`.** `InvalidItemKey` is the traced case; the other three are routed by symmetry, not by observation — nothing here establishes that a Core emits them, or with those exact names. They leak identically through the same catch-all, so routing one and leaving three would have left three 10s hangs behind. #405 lists this as a soft constraint.

**The two image rejections are routed too** — called out separately, as it is beyond the issue's title. `pending_images` has the identical shape (`req_id` key, image key stored) and the identical defect: `ImageNotFound` was dropped, so a missing cover art blocked `get_image` for its own 10s timeout. Routing sends `None`, which `get_image` already reads as `Image not found` (`src/adapters/roon.rs:563`). No type or signature change.

**`RoonApiError` is matched exhaustively, with no wildcard.** If the pinned fork gains a variant, this stops compiling instead of silently resuming the swallow. That is deliberate.

**Success arms untouched.** They cannot be improved without a fork change (no `req_id` on `Parsed::BrowseResult`), and this PR is behavior-additive.

## Test strategy — unit-scoped, deliberately not a Roon mock

Stated explicitly because #405 requires it. **The correlation logic is unit-scoped. No Roon mock was built, and `tests/mock_servers/roon.rs` was not extended.**

`MockRoonCore` holds state and speaks no protocol — nothing in the repo can make the adapter emit a `Parsed::Error`. Building a protocol-level Roon fake is **#408**, which is open and explicitly lists "can emit `Parsed::Error(RoonApiError::BrowseInvalidItemKey)` on demand, so #405 can test error correlation" as its own acceptance criterion. Building one here would duplicate and conflict with that work. #408 had not landed when this was written (no PR on it), so it was not available to use.

18 tests in `src/adapters/roon.rs`'s existing test module (private `RoonState` access), 12 of them new:

- `browse_rejection_resolves_only_the_matching_waiter` — five requests in flight (3 browses, 2 loads), one rejected; asserts the right receiver resolved, the other four still waiting, and only one map entry removed. This is #405's concurrency criterion.
- `load_rejection_resolves_the_load_not_a_browse_sharing_its_session` — the case a session-key scan gets wrong.
- `rejection_with_a_mismatched_session_key_leaves_the_waiter_alone` — the reconnect `req_id`-reuse guard.
- `a_mismatched_entry_in_one_map_does_not_hide_a_match_in_the_other` — the cross-map liveness case above.
- `rejection_for_an_unknown_req_id_is_a_noop`, `rejection_after_the_caller_timed_out_is_a_noop`, `a_second_rejection_for_the_same_req_id_cannot_double_resolve`, `rejection_with_a_dropped_receiver_still_clears_the_entry` — no strand, no double-resolve.
- `every_browse_rejection_kind_is_routed` — all four kinds.
- `a_rejected_item_key_resolves_far_inside_the_browse_timeout` — awaits the real `BROWSE_TIMEOUT` exactly as `browse()` does and asserts it returns in under a second.
- `image_rejection_resolves_the_matching_image_request`, `image_rejection_with_a_mismatched_key_leaves_the_waiter_alone`.
- `a_rejection_is_distinguishable_from_a_timeout_and_a_lost_core`, `only_reference_rejections_are_marked_recoverable`.

**Coverage boundary, stated so nobody over-trusts it** (and repeated in a comment in the test module): these exercise the real routing functions, the real `oneshot` channels and the real `BROWSE_TIMEOUT`. They do **not** exercise the fork's `parse_msg`, nor the three-line result match in `browse()`/`load()` that hands the oneshot payload to the caller. When #408 lands, this suite should gain one end-to-end case through the fake.

## The empirical question: is an `item_key` portable across `multi_session_key`s?

**Not answered. It needs a host with LAN access to the Core — this one does not have it.** Not declared working.

What was established:

- **A live Core is present and answering.** A SOOD query (replicating the fork's `src/sood.rs:120-188` wire format byte for byte, sent from this Mac to `192.168.1.255:9003`) got a reply from `192.168.1.208`: `name=nuc14`, `display_version=2.70 (build 1671) production`, `tcp_port=9150`, `http_port=9330`, `service_id=00720724-5143-4a9b-abac-0e50cba674bb` — the exact extension service id the adapter looks for.
- **But this Mac cannot initiate unicast traffic to it.** `connect()` to `192.168.1.208:9150` and `:9330` returns `EHOSTUNREACH`; `ping` gets no reply; the multicast send to `239.255.90.90` also fails `EHOSTUNREACH`. Identical with the tool sandbox disabled, so it is not the sandbox. Meanwhile `route -n get 192.168.1.208` shows a valid route on `en7` and `arp -n` has the MAC (`88:ae:dd:72:cc:62`) — and the Core's SOOD reply arrived as a *unicast* packet back to this Mac. So inbound unicast works and outbound unicast does not: the shape of a host-level local-network egress restriction, not a broken network. Lifting it means changing a system privacy or security setting, which is the operator's to do, not mine.
- This Mac has also never paired: there is no `~/.config/unified-hifi/roon_state.json`, and no UHC instance is running locally or anywhere reachable.

Even with egress fixed, the first run needs the operator to enable the extension in **Roon Settings → Extensions**, which is theirs to click.

### The exact check, for whoever has the rig

Run UHC against the Core, then:

```bash
UHC=http://localhost:8088

# 1. Mint a key in session A: browse the root, load one level, take an item_key.
curl -s -X POST $UHC/roon/browse -H 'content-type: application/json' \
  -d '{"pop_all":true,"session_key":"portability_A"}' | jq '.items[] | {title, item_key}'

KEY='<paste an item_key from session A>'

# 2. Use it in a *different*, fresh session B - what play_item() does internally.
curl -s -o /tmp/b.json -w 'HTTP %{http_code} in %{time_total}s\n' \
  -X POST $UHC/roon/browse -H 'content-type: application/json' \
  -d "{\"item_key\":\"$KEY\",\"session_key\":\"portability_B\"}"
jq . /tmp/b.json

# 3. Control: the same key re-used in its own session A must still work.
curl -s -X POST $UHC/roon/browse -H 'content-type: application/json' \
  -d "{\"item_key\":\"$KEY\",\"session_key\":\"portability_A\"}" | jq '.list, (.items | length)'
```

Reading the result — and this is only legible *because* of this PR:

- **`HTTP 404` in well under a second**, with `Roon Core rejected the item key`, means **keys are not portable**. Then `/roon/play_item` has never worked (it mints `play_item_<nanos>` and browses the caller's key inside it, `src/adapters/roon.rs:1107-1123`), and **#396's ref record must resolve through the minting session key, not just carry it** — the ref table has to re-drive the browse in the original session, or `play_item` needs a session-key parameter. Say so loudly in #396 before its design is fixed.
- **`HTTP 200` with a `list`** means keys are portable and the existing assumption holds.
- **`HTTP 500` with `Browse request timed out` after 10s** means something else is wrong — with this PR merged, that answer no longer means "stale key".

Before this PR, cases 1 and 3 were indistinguishable, so the question was not answerable from the HTTP surface at all.

Note step 2 uses `/roon/browse`, not `/roon/play_item`: it answers the same question and starts no playback. `/roon/play_item` still has zero in-repo callers and no test.

**That same run also closes this PR's other unverified link.** Nothing on this branch proves that a real Core's error names match the four literals the fork matches on (`InvalidItemKey`, `InvalidLevels`, `UnexpectedError`, `ZoneNotFound` — fork `src/browse.rs:174-188`). If a Core spelled one differently, `parse_msg` would yield `Parsed::None`, the fork would drop it, the caller would time out exactly as before — **and this test suite would still be green.** A fast 404 from step 2 confirms the wire half end to end. Until then, that half rests on a reading of the pinned checkout. It is #408's to pin permanently.

## Verification

**Correction to a premise this PR started with.** The "feature-branch PRs run zero CI" note applies to PRs based on *another feature branch*. This one bases on `v3`, so it gets the full workflow set — **Lint, Build and API Guard all run here, and the checks are meaningful.** The first run caught something local verification had missed: CI's Lint job runs `cargo fmt --check` *before* clippy, and it failed on five hunks in the new test module (`96d3ebd` fixes it). Worth recording, because `cargo clippy -- -D warnings` was clean throughout — clippy is not the whole gate.

**CI on head (`96d3ebd`) is green:** `Lint`, `Test`, `Build Linux x64`, `Build WASM Assets`, `Check API Changes`, `Smoke Test Binary`, `Synology Package Contract`, `Build Summary`. `Check API Changes` passing is independent confirmation that the route contract is untouched.

Everything below was also run locally, on head:

- `cargo fmt --check` — clean (after `96d3ebd`).
- `cargo clippy -- -D warnings` — clean. (Not `--all-targets`: it cannot pass, `src/lib.rs:20-22` denies `unwrap_used`/`expect_used`/`panic` crate-wide and src's own test modules violate it 76 times, pre-existing and untouched here.)
- `cargo test --no-fail-fast` — all suites pass except the two pre-existing `/hqp/discover` failures (`client_harness::shared_endpoints::get_hqp_discover`, `protocol_integration::api_endpoints::hqp_discover_returns_json`), which do live UDP SSDP and are unrelated. They are local-only: CI's `Test` job is green, which is worth knowing before anyone else spends time on them.
- Lint suites named as hard constraints in #405: `spawn_cancellation_lint`, `unbounded_channel_lint`, `await_in_lock_lint` — pass. Also `oneshot_leak_lint`, `ignored_send_lint`, `architecture_lint`, `aggregator_lint`, `arbitrary_find_lint` — pass.
- `cargo test --test api_contract` — passes; `tests/fixtures/api_routes.txt` untouched, so no `api-change-approved` is needed.
- `dx build --release --platform web --features web` — exit 0, both `Client build completed successfully` and `Server build completed successfully`. Two traps worth recording: `make css` must run first in a fresh worktree, and `~/.cargo/config.toml` sets `[build] rustc-wrapper = sccache` — `unset RUSTC_WRAPPER` does not override a config file, `CARGO_BUILD_RUSTC_WRAPPER=""` does. With sccache active the build failed at `exit status: 254`; with it off, clean. rustup's `rustc` (1.97.1) must also precede Homebrew's (1.93.1) on `PATH`.
- Red/green: `99b3779` (tests plus today's behavior behind the new seam) → 9 of 11 new tests fail, suite **10.01s**. Head → **18 pass, suite 0.00s**.

**What each piece of evidence proves.** The 10.01s → 0.00s drop proves *promptness* and nothing else — an implementation that promptly resolved the *wrong* waiter would post the same number. Correlation is proved only by `browse_rejection_resolves_only_the_matching_waiter` and `a_mismatched_entry_in_one_map_does_not_hide_a_match_in_the_other`.

No `.await` is taken while the state lock is held: the routing call is synchronous and the guard is dropped before the logging match. That is also what makes resolution exactly-once — the caller's timeout cleanup and the routing path are serialised by the one state `RwLock`, so they can interleave but never overlap.

## One judgment call the operator may want to veto

`/roon/browse` and `/roon/play_item` now answer **404** instead of 500 when the Core refuses an item key or a zone. Routes, HTTP methods, request bodies and the `{"error": "…"}` response shape are all unchanged, and `tests/fixtures/api_routes.txt` is untouched — but a status code is still observable behavior.

The case for it: 500 says "server bug", which misdirects the caller, and the alternative is making clients match on message text — the fragility this PR exists to remove. And the old 500 only ever arrived *after* a 10s timeout, so any client tuned to this condition was already treating it as an outage; telling it "your reference is stale" is the fix rather than a break. (The stronger claim — that no client could be depending on the old status — would be false: a client can branch on 500 today and would now take a different branch.)

To revert just this: in `src/api/mod.rs`, make `roon_browse_failure` return `StatusCode::INTERNAL_SERVER_ERROR` unconditionally. The adapter fix, the promptness and the precise message all survive that change.

## The neighbouring case, checked and already fine

The obvious sequel would be "a request in flight when the Core connection drops". It is **not** broken: the `CoreEvent::Lost` arm already clears all three pending maps immediately (`src/adapters/roon.rs:1944-1946`), which drops every sender, so each waiting receiver yields `RecvError` and its caller returns `"Browse request cancelled"` / `"Image request cancelled"` promptly — no 10s wait, no strand, and already distinguishable from a timeout. The ship-gate dissent asserted otherwise and was wrong; the retraction is in the comments. Recorded here so nobody else files it.

## Rollback

revert the merge commit, or `git revert 96d3ebd c7acd64 d6b1e1b b24ae86 39c7680 8df4df3 99b3779`. The change is confined to `src/adapters/roon.rs` and `src/api/mod.rs`, adds no dependency, no migration, no config, and no persisted state. Reverting restores the pre-#405 behavior exactly: rejections are dropped and callers time out.

Do not merge without the operator's explicit approval (AGENTS.md).

## Summary by CodeRabbit

- **Bug Fixes**
  - Roon browsing and playback requests now return more accurate error responses.
  - Invalid item keys and unavailable zones are reported as “not found” instead of generic server errors.
  - Other Roon Core failures continue to return server-error responses with relevant details preserved.
  - Browse failures are handled more reliably, including request timeouts, disconnected services, and unrelated responses.

### DIFF
--- a/src/adapters/roon.rs
+++ b/src/adapters/roon.rs
@@ -134,6 +134,22 @@ type LoadRequest = oneshot::Sender<Result<LoadResult>>;
 // out the full `BROWSE_TIMEOUT` and then reported "Browse request timed out" -
 // the same thing an unreachable Core produces. The types below carry the
 // Core's actual answer back to the caller, promptly and distinguishably.
+//
+// Two things about the delivery path, both established by reading the pinned
+// fork and neither observable from here, so they are recorded rather than
+// assumed:
+//
+// - Services are tried in order with a `break` on the first that claims a
+//   message (fork `src/lib.rs:597-631`), and Transport is tried before Browse.
+//   `Transport::parse_msg` gates every push on its own subscription ids, so it
+//   returns an empty `Vec` for a browse message and declines to claim it
+//   (fork `src/transport.rs:451-540`). If that ever changes, browse errors get
+//   swallowed inside the dependency and nothing here can see it.
+// - `Browse::parse_msg` matches four literal error names against `msg["name"]`
+//   (fork `src/browse.rs:174-188`). A Core that spells one differently yields
+//   `Parsed::None`, which the fork drops, and the caller times out as it did
+//   before - with this module's tests still green. That half is wire-level and
+//   is #408's to pin.
 
 /// Which browse/load rejection Roon Core reported.
 #[derive(Debug, Clone, Copy, PartialEq, Eq)]
@@ -141,6 +157,12 @@ pub enum RoonBrowseErrorKind {
     /// The `item_key` is not valid in the browse session it was used in: it has
     /// expired, the session was reset, or it was minted somewhere else.
     /// Recoverable - re-acquire the key (search or browse again) and retry.
+    ///
+    /// The recovery instruction in [`RoonBrowseErrorKind::explain`] is written
+    /// for a caller that holds a key - `play_item` and `/roon/browse`. `search`
+    /// mints and consumes its own keys internally, so a rejection there reaches
+    /// a client that never held one; retrying the search is still the right
+    /// move, but a caller-facing surface may want to reword it (#395, #396).
     InvalidItemKey,
     /// The `level` / `pop_levels` in the request do not exist in this session.
     /// Recoverable - browse again from a level the caller still holds.
@@ -435,10 +457,14 @@ impl RoonState {
     /// long-lived browse tree is ever built on one session key (#399), stamp the
     /// pending entries with a connection generation instead.
     ///
-    /// Resolution is exactly-once: every path removes the entry before sending,
-    /// so a repeated rejection finds nothing and a waiter can never be resolved
-    /// twice. A `KeyMismatch` leaves the entry for the caller's own timeout to
-    /// clean up, so no path strands one.
+    /// Resolution is exactly-once because every path removes the entry before
+    /// sending *and* every mutation of these maps is synchronous while the one
+    /// state `RwLock` is held - so a caller's timeout cleanup and this routing
+    /// can interleave but never overlap. A repeated rejection finds nothing; a
+    /// caller that timed out first leaves nothing to find. A `KeyMismatch`
+    /// deliberately does **not** remove: that entry belongs to a caller still
+    /// inside its `BROWSE_TIMEOUT`, whose own cleanup owns it. Removing it here
+    /// would be the strand, not the fix.
     ///
     /// Both maps are searched for an entry where *both* the request id and the
     /// session key match before any mismatch is reported. A mismatched entry in
@@ -2410,7 +2436,17 @@ mod tests {
     // tests exercise the real routing functions, the real `oneshot` channels
     // and the real `BROWSE_TIMEOUT`. They do not exercise the fork's
     // `parse_msg`, nor the three-line result match in `browse()`/`load()` that
-    // hands the oneshot payload back to the caller.
+    // hands the oneshot payload back to the caller. In particular, nothing here
+    // proves that a real Core's error names match the four literals the fork
+    // matches on - if they did not, these tests would stay green and the
+    // ten-second hang would be live. That half is #408's.
+    //
+    // Which evidence supports which claim, since it is easy to conflate:
+    // `a_rejected_item_key_resolves_far_inside_the_browse_timeout` is the
+    // promptness proof; `browse_rejection_resolves_only_the_matching_waiter`
+    // and `a_mismatched_entry_in_one_map_does_not_hide_a_match_in_the_other`
+    // are the correlation proofs. A fast suite alone would not have caught an
+    // implementation that promptly resolved the *wrong* waiter.
 
     use std::time::Instant;
     use tokio::sync::oneshot::error::TryRecvError;
--- a/src/adapters/roon.rs
+++ b/src/adapters/roon.rs
@@ -120,6 +120,159 @@ type BrowseRequest = oneshot::Sender<Result<BrowseResult>>;
 /// Pending load request - stores the oneshot sender to deliver the result
 type LoadRequest = oneshot::Sender<Result<LoadResult>>;
 
+// =============================================================================
+// Core-reported browse rejections (issue #405)
+// =============================================================================
+//
+// Roon Core answers an unusable browse or load request with a named error
+// instead of a result. The pinned fork surfaces that as
+// `Parsed::Error(RoonApiError::Browse*)`, carrying both the originating
+// `req_id` and the request's `multi_session_key` (see the fork's
+// `src/browse.rs:174-188`).
+//
+// Before #405 the event loop dropped those messages, so the waiting caller sat
+// out the full `BROWSE_TIMEOUT` and then reported "Browse request timed out" -
+// the same thing an unreachable Core produces. The types below carry the
+// Core's actual answer back to the caller, promptly and distinguishably.
+
+/// Which browse/load rejection Roon Core reported.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum RoonBrowseErrorKind {
+    /// The `item_key` is not valid in the browse session it was used in: it has
+    /// expired, the session was reset, or it was minted somewhere else.
+    /// Recoverable - re-acquire the key (search or browse again) and retry.
+    InvalidItemKey,
+    /// The `level` / `pop_levels` in the request do not exist in this session.
+    /// Recoverable - browse again from a level the caller still holds.
+    InvalidLevels,
+    /// The `zone_or_output_id` in the request is unknown to the Core. Not
+    /// fixed by retrying: the caller has to name a zone that exists.
+    ZoneNotFound,
+    /// The Core reported an unexpected failure for this request.
+    UnexpectedError,
+}
+
+impl RoonBrowseErrorKind {
+    /// Stable machine-readable tag for logs and callers.
+    pub fn as_str(self) -> &'static str {
+        match self {
+            Self::InvalidItemKey => "invalid_item_key",
+            Self::InvalidLevels => "invalid_levels",
+            Self::ZoneNotFound => "zone_not_found",
+            Self::UnexpectedError => "unexpected_error",
+        }
+    }
+
+    /// True when the caller can recover by re-acquiring its reference and
+    /// retrying immediately, rather than by waiting for the Core.
+    pub fn is_recoverable(self) -> bool {
+        matches!(self, Self::InvalidItemKey | Self::InvalidLevels)
+    }
+
+    /// Human-readable explanation, including the recovery instruction where
+    /// there is one. Deliberately never contains the word "timeout" - callers
+    /// and operators must be able to tell these apart at a glance.
+    fn explain(self) -> &'static str {
+        match self {
+            Self::InvalidItemKey => {
+                "Roon Core rejected the item key: it is no longer valid in this \
+                 browse session - search or browse again to get a fresh key"
+            }
+            Self::InvalidLevels => {
+                "Roon Core rejected the browse level: this browse session is not \
+                 at that level - browse again from a level you still hold"
+            }
+            Self::ZoneNotFound => {
+                "Roon Core does not recognise the zone or output in this browse \
+                 request"
+            }
+            Self::UnexpectedError => {
+                "Roon Core reported an unexpected error for this browse request"
+            }
+        }
+    }
+}
+
+/// A browse or load request that Roon Core answered with a rejection.
+///
+/// Distinct from a timeout: the Core answered, and it answered "no". Returned
+/// inside the `anyhow::Error` that `browse`, `load`, `search`, `play_item` and
+/// `search_and_play` already return, so no signature changes; recover it with
+/// [`RoonBrowseError::from_error`].
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct RoonBrowseError {
+    /// What the Core rejected.
+    pub kind: RoonBrowseErrorKind,
+    /// The `multi_session_key` the rejected request ran in, as reported by the
+    /// Core. `None` only if the request carried no session key.
+    pub session_key: Option<String>,
+}
+
+impl RoonBrowseError {
+    fn new(kind: RoonBrowseErrorKind, session_key: Option<&str>) -> Self {
+        Self {
+            kind,
+            session_key: session_key.map(str::to_string),
+        }
+    }
+
+    /// Recover the typed rejection from an error returned by the browse-backed
+    /// adapter methods. `None` means the failure was something else - a
+    /// timeout, a lost Core, or Browse not being connected - so callers can
+    /// keep those classes apart without matching on message text.
+    pub fn from_error(err: &anyhow::Error) -> Option<&Self> {
+        err.downcast_ref::<Self>()
+    }
+}
+
+impl std::error::Error for RoonBrowseError {}
+
+impl std::fmt::Display for RoonBrowseError {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.kind.explain())?;
+        if let Some(session_key) = &self.session_key {
+            write!(f, " (browse session '{}')", session_key)?;
+        }
+        Ok(())
+    }
+}
+
+/// Where a Core rejection ended up. Reported for logging and asserted by the
+/// correlation tests; the caller-visible effect is the resolved oneshot.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum ErrorRouting {
+    /// Delivered to the pending browse holding this `req_id`.
+    Browse,
+    /// Delivered to the pending load holding this `req_id`.
+    Load,
+    /// Delivered to the pending image request holding this `req_id`.
+    Image,
+    /// The waiter was found and removed, but its receiver was already gone.
+    ReceiverGone,
+    /// No request with this `req_id` is pending - it was already resolved, or
+    /// the caller timed out and cleaned up. Dropping the rejection is correct.
+    NoWaiter,
+    /// A request with this `req_id` is pending but carries a different session
+    /// or image key, so it was left alone to hit its own timeout rather than
+    /// being resolved with someone else's rejection.
+    KeyMismatch,
+}
+
+/// Deliver a Core rejection to a waiting browse or load request.
+/// Returns whether the waiter was still listening.
+fn deliver_browse_rejection<T>(
+    sender: oneshot::Sender<Result<T>>,
+    kind: RoonBrowseErrorKind,
+    session_key: Option<&str>,
+) -> bool {
+    sender
+        .send(Err(anyhow::Error::new(RoonBrowseError::new(
+            kind,
+            session_key,
+        ))))
+        .is_ok()
+}
+
 /// Image data returned from Roon
 #[derive(Debug, Clone)]
 pub struct ImageData {
@@ -244,6 +397,29 @@ struct RoonState {
     pending_browses: HashMap<usize, (Option<String>, BrowseRequest)>,
     /// Pending load requests: request_id -> (session_key, oneshot sender)
     pending_loads: HashMap<usize, (Option<String>, LoadRequest)>,
+}
+
+impl RoonState {
+    /// Route a Core browse/load rejection to the exact request waiting on it.
+    ///
+    /// TODO(#405): not implemented yet - this reproduces the pre-#405 behavior
+    /// (the rejection is dropped and no waiter is resolved) so the correlation
+    /// tests fail against it.
+    fn route_browse_rejection(
+        &mut self,
+        _req_id: usize,
+        _session_key: Option<&str>,
+        _kind: RoonBrowseErrorKind,
+    ) -> ErrorRouting {
+        ErrorRouting::NoWaiter
+    }
+
+    /// Route a Core image rejection to the request waiting on it.
+    ///
+    /// TODO(#405): not implemented yet - see `route_browse_rejection`.
+    fn route_image_rejection(&mut self, _req_id: usize, _image_key: &str) -> ErrorRouting {
+        ErrorRouting::NoWaiter
+    }
 }
 
 /// Roon adapter
@@ -2050,6 +2226,361 @@ mod tests {
         assert!(
             bus_zone.volume_control.is_none(),
             "should be None when output has no volume"
+        );
+    }
+
+    // =========================================================================
+    // Core-reported browse rejections (issue #405)
+    // =========================================================================
+    //
+    // Test strategy: these unit-scope the correlation logic. They do NOT drive
+    // a Roon protocol mock - `tests/mock_servers/roon.rs` holds state and
+    // speaks no protocol, and building a protocol-level Roon fake is #408's
+    // job. Duplicating it here would conflict with that work. What is scoped
+    // here is exactly the concurrency-sensitive part: which waiter a rejection
+    // resolves when several requests are in flight, and what happens to the
+    // ones it must not touch.
+    //
+    // Coverage boundary, stated so the next agent does not over-trust it: the
+    // tests exercise the real routing functions, the real `oneshot` channels
+    // and the real `BROWSE_TIMEOUT`. They do not exercise the fork's
+    // `parse_msg`, nor the three-line result match in `browse()`/`load()` that
+    // hands the oneshot payload back to the caller.
+
+    use std::time::Instant;
+    use tokio::sync::oneshot::error::TryRecvError;
+
+    /// Queue a pending browse the way `browse()` does, returning its receiver.
+    fn pending_browse(
+        state: &mut RoonState,
+        req_id: usize,
+        session_key: &str,
+    ) -> oneshot::Receiver<Result<BrowseResult>> {
+        let (tx, rx) = oneshot::channel();
+        state
+            .pending_browses
+            .insert(req_id, (Some(session_key.to_string()), tx));
+        rx
+    }
+
+    /// Queue a pending load the way `load()` does, returning its receiver.
+    fn pending_load(
+        state: &mut RoonState,
+        req_id: usize,
+        session_key: &str,
+    ) -> oneshot::Receiver<Result<LoadResult>> {
+        let (tx, rx) = oneshot::channel();
+        state
+            .pending_loads
+            .insert(req_id, (Some(session_key.to_string()), tx));
+        rx
+    }
+
+    /// Queue a pending image request the way `get_image()` does.
+    fn pending_image(
+        state: &mut RoonState,
+        req_id: usize,
+        image_key: &str,
+    ) -> oneshot::Receiver<Option<ImageData>> {
+        let (tx, rx) = oneshot::channel();
+        state
+            .pending_images
+            .insert(req_id, (image_key.to_string(), tx));
+        rx
+    }
+
+    /// The rejection a receiver was resolved with, or a description of why it
+    /// was not resolved.
+    fn rejection_kind<T>(rx: &mut oneshot::Receiver<Result<T>>) -> RoonBrowseErrorKind {
+        match rx.try_recv() {
+            Ok(Err(err)) => match RoonBrowseError::from_error(&err) {
+                Some(rejection) => rejection.kind,
+                None => unreachable!("resolved with an untyped error: {err}"),
+            },
+            Ok(Ok(_)) => unreachable!("resolved with a success result"),
+            Err(TryRecvError::Empty) => unreachable!("waiter was never resolved"),
+            Err(TryRecvError::Closed) => unreachable!("waiter's sender was dropped"),
+        }
+    }
+
+    fn assert_still_waiting<T>(rx: &mut oneshot::Receiver<Result<T>>, what: &str) {
+        assert!(
+            matches!(rx.try_recv(), Err(TryRecvError::Empty)),
+            "{what} must still be waiting - it was resolved or its sender dropped"
+        );
+    }
+
+    #[test]
+    fn browse_rejection_resolves_only_the_matching_waiter() {
+        // Several browse and load requests in flight, as the adapter routinely
+        // has during search() and play_item(). One is rejected by the Core.
+        let mut state = RoonState::default();
+        let mut first = pending_browse(&mut state, 11, "search_a");
+        let mut rejected = pending_browse(&mut state, 12, "search_b");
+        let mut third = pending_browse(&mut state, 13, "search_c");
+        let mut load_a = pending_load(&mut state, 14, "search_a");
+        let mut load_b = pending_load(&mut state, 15, "search_b");
+
+        let routing = state.route_browse_rejection(
+            12,
+            Some("search_b"),
+            RoonBrowseErrorKind::InvalidItemKey,
+        );
+
+        assert_eq!(routing, ErrorRouting::Browse);
+        assert_eq!(
+            rejection_kind(&mut rejected),
+            RoonBrowseErrorKind::InvalidItemKey
+        );
+        assert_still_waiting(&mut first, "browse 11");
+        assert_still_waiting(&mut third, "browse 13");
+        assert_still_waiting(&mut load_a, "load 14");
+        // load 15 shares the rejected request's session key: proof that
+        // correlation is by req_id, not by scanning for a session key.
+        assert_still_waiting(&mut load_b, "load 15");
+
+        // Only the resolved entry leaves the map; nothing is stranded.
+        assert!(!state.pending_browses.contains_key(&12));
+        assert_eq!(state.pending_browses.len(), 2);
+        assert_eq!(state.pending_loads.len(), 2);
+    }
+
+    #[test]
+    fn load_rejection_resolves_the_load_not_a_browse_sharing_its_session() {
+        // A browse and a load in the same browse session, both pending. A
+        // session-key scan would resolve whichever the map yielded first; the
+        // req_id in the rejection says exactly which one the Core refused.
+        let mut state = RoonState::default();
+        let mut browse = pending_browse(&mut state, 20, "play_item_1");
+        let mut load = pending_load(&mut state, 21, "play_item_1");
+
+        let routing = state.route_browse_rejection(
+            21,
+            Some("play_item_1"),
+            RoonBrowseErrorKind::InvalidItemKey,
+        );
+
+        assert_eq!(routing, ErrorRouting::Load);
+        assert_eq!(
+            rejection_kind(&mut load),
+            RoonBrowseErrorKind::InvalidItemKey
+        );
+        assert_still_waiting(&mut browse, "browse 20");
+        assert!(state.pending_browses.contains_key(&20));
+        assert!(!state.pending_loads.contains_key(&21));
+    }
+
+    #[test]
+    fn rejection_with_a_mismatched_session_key_leaves_the_waiter_alone() {
+        // `Moo` restarts its request-id counter at 0 on every reconnect, so a
+        // rejection on a new connection can carry a req_id that a stale, not
+        // yet timed-out waiter still occupies. Resolving it would be a
+        // mis-correlation; the waiter is left to its existing timeout instead.
+        let mut state = RoonState::default();
+        let mut stale = pending_browse(&mut state, 7, "browse_old");
+
+        let routing = state.route_browse_rejection(
+            7,
+            Some("browse_new"),
+            RoonBrowseErrorKind::InvalidItemKey,
+        );
+
+        assert_eq!(routing, ErrorRouting::KeyMismatch);
+        assert_still_waiting(&mut stale, "browse 7");
+        assert!(
+            state.pending_browses.contains_key(&7),
+            "the waiter must stay in the map so its own timeout still cleans up"
+        );
+    }
+
+    #[test]
+    fn rejection_for_an_unknown_req_id_is_a_noop() {
+        let mut state = RoonState::default();
+        let mut other = pending_browse(&mut state, 30, "browse_x");
+
+        let routing = state.route_browse_rejection(
+            999,
+            Some("browse_x"),
+            RoonBrowseErrorKind::UnexpectedError,
+        );
+
+        assert_eq!(routing, ErrorRouting::NoWaiter);
+        assert_still_waiting(&mut other, "browse 30");
+        assert_eq!(state.pending_browses.len(), 1);
+    }
+
+    #[test]
+    fn a_second_rejection_for_the_same_req_id_cannot_double_resolve() {
+        let mut state = RoonState::default();
+        let mut rx = pending_browse(&mut state, 40, "browse_y");
+
+        let first = state.route_browse_rejection(
+            40,
+            Some("browse_y"),
+            RoonBrowseErrorKind::InvalidItemKey,
+        );
+        let second = state.route_browse_rejection(
+            40,
+            Some("browse_y"),
+            RoonBrowseErrorKind::InvalidItemKey,
+        );
+
+        assert_eq!(first, ErrorRouting::Browse);
+        assert_eq!(second, ErrorRouting::NoWaiter);
+        assert_eq!(rejection_kind(&mut rx), RoonBrowseErrorKind::InvalidItemKey);
+        assert!(state.pending_browses.is_empty());
+    }
+
+    #[test]
+    fn rejection_after_the_caller_timed_out_is_a_noop() {
+        // `browse()` removes its own entry when BROWSE_TIMEOUT fires. A
+        // rejection arriving afterwards must not panic or resurrect anything.
+        let mut state = RoonState::default();
+        let _rx = pending_browse(&mut state, 50, "browse_z");
+        state.pending_browses.remove(&50);
+
+        let routing =
+            state.route_browse_rejection(50, Some("browse_z"), RoonBrowseErrorKind::InvalidLevels);
+
+        assert_eq!(routing, ErrorRouting::NoWaiter);
+        assert!(state.pending_browses.is_empty());
+    }
+
+    #[test]
+    fn rejection_with_a_dropped_receiver_still_clears_the_entry() {
+        let mut state = RoonState::default();
+        let rx = pending_browse(&mut state, 60, "browse_w");
+        drop(rx);
+
+        let routing = state.route_browse_rejection(
+            60,
+            Some("browse_w"),
+            RoonBrowseErrorKind::InvalidItemKey,
+        );
+
+        assert_eq!(routing, ErrorRouting::ReceiverGone);
+        assert!(
+            state.pending_browses.is_empty(),
+            "a gone receiver must not leave a stranded map entry"
+        );
+    }
+
+    #[test]
+    fn every_browse_rejection_kind_is_routed() {
+        // Routing only InvalidItemKey would leave the other three Core
+        // rejections hanging their callers for the full timeout.
+        for kind in [
+            RoonBrowseErrorKind::InvalidItemKey,
+            RoonBrowseErrorKind::InvalidLevels,
+            RoonBrowseErrorKind::ZoneNotFound,
+            RoonBrowseErrorKind::UnexpectedError,
+        ] {
+            let mut state = RoonState::default();
+            let mut rx = pending_browse(&mut state, 70, "browse_all");
+
+            let routing = state.route_browse_rejection(70, Some("browse_all"), kind);
+
+            assert_eq!(routing, ErrorRouting::Browse, "kind {kind:?} was not routed");
+            assert_eq!(rejection_kind(&mut rx), kind);
+        }
+    }
+
+    #[tokio::test]
+    async fn a_rejected_item_key_resolves_far_inside_the_browse_timeout() {
+        // The point of the issue: promptness. Awaited exactly as `browse()`
+        // awaits it, with the real BROWSE_TIMEOUT.
+        let mut state = RoonState::default();
+        let rx = pending_browse(&mut state, 80, "browse_prompt");
+
+        let started = Instant::now();
+        let routing = state.route_browse_rejection(
+            80,
+            Some("browse_prompt"),
+            RoonBrowseErrorKind::InvalidItemKey,
+        );
+        let result = tokio::time::timeout(BROWSE_TIMEOUT, rx).await;
+        let elapsed = started.elapsed();
+
+        assert_eq!(routing, ErrorRouting::Browse);
+        assert!(
+            elapsed < Duration::from_secs(1),
+            "a rejected item key must not wait out BROWSE_TIMEOUT ({BROWSE_TIMEOUT:?}); \
+             took {elapsed:?}"
+        );
+        let Ok(Ok(Err(err))) = result else {
+            unreachable!("expected a typed rejection inside the timeout");
+        };
+        let Some(rejection) = RoonBrowseError::from_error(&err) else {
+            unreachable!("expected a RoonBrowseError, got: {err}");
+        };
+        assert_eq!(rejection.kind, RoonBrowseErrorKind::InvalidItemKey);
+        assert_eq!(rejection.session_key.as_deref(), Some("browse_prompt"));
+    }
+
+    #[test]
+    fn image_rejection_resolves_the_matching_image_request() {
+        let mut state = RoonState::default();
+        let mut wanted = pending_image(&mut state, 90, "image_a");
+        let mut other = pending_image(&mut state, 91, "image_b");
+
+        let routing = state.route_image_rejection(90, "image_a");
+
+        assert_eq!(routing, ErrorRouting::Image);
+        // `get_image` reads None as "Image not found".
+        assert!(matches!(wanted.try_recv(), Ok(None)));
+        assert!(matches!(other.try_recv(), Err(TryRecvError::Empty)));
+        assert!(!state.pending_images.contains_key(&90));
+        assert!(state.pending_images.contains_key(&91));
+    }
+
+    #[test]
+    fn image_rejection_with_a_mismatched_key_leaves_the_waiter_alone() {
+        let mut state = RoonState::default();
+        let mut rx = pending_image(&mut state, 92, "image_a");
+
+        let routing = state.route_image_rejection(92, "image_somewhere_else");
+
+        assert_eq!(routing, ErrorRouting::KeyMismatch);
+        assert!(matches!(rx.try_recv(), Err(TryRecvError::Empty)));
+        assert!(state.pending_images.contains_key(&92));
+    }
+
+    #[test]
+    fn a_rejection_is_distinguishable_from_a_timeout_and_a_lost_core() {
+        // The three failure classes the issue insists must stay apart.
+        let rejected = anyhow::Error::new(RoonBrowseError::new(
+            RoonBrowseErrorKind::InvalidItemKey,
+            Some("browse_1"),
+        ));
+        let timed_out = anyhow::anyhow!("Browse request timed out");
+        let not_connected =
+            anyhow::anyhow!("Browse service not available - not connected to Roon");
+
+        let Some(rejection) = RoonBrowseError::from_error(&rejected) else {
+            unreachable!("a rejection must be recoverable from the anyhow error");
+        };
+        assert_eq!(rejection.kind, RoonBrowseErrorKind::InvalidItemKey);
+        assert!(rejection.kind.is_recoverable());
+        assert!(RoonBrowseError::from_error(&timed_out).is_none());
+        assert!(RoonBrowseError::from_error(&not_connected).is_none());
+
+        // And the message a caller sees says what to do, without ever reading
+        // like a timeout.
+        let message = rejected.to_string();
+        assert!(message.contains("search or browse again"), "{message}");
+        assert!(!message.contains("timed out"), "{message}");
+        assert!(message.contains("browse_1"), "{message}");
+    }
+
+    #[test]
+    fn only_reference_rejections_are_marked_recoverable() {
+        assert!(RoonBrowseErrorKind::InvalidItemKey.is_recoverable());
+        assert!(RoonBrowseErrorKind::InvalidLevels.is_recoverable());
+        assert!(!RoonBrowseErrorKind::ZoneNotFound.is_recoverable());
+        assert!(!RoonBrowseErrorKind::UnexpectedError.is_recoverable());
+        assert_eq!(
+            RoonBrowseErrorKind::InvalidItemKey.as_str(),
+            "invalid_item_key"
         );
     }
 }
--- a/src/adapters/roon.rs
+++ b/src/adapters/roon.rs
@@ -428,16 +428,23 @@ impl RoonState {
     /// so a repeated rejection finds nothing and a waiter can never be resolved
     /// twice. A `KeyMismatch` leaves the entry for the caller's own timeout to
     /// clean up, so no path strands one.
+    ///
+    /// Both maps are searched for an entry where *both* the request id and the
+    /// session key match before any mismatch is reported. A mismatched entry in
+    /// one map must not hide a matching entry in the other - that too is only
+    /// reachable through a reconnect id collision, but reporting a mismatch
+    /// early would cost a legitimate waiter its prompt answer.
     fn route_browse_rejection(
         &mut self,
         req_id: usize,
         session_key: Option<&str>,
         kind: RoonBrowseErrorKind,
     ) -> ErrorRouting {
-        if let Some((pending_key, _)) = self.pending_browses.get(&req_id) {
-            if pending_key.as_deref() != session_key {
-                return ErrorRouting::KeyMismatch;
-            }
+        let browse_matches = matches!(
+            self.pending_browses.get(&req_id),
+            Some((pending_key, _)) if pending_key.as_deref() == session_key
+        );
+        if browse_matches {
             if let Some((_, sender)) = self.pending_browses.remove(&req_id) {
                 return if deliver_browse_rejection(sender, kind, session_key) {
                     ErrorRouting::Browse
@@ -447,10 +454,11 @@ impl RoonState {
             }
         }
 
-        if let Some((pending_key, _)) = self.pending_loads.get(&req_id) {
-            if pending_key.as_deref() != session_key {
-                return ErrorRouting::KeyMismatch;
-            }
+        let load_matches = matches!(
+            self.pending_loads.get(&req_id),
+            Some((pending_key, _)) if pending_key.as_deref() == session_key
+        );
+        if load_matches {
             if let Some((_, sender)) = self.pending_loads.remove(&req_id) {
                 return if deliver_browse_rejection(sender, kind, session_key) {
                     ErrorRouting::Load
@@ -458,6 +466,10 @@ impl RoonState {
                     ErrorRouting::ReceiverGone
                 };
             }
+        }
+
+        if self.pending_browses.contains_key(&req_id) || self.pending_loads.contains_key(&req_id) {
+            return ErrorRouting::KeyMismatch;
         }
 
         ErrorRouting::NoWaiter
@@ -2515,6 +2527,29 @@ mod tests {
             state.pending_browses.contains_key(&7),
             "the waiter must stay in the map so its own timeout still cleans up"
         );
+    }
+
+    #[test]
+    fn a_mismatched_entry_in_one_map_does_not_hide_a_match_in_the_other() {
+        // Only reachable through a reconnect id collision: a stale browse from
+        // the old connection holds req_id 8, and the new connection issues a
+        // load that also gets req_id 8. Reporting the browse's key mismatch
+        // early would cost the load its prompt answer for no safety gain.
+        let mut state = RoonState::default();
+        let mut stale_browse = pending_browse(&mut state, 8, "browse_old");
+        let mut live_load = pending_load(&mut state, 8, "load_new");
+
+        let routing =
+            state.route_browse_rejection(8, Some("load_new"), RoonBrowseErrorKind::InvalidItemKey);
+
+        assert_eq!(routing, ErrorRouting::Load);
+        assert_eq!(
+            rejection_kind(&mut live_load),
+            RoonBrowseErrorKind::InvalidItemKey
+        );
+        assert_still_waiting(&mut stale_browse, "the stale browse");
+        assert!(state.pending_browses.contains_key(&8));
+        assert!(!state.pending_loads.contains_key(&8));
     }
 
     #[test]
--- a/src/adapters/roon.rs
+++ b/src/adapters/roon.rs
@@ -422,7 +422,18 @@ impl RoonState {
     ///   connection can carry a request id that an older, not-yet-timed-out
     ///   entry still occupies. On a mismatch the waiter is left alone and falls
     ///   back to `BROWSE_TIMEOUT` - the pre-#405 behavior - because resolving
-    ///   the wrong caller is worse than resolving them late.
+    ///   the wrong caller is worse than resolving them late. Do not remove this
+    ///   check as redundant: `req_id` uniqueness is per-connection only, and
+    ///   nothing near this code says so.
+    ///
+    /// The pair is not a unique key, and this narrows the reconnect window
+    /// rather than closing it. Session keys are not unique either - a caller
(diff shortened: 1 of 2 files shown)


muness:
## Gate report — SOLUTION SPACE / review

**Gate/role:** Solution-space gate, reviewer

### Aim and scope checked

Aim: a browse or load that Roon Core refuses must resolve its caller promptly, with a failure the caller can tell apart from an unreachable Core — without ever resolving the wrong caller. Scope checked: the three axes #405 leaves soft — how the rejection is correlated to its waiter, what error type reaches callers, and which `RoonApiError` variants are routed in this pass. Test strategy was pre-decided (unit-scope; #408 owns the protocol fake) and is treated here as a hard constraint, not a choice.

### The terrain, read first

The mismatch #405 names is real but not symmetric, and that asymmetry is the whole design:

| Message | Carries `req_id`? | Carries session key? |
|---|---|---|
| `Parsed::BrowseResult` | **no** | yes |
| `Parsed::LoadResult` | **no** | yes |
| `Parsed::Error(Browse*)` | **yes** | yes |
| `Parsed::Error(Image*)` | **yes** | image key |

The success arms scan by session key because they have nothing else. The error arms are handed the exact `req_id`. So the two halves should *not* correlate the same way, and the instinct to make them symmetric is the trap.

Three fork facts establish that `req_id` is authoritative, all verified in the pinned checkout `06dd807`:

- `Browse::browse`/`Browse::load` return the id UHC stores as its map key (`src/browse.rs:126-154`).
- `Moo::send_req` allocates ids from one monotonic counter per connection (`src/moo.rs:124-135`).
- `parse_msg` reads `request_id` back off the wire and hands it to the error variant (`src/browse.rs:157, 174-188`).

And one fact establishes the guard it needs: `Moo::new` creates that counter at 0 on every connection (`src/moo.rs:92`), so ids restart after a reconnect.

Two further checks that shaped the design:

- `parse_msg` **removes** the req_id from its own `session_keys` map on the first message (`:164`), so at most one `Parsed` is ever produced per request. An error and a result cannot both arrive for one `req_id` — which removes an entire class of ordering hazard.
- The error body cannot be mistaken for a result. `parse_msg` tries `BrowseResult` then `LoadResult` before reading `msg["name"]`, and both require non-`Option` fields (`action`; `items`/`offset`/`list`) that an error body does not have. The fall-through is sound.

### Options compared — correlation

**Option A — scan by session key, mirroring the success arms.**
Rejected. A session key is not a request identifier: `search()` reuses one across six requests (`src/adapters/roon.rs:682-772`), and `/roon/browse` lets a caller supply its own (`src/api/mod.rs:627`). A scan cannot say which request in a session was refused, and — worse — cannot say whether it was a browse or a load, because it has to pick a map to scan first. With a browse and a load pending in one session, scanning `pending_browses` first resolves the browse when the Core refused the load. That is #405's "resolving the wrong waiter" failure, reached by the most natural-looking implementation.

**Option B — `req_id` exact lookup, session key as a cross-check (selected).**
`remove(&req_id)` from whichever map holds it, but only after the stored session key matches the one the Core reported. Exact, O(1), no map-order dependency. The cross-check exists solely for the reconnect id-reuse window; on mismatch the waiter is left to `BROWSE_TIMEOUT`, i.e. exactly today's behavior, because late is better than wrong.

**Option C — `req_id` alone, no cross-check.**
Rejected. Simpler and correct on a stable connection, but a rejection arriving on a fresh connection can carry an id that a stale, not-yet-timed-out entry still occupies, and it would resolve that stale caller with another request's error. The window is narrow (a reconnect while a request is inside its 10s), but the whole point of #405 is that a caller can trust what it is told.

**Option D — change the fork so success carries `req_id` too, then key everything by it.**
Rejected for this issue, though it is the better end state. It would also close the latent ambiguity in the success arms. But it modifies a pinned dependency, puts the fix behind a fork release, and #408 explicitly warns against growing a second implementation of Roon semantics. Worth its own issue; noted under residual risk.

### Options compared — error type reaching callers

**Option A — typed error inside `anyhow`, recovered by downcast (selected).** `RoonBrowseError` implements `std::error::Error`, is boxed into the `anyhow::Error` these methods already return, and survives `?` through `search()`, `play_item()` and `search_and_play()` untouched. Zero signature changes; callers that care classify with `RoonBrowseError::from_error`, callers that do not see a better message.

**Option B — change return types to a concrete error enum.** Rejected: ripples through every browse-backed method and both HTTP handlers, and mixes with `anyhow` at each boundary. Not behavior-additive in the sense #405 requires.

**Option C — leave it a string and match on text.** Rejected outright. That is the fragility the issue exists to remove; a message edit would silently un-distinguish the classes.

### Options compared — which variants to route

Selected: all four browse variants plus both image variants.

Routing only `InvalidItemKey` would satisfy the issue's title and leave three identical 10s hangs behind the same catch-all — the defect, minus one instance. The image pair is the same shape (`pending_images` is keyed by `req_id` with the image key stored) and the same defect, and needs no type change because `get_image` already reads `None` as "Image not found" (`src/adapters/roon.rs:563`).

`RoonApiError` is matched exhaustively with no wildcard, so a fork that gains a variant breaks the build rather than quietly resuming the swallow.

### Evidence inspected

- `src/adapters/roon.rs` — event loop and both success arms (`:1865-1911` pre-change), `browse()`/`load()` (`:580-672`), `search()` (`:675-776`), `play_item()` (`:1094-1201`), `get_image()` (`:511-567`), `RoonState` (`:232-247`)
- Fork `06dd807`: `src/browse.rs:100-190`, `src/lib.rs:919-991`, `src/moo.rs:67-135`, `src/image.rs:131`, `src/sood.rs:120-188`
- `grep -rn "Parsed::" src/adapters/roon.rs` → one match site, the event loop. Nothing else in the repo swallows `Parsed::Error`.
- `grep -rn "roon/browse\|play_item" src/` → `/roon/play_item` has zero in-repo callers; `/roon/browse` has none either (the Dioxus UI has no browse page). Both are external-client surface only.
- `tests/mock_servers/roon.rs` in full — 281 lines of state holder, no transport, no protocol. Confirms #405's and #408's reading.
- `tests/fixtures/api_routes.txt`, `tests/api_contract.rs` — the contract pins route paths and methods only; no status codes, no payload shapes.
- `Cargo.toml:10-35` — `default = ["server"]`, so `cargo clippy` covers `src/api` and `src/adapters` without extra flags.

### Findings

**P2 — the natural implementation resolves the wrong waiter, and a plausible test suite would not notice.** A session-key scan (Option A) with one browse and one load pending in the same session resolves whichever map is scanned first. A test that only ever has *one* request in flight passes against it. Required: a test with a browse and a load sharing a session key where the load is the one refused.

**P2 — `req_id` is not globally unique, only per-connection, and the code that must know this is nowhere near `Moo`.** `Moo::new` resets the counter (`src/moo.rs:92`). Nothing in `src/adapters/roon.rs` hints at it. Required: the cross-check, plus the reason recorded at the routing function itself — a future reader who deletes the check as redundant re-introduces a mis-resolve that no ordinary test will catch.

**P3 — the two image rejection variants collapse into one caller-visible outcome.** `pending_images` senders carry `Option<ImageData>`, so both `ImageNotFound` and `ImageUnexpectedError` can only deliver `None`, which `get_image` reports as "Image not found". Distinguishing them means changing `ImageRequest` to carry a `Result` — a wider change than this issue's scope justifies for cover art. Required: keep `None`, log the distinction (the `RoonApiError` Display already names it), and record the lossiness rather than leave it to be discovered.

**P3 — resolution must be exactly-once by construction, not by convention.** Two removers exist for each map: the caller's timeout path (`:616-619`, `:662-665`) and now the error path. Required: every delivery path must `remove` before it sends, so a repeated rejection finds an empty slot instead of relying on a flag. Tests must cover repeat-rejection, post-timeout rejection, and a dropped receiver.

**P3 — the HTTP status code is the one place this can overreach.** `/roon/browse` and `/roon/play_item` return 500 for every adapter failure today. Answering 404 for a refused reference is more honest and is what makes the class machine-distinguishable at the HTTP edge — but a status code is observable behavior, and AGENTS.md guards API changes. Required: keep routes, methods, request bodies and the `{"error": …}` shape byte-identical; scope the new status strictly to `InvalidItemKey`/`ZoneNotFound`; flag it in the PR body as operator-vetoable with the exact one-line revert. Do not touch `tests/fixtures/api_routes.txt`.

**P4 — the empirical portability question is not answerable from this machine, and that must be established rather than assumed.** Required: probe for a Core, record what was found, and write the check as a runnable script for whoever has the rig — not a prose description.

### Required actions

1. Correlate by `req_id`, cross-checked against the session key. Record the three fork facts and the reconnect reset at the routing function.
2. Test with several requests in flight, including a browse and a load sharing a session key, where the load is the one refused.
3. Every delivery path removes before sending. Test repeat, post-timeout, and dropped-receiver.
4. Route all four browse variants; route the image pair and call it out separately in the PR; match `RoonApiError` exhaustively.
5. Keep the routing call synchronous under the state lock and drop the guard before logging — `await_in_lock_lint` must stay green.
6. Status code change confined to the two reference-rejection kinds, flagged as vetoable with its revert.
7. State the unit-scope test strategy and its coverage boundary in both the PR and the test module.
8. Probe for a live Core; report what is needed rather than declaring anything verified.

### Disposition

Option B (`req_id` + cross-check), Option A for the error type, all six variants routed. Actions 1–8 carried into the execute gate.

### Residual risk

Option D is still the right end state, and not doing it leaves the *success* arms correlating by a key that does not identify a request. This PR does not make that worse and cannot fix it without touching the pinned fork — but a reader who sees the error path correlate exactly may reasonably assume the success path does too. That has to be said in the code, not only in this report.

### Next gate authorization

Authorized to proceed to execute gate.

muness:
## Gate report — SOLUTION SPACE / dissent

**Gate/role:** Solution-space gate, dissent

Attacking the concurrency reasoning specifically, as #405 asks. The selected design survives, but two of the review's claims are overstated, one guard has a liveness cost the review did not price, and the review's own P2 finding is weaker than it thinks.

### Contrary case

**P2 — "a `req_id` cannot be in both maps" is false as stated, and the code the review authorised short-circuits on it.**

The review says: *"a `req_id` names exactly one submitted request, and `pending_browses` / `pending_loads` cannot both hold it."* True per connection. But the review then establishes, in its own next paragraph, that ids restart at 0 on reconnect — which is exactly the condition under which both maps *can* hold the same id. The review asserts uniqueness and then disproves it, and does not notice that the implementation shape it authorised reads the maps in order.

The concrete sequence:

1. Connection A: `browse()` gets `req_id` 8, session `browse_old`. Inserted into `pending_browses`. Caller is inside its 10s.
2. Connection A drops. Nothing clears the pending maps until the run function returns — and `CoreEvent::Lost` handling does not clear them.
3. Connection B: `load()` gets `req_id` 8, session `load_new`. Inserted into `pending_loads`.
4. Core refuses the load. Rejection arrives: `(8, Some("load_new"))`.

A "check `pending_browses`, return `KeyMismatch` on mismatch" implementation stops at step 4 with a mismatch and never looks in `pending_loads`. The refused load then waits out its full timeout — the exact symptom #405 exists to remove, reintroduced by the guard meant to protect it. Safe, but not live, and it is reached by the same reconnect the guard was written for. If the guard's justifying scenario also defeats the fix, the guard is mis-specified.

**Required (P2):** search *both* maps for a full `(req_id, session key)` match before reporting any mismatch. `KeyMismatch` must mean "no matching waiter anywhere, but some entry holds this id", not "the first map I looked in disagreed".

*Disposition: accepted and implemented in `b24ae86`, with `a_mismatched_entry_in_one_map_does_not_hide_a_match_in_the_other` pinning it. That commit exists because of this finding.*

**P3 — the session-key cross-check is weaker than the review implies, and the PR should not claim it closes the reconnect window.**

Session keys are not unique either. `/roon/browse` accepts a caller-supplied `session_key` (`src/api/mod.rs:575`), so a client that reuses a stable key across a reconnect can produce a stale entry and a fresh request with *the same* key and *the same* id. Both checks pass and the stale waiter is resolved with the new request's rejection; the new request then times out.

This is narrower than the unguarded case and strictly better than today (where both callers time out), and no caller is ever told a wrong *success*. But the honest claim is "materially narrows the window", not "closes it". The only thing that closes it is generation-stamping pending entries per connection — worth doing if the maps ever gain more machinery, not worth it here for a window this narrow.

**Required (P3):** state the limit in the code comment and the PR body. Do not let a later reader believe the pair is a unique key.

**P3 — the review's P2 test recommendation, taken literally, does not fail against Option A.**

The review asks for "a browse and a load sharing a session key where the load is the one refused". Whether a session-key scan resolves the wrong one depends on which map it scans first — an implementation detail of the rejected option. Written against `HashMap` iteration it is worse: the test would be non-deterministic against the very design it is meant to rule out.

What makes the test load-bearing is not that it fails against a hypothetical Option A; it is that it asserts a positive invariant of Option B — the refused request is identified by id, and a same-session sibling in the *other* map is untouched. That is deterministic and checkable. The PR should describe it that way rather than claiming it "pins the case a scan gets wrong", which is only true for one of the two scan orders.

**P3 — routing the image variants is scope creep with a real, if small, regression surface, and "same shape" is not sufficient justification.**

Cover art currently fails by 10s timeout on a bad image key. After this change it fails immediately with "Image not found". That is better — but `get_image` is called from the `/roon/image` route, which the *web UI does* use, unlike `/roon/browse`. So the image change touches a live path while the browse change touches only external-client surface. That inverts the risk ordering implied by the issue title, and the review waved it through on structural similarity alone.

It should still ship — a 10s block per missing cover is worse than a fast failure, and `None` is already a modelled outcome (`src/adapters/roon.rs:563`) — but it deserves its own line in the PR and its own two tests, not a footnote. And the collapse of `ImageUnexpectedError` into "not found" is a real, if minor, dishonesty: an unexpected Core error is reported to the user as a missing image.

**P4 — the 404 argument is right, and the "no client can be relying on the old 500" reasoning is not.**

The status-code change is defensible. But the justification offered — that no client can depend on the old behavior because it was ambiguous — proves too much. A client can perfectly well branch on `500` today and would now take a different branch. The honest argument is narrower: the previous 500 arrived only after a 10s timeout, so any client tuned to that condition was already treating it as an outage, and telling it "your reference is stale" instead is the fix, not a break. Keep the change; fix the reasoning.

**P4 — exhaustive matching on a foreign enum is a deliberate future build break, and calling it "fails loudly" undersells the cost.** A fork bump now cannot compile until someone triages the new variant. That is the right trade for an issue about silently swallowed errors, but the next person to bump the fork should find the reason in the code, not infer it from a compile error.

### Where the review is right

Rejecting the session-key scan is correct, and for a better reason than the review gives: the scan cannot distinguish a browse from a load *at all*, so no amount of care makes it correct — it is not a weaker version of the right answer, it is a different and unsound one.

Rejecting Option D (fork change) is correct and well-reasoned. The success arms genuinely are ambiguous, and fixing them here would have coupled a local defect fix to a dependency release.

The `parse_msg` single-delivery check (`:164`) is the most valuable piece of evidence in the review and is easy to miss. It rules out error-after-success and success-after-error ordering hazards outright, which is why the implementation needs no sequencing logic at all. Without it, the whole design would need a resolved-flag.

Establishing that the error body cannot deserialize as a `BrowseResult` is the other one. If it could, this entire PR would be routing a message that never arrives.

### Required actions

1. **(P2)** Search both maps for a full match before reporting `KeyMismatch`. Test it. — *done, `b24ae86`*
2. **(P3)** State the cross-check's limit (session keys are not unique either; the window is narrowed, not closed) in the code and the PR.
3. **(P3)** Describe the sibling-in-the-other-map test as a positive invariant of the chosen design, not as a refutation of the rejected one.
4. **(P3)** Give the image routing its own PR section and its own tests; name the `ImageUnexpectedError` → "not found" collapse as accepted lossiness.
5. **(P4)** Replace the "no client can be relying on 500" argument with the narrower, true one.
6. **(P4)** Say in the code why `RoonApiError` is matched exhaustively.

### Disposition

Pass-with-actions. Action 1 was blocking and is implemented. 2–6 are PR-body and comment work, carried into the execute and ship gates.

### Residual risk after actions

The reconnect window is narrowed, not closed, and the remaining case needs a client reusing a session key across a reconnect inside 10s. If `/roon/browse` ever gains a caller that holds one session key for a long-lived browse tree — which is precisely what #399 will build — that window widens, because long-lived session keys are exactly the input this guard cannot disambiguate. #399 should generation-stamp the pending maps, or stop reusing session keys across reconnects. Nothing here forces that, and nothing here will remind anyone.

### Next gate authorization

Authorized to proceed to execute gate, with action 1 already incorporated.

muness:
## Gate report — EXECUTE / review

**Gate/role:** Execute gate, reviewer

### Aim and scope checked

Aim unchanged: prompt, correctly-correlated, distinguishable failure for a browse or load the Core refuses. Scope checked at this gate: that the implementation matches the selected design; that the tests were written first and genuinely failed; that no request can now leak, double-resolve or resolve into the wrong caller; and that a valid browse or load is untouched.

### Test-first evidence

Two commits, in that order, both on the branch:

- `99b3779` — tests plus the types and the routing seam, with the routing functions returning `NoWaiter` unconditionally. That is not a stub for convenience: it is a faithful encoding of the pre-#405 behavior (`_ => {}` drops the rejection, nothing is resolved), so the red state is the defect, expressed behind the new seam.

  ```
  test result: FAILED. 8 passed; 9 failed; 0 ignored; 82 filtered out; finished in 10.01s
  ```

  The 10.01s is the finding, not an artifact. `a_rejected_item_key_resolves_far_inside_the_browse_timeout` awaits the real `BROWSE_TIMEOUT` and, with nothing resolving the waiter, waits the whole thing out. The red suite reproduces the ten-second hang as wall-clock time.

- `8df4df3` — the routing implementation.

  ```
  test result: ok. 17 passed; 0 failed; 82 filtered out; finished in 0.00s
  ```

  10.01s → 0.00s is the acceptance criterion "returns promptly, no `BROWSE_TIMEOUT` wait", measured rather than asserted.

Two further commits followed: `b24ae86` (the dissent's blocking finding, with its own test — 18 passing) and `d6b1e1b` (comments only, recording the limits the dissent named).

### Evidence inspected

- Full diff `c47d36a..d6b1e1b`: `src/adapters/roon.rs`, `src/api/mod.rs`. Nothing else touched.
- `cargo clippy -- -D warnings` — clean at every commit checked.
- `cargo test --no-fail-fast` — all suites pass except the two pre-existing `/hqp/discover` failures (`client_harness::shared_endpoints::get_hqp_discover`, `protocol_integration::api_endpoints::hqp_discover_returns_json`), which do live UDP SSDP.
- Hard-constraint lints: `spawn_cancellation_lint`, `unbounded_channel_lint`, `await_in_lock_lint` — pass. Also `oneshot_leak_lint`, `ignored_send_lint`, `architecture_lint`, `aggregator_lint`, `arbitrary_find_lint` — pass.
- `cargo test --test api_contract` — passes; `git diff --stat` confirms `tests/fixtures/api_routes.txt` untouched.
- `dx build --release --platform web --features web` — exit 0, both "Client build completed successfully" and "Server build completed successfully".
- Read for leaks: the two timeout cleanup paths (`browse()` and `load()`), the three delivery paths in the routing functions, and the disconnect cleanup that clears all three pending maps.

### Findings

**P2 — the liveness hole in the mismatch guard, found by the dissent and fixed at `b24ae86`.** The first implementation checked `pending_browses`, and on a key mismatch returned without looking in `pending_loads`. A reconnect id collision could therefore let a stale browse hide a legitimately-refused load, costing it its prompt answer. Now both maps are searched for a full `(req_id, session key)` match before any mismatch is reported. `a_mismatched_entry_in_one_map_does_not_hide_a_match_in_the_other` pins it. Worth stating plainly: the guard against mis-correlation had itself reintroduced a narrow instance of the timeout this issue exists to remove.

**P3 — no leak is possible on any of the six paths, and this was checked path by path rather than argued.**

| Path | Entry removed? | Waiter resolved? |
|---|---|---|
| Rejection matches a pending browse | yes, before send | yes, typed error |
| Rejection matches a pending load | yes, before send | yes, typed error |
| Rejection matches a pending image | yes, before send | yes, `None` |
| Receiver already dropped | yes | n/a — nothing to resolve |
| `KeyMismatch` | **no, deliberately** | no — falls to `BROWSE_TIMEOUT`, whose cleanup removes it |
| `NoWaiter` | nothing to remove | already resolved or already timed out |

`KeyMismatch` is the only non-removing path and the only one where not removing is correct: the entry belongs to a caller that is still waiting, and that caller's own timeout cleanup (`src/adapters/roon.rs:616-619`, `:662-665`) owns it. Removing it there would be the leak — the caller would wait 10s and then find its entry already gone.

**P3 — double-resolve is impossible by construction, not by convention.** Every delivery path calls `HashMap::remove` and takes ownership of the sender before sending. A second rejection for the same id finds an empty slot (`a_second_rejection_for_the_same_req_id_cannot_double_resolve`). The only other remover is the caller's timeout path, and `oneshot::Sender` can be consumed once. There is no flag to get wrong.

**P3 — a valid browse or load is provably untouched.** The success arms are byte-identical in the diff apart from an added comment; the new arm sits between `Parsed::LoadResult` and the surviving `_ => {}`, so no message previously handled reaches different code. `browse()` and `load()` are unchanged — same insert, same timeout, same cleanup — and the 82 filtered-plus-99 other lib tests and 42 adapter-integration tests pass unchanged.

**P3 — the lock is held only across synchronous work.** The routing call sits in a block that acquires `state_for_events.write()`, calls one synchronous function and drops the guard; the logging match runs after. `oneshot::Sender::send` is synchronous and non-blocking. `await_in_lock_lint` passes, and the shape matches the existing `Parsed::Jpeg`/`BrowseResult` arms.

**P3 — every send result is consumed.** `deliver_browse_rejection` returns `send(..).is_ok()`, and the image path uses `send(None).is_ok()`. `ignored_send_lint` passes.

**P4 — `rejection_kind` and the `else`-branches in the tests use `unreachable!`, not `unwrap`/`expect`.** Deliberate: `src/lib.rs:20-22` denies `unwrap_used`/`expect_used`/`panic` crate-wide, and src's test modules already violate it 76 times. The new tests add none, so the pre-existing `--all-targets` count is unchanged by this PR.

### Required actions

1. ~~Search both maps before reporting a mismatch.~~ — done, `b24ae86`.
2. ~~Record in code: the pair is not a unique key; the check is not redundant; why the match is exhaustive; the image lossiness; the success-arm asymmetry.~~ — done, `d6b1e1b`.
3. Carry the unverified portability question into the ship gate as an explicit open item, with the runnable check in the PR body. Do not let it read as verified.
4. State in the PR that this repo runs no CI on feature-branch PRs, so every result above is local.

### Disposition

Pass-with-actions. Actions 1 and 2 are complete on the branch. 3 and 4 are ship-gate obligations.

### Residual risk

The tests exercise the routing functions, the real channels and the real timeout, but not the fork's `parse_msg` and not the three-line result match in `browse()`/`load()`. If `parse_msg` stopped producing `Parsed::Error` — a fork change, or a Core that renamed its error strings — this suite would stay green while the defect returned in full. That is exactly the hole #408 exists to close, and it is the single most important thing for a reader of this PR not to forget.

### Next gate authorization

Authorized to proceed to ship gate.

muness:
## Gate report — EXECUTE / dissent

**Gate/role:** Execute gate, dissent

The correlation logic holds up. My attack moved to the assumption underneath it — that the message ever reaches this code — and to what the green suite is actually evidence of.

### Contrary case

**P1 → resolved — nothing in the review, or in #405, establishes that `Parsed::Error` reaches UHC's event loop at all. If it does not, every test on this branch passes and the ten-second hang is untouched.**

This is the load-bearing assumption of the whole PR and it was asserted from the issue text rather than verified. The issue says the catch-all swallows the error; that presumes delivery. Two places in the fork could eat it first, and the second one is nasty:

1. `lib.rs:620-631` — the browse arm forwards anything that is not `Parsed::None`. Fine.
2. `lib.rs:597-631` — but services are tried **in order, with a `break` on the first that claims the message**, and Transport is tried before Browse. Worse, if `transport.parse_msg` returns `Err`, the fork sends `Parsed::None` and **breaks** (`:613-618`) — the browse error would be discarded inside the dependency, where no amount of UHC code could see it.

Traced to ground:

- `Transport::parse_msg` (`src/transport.rs:451-540`) gates every push on `req_id == <one of its own subscription ids>`. A browse error's id matches none, so it returns `Ok(vec![])` — empty, not `Err`. Empty means `parsed.pop()` is `None`, so no `break`, and the loop proceeds to the Browse service.
- It also does `msg["request_id"].as_str().unwrap()` and `msg["name"].as_str().unwrap()` at `:452-453`. A browse error message carries both (`request_id`, and `name` = `"InvalidItemKey"`), so no panic on this path.
- `Browse::parse_msg` then returns `Parsed::Error(..)`, the fork forwards it on `core_tx` (`:627`), and UHC's event loop is the only consumer.

So the chain holds: **Core → Moo → transport (no claim) → `Browse::parse_msg` → `core_tx` → UHC's event loop → the new arm.** The finding is downgraded from P1 to resolved, but it should never have been an assumption. It is now recorded here because the next person to touch the event loop needs to know the delivery path is order-dependent inside the dependency.

**P2 — the green suite is not evidence that a real Core-rejected item key now resolves promptly, and the PR must not let a reader believe otherwise.**

Everything asserted on this branch starts *downstream* of `Parsed::Error` being constructed with the right `req_id` and session key. Nothing tests that Roon Core's actual error names match what the fork looks for. `Browse::parse_msg` matches exactly four literals — `"InvalidItemKey"`, `"InvalidLevels"`, `"UnexpectedError"`, `"ZoneNotFound"` (`src/browse.rs:174-188`) — against `msg["name"]`. If a Core answers with any other spelling, `parse_msg` returns `Parsed::None`, the fork drops it at `lib.rs:625`, and the caller times out exactly as it does today. **The suite would be green and the bug would be live.**

These are the names Roon's own Node browse API uses, so this is likely right — but "likely" is the whole problem, and this branch cannot close it: the pinned fork's names are unverified against the Core actually on the network (2.70, build 1671). This is precisely #408's "where a response shape was uncertain, record it from a real Core rather than guess", and it is a second reason — independent of correlation testing — that #408 matters to this fix.

**Required (P2):** the PR must state that the *wire* half is unverified, in the same breath as the promptness claim. And the operator's live check answers this as a side effect: if the runnable script returns a fast 404, the wire half is confirmed end to end. That should be said, because it turns an unrelated-looking errand into the missing verification.

**P2 — "resolution is exactly-once by construction" is true of the routing functions and not quite true of the system.**

`browse()` on timeout does `remove(&req_id)`, which drops the sender — the caller has already stopped listening. Now consider the interleaving: the caller's `tokio::time::timeout` expires, and before it re-acquires the write lock at `:617`, the event loop takes the lock and routes a rejection for that id. Routing removes the entry and sends into a receiver that is already dropped → `ReceiverGone`. The caller then re-acquires the lock, removes nothing, and returns `"Browse request timed out"`.

That is correct — one resolution attempt, no leak, no panic, the caller gets today's answer — but it is correct *because the maps are behind one `RwLock` and every mutation is synchronous while it is held*, not because removal is idempotent. The claim should be stated as "serialised by the state lock", which is the property actually doing the work. `rejection_with_a_dropped_receiver_still_clears_the_entry` covers the outcome; nothing states the reason.

**P3 — the recovery instruction is written for a caller that holds a key, and `search()` does not.**

`InvalidItemKey`'s message says "search or browse again to get a fresh key". For `/roon/play_item` and `/roon/browse` that is exactly right. But `search()` mints and consumes its own keys internally across six requests (`src/adapters/roon.rs:682-772`), and a rejection there surfaces through MCP's `hifi_search`, where the client never held a key and cannot act on the instruction as written. It is accidentally still good advice — retrying the search re-mints everything — but nobody chose that. Worth one line so #395/#396 do not inherit a message whose audience is ambiguous.

**P3 — the red commit is honest but the red *number* is doing more work than it should.**

10.01s → 0.00s is the most persuasive line in the PR and it is real. But it measures the promptness test alone; the other eight failures were assertion mismatches that would have failed just as fast with a broken implementation that resolved the *wrong* waiter. A future reader who takes "0.00s" as the proof of correctness will have read the timing and skipped the correlation. The timing proves promptness; only `browse_rejection_resolves_only_the_matching_waiter` and `a_mismatched_entry_in_one_map_does_not_hide_a_match_in_the_other` prove correlation. Say which claim each piece of evidence supports.

**P3 — routing the image variants is now touching the one path the web UI actually uses, and it has two tests.**

Restating from the last gate because the balance has not changed: `/roon/browse` and `/roon/play_item` have zero in-repo callers, while `/roon/image` is used by the UI. So the *incidental* part of this PR carries more live-path risk than the headline part. Two tests and an already-modelled `None` outcome are adequate, and a fast "not found" beats a 10s block on every missing cover — but the asymmetry deserves to be visible to a reviewer rather than discovered by one.

**P4 — `unreachable!` in the test helpers trades one denied macro for one that is merely not denied.** `src/lib.rs` denies `panic`, `unwrap_used`, `expect_used`; `unreachable!` slips through and panics identically. It keeps the pre-existing `--all-targets` violation count unchanged, which is the point, but it is a lint-shaped choice rather than a correctness one and should not be mistaken for rigour.

### Where the review is right

The path-by-path leak table is the right instrument, and its most valuable row is the one that looks wrong: `KeyMismatch` deliberately does not remove. Getting that backwards — "clean up everything you touch" — would silently break the caller's own timeout cleanup and produce exactly the stranded-entry class the issue forbids. The review found the one place where the tidy instinct is the bug.

Reading the success arms in the diff to confirm they are byte-identical, rather than reasoning that they were not touched, is the correct standard for a change that claims to be behavior-additive.

And the review is right that `parse_msg` removing its own bookkeeping entry on first delivery (`src/browse.rs:164`) is what makes the whole design sequencing-free. That fact deserves to survive in the code, not only in a gate report.

### Required actions

1. **(P2)** State in the PR that the wire half — Core error names matching the four literals the fork looks for — is unverified, alongside the promptness claim. Say that the operator's live check confirms it as a side effect.
2. **(P2)** Restate exactly-once as "serialised by the state lock", and record why in the routing doc comment.
3. **(P3)** Note that the `InvalidItemKey` recovery instruction is written for a caller holding a key, and that `search()`'s internal keys surface it too.
4. **(P3)** Attribute the evidence: 10.01s → 0.00s proves promptness; the two correlation tests prove correlation.
5. **(P3)** Keep the image routing visible as its own item, with its live-path asymmetry stated.
6. **(P4)** Do not present `unreachable!` as anything but lint hygiene.

### Disposition

Pass-with-actions. Nothing blocking: the P1 candidate resolved on inspection, and the P2s are honesty-of-claim obligations rather than code changes.

### Residual risk after actions

Two unverified links remain, and they are different in kind. The wire half (do Roon's error names match the fork's literals?) is settled by one live run and is cheap. The delivery-order half (does Transport keep declining to claim browse messages?) is settled only by reading a dependency that could change under a version bump — and there is no test anywhere in this repo that would notice if it did. #408 closes both. Until it lands, this PR's correctness rests on a manual reading of a pinned checkout, and that reading is recorded here rather than in the code.

### Next gate authorization

Authorized to proceed to ship gate.

muness:
## Gate report — SHIP / review

**Gate/role:** Ship gate, reviewer

No merge is requested. This gate checks that what is on the branch is installable, that the evidence is reproducible from the PR body alone, that rollback is real, and that the two open items are stated as open rather than buried.

### Aim and scope checked

Aim: the operator can install this, see the difference, and back it out in one command. Scope checked: the acceptance criteria of #405 one by one; buildability of the shipping artifact; the rollback path; and whether the PR body would let a reader reproduce every claim without this thread.

### Acceptance criteria, one by one

| #405 criterion | State | Evidence |
|---|---|---|
| `Parsed::Error(BrowseInvalidItemKey)` resolves the correct waiter with a typed, recoverable error | met | `RoonState::route_browse_rejection`; `browse_rejection_resolves_only_the_matching_waiter` |
| Returns promptly, distinguishable from unreachable Core and Browse-not-connected | met | `a_rejected_item_key_resolves_far_inside_the_browse_timeout`; `a_rejection_is_distinguishable_from_a_timeout_and_a_lost_core`; suite 10.01s → 0.00s |
| Correlation correct under concurrency, others complete normally | met | 5 requests in flight, 1 rejected, 4 asserted still waiting; plus the same-session browse/load sibling case |
| No stranded entries, no double-resolve; lint suites pass | met | path-by-path table in the execute gate; `spawn_cancellation_lint`, `unbounded_channel_lint`, `await_in_lock_lint`, `oneshot_leak_lint`, `ignored_send_lint` all pass |
| `/roon/browse` and `/roon/play_item` surface the condition meaningfully | met | `roon_browse_failure`; 404 + the Core's explanation + recovery instruction, immediately |
| Test strategy stated explicitly in the PR | met | its own PR section, with the coverage boundary and why #408 owns the fake |
| **Empirical: is an `item_key` portable across session keys?** | **open** | needs a host that can reach the Core; runnable check in the PR body |

Six of seven met. The seventh cannot be met from this machine, and the PR says so in those words.

### Installable / verifiable evidence

- `dx build --release --platform web --features web` on head (`c7acd64`) — exit 0, client and server both reported successful. This is the artifact the operator runs.
- The two build traps are recorded in the PR rather than left for the next person: `make css` first in a fresh worktree, and `~/.cargo/config.toml` sets `[build] rustc-wrapper = sccache`, which `unset RUSTC_WRAPPER` does not override — it needs `CARGO_BUILD_RUSTC_WRAPPER=""`. With sccache active the build failed at `exit status: 254`; with it off, clean. That is a genuine finding: the documented build command does not work on this machine as written.
- No new dependency, no migration, no config key, no persisted state, no route change. `tests/fixtures/api_routes.txt` untouched, `cargo test --test api_contract` green, so no `api-change-approved` is needed.
- Diff is two files: `src/adapters/roon.rs`, `src/api/mod.rs`.

### Findings

**P3 — the status-code change is the only thing in this PR that a client can observe without an error occurring, and it is correctly isolated.** `roon_browse_failure` is the single place any browse/load failure becomes a response, the 404 is gated on two enum variants, and reverting is one expression. It is flagged in the PR as operator-vetoable with the exact revert. Nothing else in the PR needs the operator's judgment.

**P3 — the rollback is genuinely a revert, not a rollback plan.** Four functional commits, no schema, no state, no dependency. Reverting restores the exact pre-#405 behavior: rejections dropped, callers time out. The PR lists the commits.

**P3 — the PR body is self-sufficient.** Every claim carries the command that produced it and the commit it was run on. A reader who never sees this thread can re-run all of it. The one thing they cannot re-run is the live check, and that is precisely what is marked open.

**P4 — two commits in this branch are comments only** (`d6b1e1b`, `c7acd64`). They exist because the gate dissents demanded that the design's limits be recorded in the code rather than in a PR comment that will not be read again. They are worth keeping separate: a reviewer can see exactly what was claimed and where it was written down.

### Required actions

1. Do not merge. AGENTS.md requires the operator's explicit approval, and the portability criterion is still open.
2. Leave the portability question open in #405 rather than closing the issue on merge, or reopen it as its own item if the issue is closed — the answer changes #396's design.
3. When #408 lands, add one end-to-end case through the fake and delete the "wire half unverified" caveat. Not before.

### Residual risk

The correctness of this fix rests on two readings of a pinned dependency that no test in this repo would notice going stale: that Transport keeps declining to claim browse messages (fork `src/lib.rs:597-631`, `src/transport.rs:451-540`), and that Roon's error names still match the fork's four literals (`src/browse.rs:174-188`). Both readings are now recorded in the module's own comments, which is the best this PR can do without #408. A fork bump is the event that would invalidate them, and the exhaustive `match` on `RoonApiError` will catch only the subset of that risk which adds a variant.

### Next gate authorization

Ready for operator review. Not authorized to merge.

muness:
## Gate report — SHIP / dissent

**Gate/role:** Ship gate, dissent

The code is ready. My objections are to what the PR lets a reader conclude, and to one criterion the review marked "met" that is met more narrowly than the tick implies.

### Contrary case

**P2 — "surfaces the condition meaningfully" is ticked on a surface with no in-repo consumer, and the PR should say who is actually affected today.**

`/roon/browse` and `/roon/play_item` have **zero in-repo callers**. The Dioxus UI has no browse page; nothing in `src/app/` calls either route. So the headline user-visible improvement lands on external clients only — the iOS app, the knob, or the operator's own tooling — and neither this PR nor #405 establishes that any of them call these routes. `/roon/play_item` may never have been called by anything, ever, which is the same evidence that made its portability suspect.

The improvement that reaches a *known* consumer today is different and is not the headline: `search()` calls `browse()`/`load()` six times, and MCP's `hifi_search` and `hifi_play` go through it. A rejection anywhere in that sequence previously cost 10s; now it returns immediately. That is the change the operator will actually notice, and it is mentioned only in passing.

The second real one is the image path — `/roon/image` *is* used by the UI, so cover art that the Core refuses now fails fast instead of blocking for 10s. The PR treats that as an aside because it is not in the issue title.

**Required (P2):** state plainly which surfaces have consumers today. `search`/MCP and `/roon/image` are the paths with known callers; `/roon/browse` and `/roon/play_item` are the paths the issue is about. Both are worth fixing, but a reader should not infer from this PR that a user-facing timeout was removed from a path anyone exercises.

**P2 — the PR presents "18 tests pass" and "0.00s" close enough together to read as a correctness proof, and the review repeated the framing.**

The mitigations are now in the PR body and the test module, so this is not blocking. But it is worth stating once more in the record, because it is the single most likely misreading of this branch: **a fast green suite is consistent with an implementation that resolves the wrong waiter promptly.** Only two of the eighteen tests distinguish those worlds. If a future change makes the correlation wrong, sixteen of the eighteen will keep passing and the headline number will not move.

**P3 — the "no new leak" claim is verified against the paths this PR adds, not against the pending maps as a whole, and there is a pre-existing leak window the PR does not mention.**

On `CoreEvent::Lost`, the pending maps are not cleared where the loss is observed; they are cleared only when the run function returns (`src/adapters/roon.rs`, end of the loop). Between those two points, callers inside their 10s hold entries that can never be resolved, because the connection carrying their reply is gone. They time out — correct, but slowly, and that slow timeout is *exactly* the symptom #405 exists to remove, for the neighbouring cause.

This PR does not make it worse, does not touch it, and could not have fixed it without touching connection lifecycle. But "no new leak" is a narrower statement than a reader will hear, and there is a real 10s-ambiguity left in the adapter after this merges. It deserves an issue rather than silence: **on `CoreEvent::Lost`, drain all three pending maps and resolve every waiter with a distinct "Roon Core connection lost" error.** That would be the natural sequel and it reuses everything this PR built.

**P3 — routing all four browse variants is right, but only one of the four has any evidence of occurring in practice.**

`InvalidItemKey` is the one #405 traced. `InvalidLevels`, `ZoneNotFound` and `UnexpectedError` are routed on the reasonable argument that they leak identically — but nothing establishes that a Core emits them, or with those exact names. So three of the four are speculative surface: harmless if never emitted, and untested against a real Core if they are. The PR should not imply four verified paths where it has one traced path and three by symmetry.

**P4 — the `is_recoverable()` boundary is a judgment presented as a fact.** `InvalidItemKey` and `InvalidLevels` are marked recoverable; `ZoneNotFound` is not, on the reasoning that a caller cannot fix it by re-acquiring a reference. But a caller *can* fix it by naming a different zone, which is just as actionable. The distinction being drawn is "retry after re-acquiring your reference" versus "fix your input", and only the doc comment says so. #395's envelope will have to decide whether those are one category or two; it should not inherit this line as settled.

**P4 — the sccache finding is more valuable than its footnote.** The documented build command in AGENTS.md does not work on this machine as written: `~/.cargo/config.toml` sets a `rustc-wrapper` that `unset RUSTC_WRAPPER` cannot override, and the failure is an opaque `exit status: 254` with a thousand-character command line and no mention of sccache in the error. The next agent will lose the same twenty minutes. That belongs in the repo's build docs, not only in this PR body.

### Where the review is right

Walking the acceptance criteria as a table and marking the seventh **open** rather than "met with caveats" is the correct call, and it is the one place where the temptation to round up is strongest. Six of seven, with the seventh named and scripted, is an honest result.

Isolating the status-code decision behind a single function so that the veto is a one-expression change is good engineering *and* good politics: it makes the operator's decision cheap, which is what makes it a real decision rather than a formality.

And the review is right that the two comment-only commits belong in the history. The dissents' findings were about what the code fails to say about itself; a comment-only commit is exactly the right shape for that fix, and keeping it separate makes it auditable.

### Required actions

1. **(P2)** State which surfaces have in-repo consumers: `search`/MCP and `/roon/image` do; `/roon/browse` and `/roon/play_item` do not.
2. **(P3)** Open a follow-up for draining the pending maps on `CoreEvent::Lost`, resolving waiters with a distinct connection-lost error. Reference it from the PR.
3. **(P3)** Say that `InvalidItemKey` is the traced case and the other three are routed by symmetry, not by observation.
4. **(P4)** Note that the `is_recoverable` boundary is a judgment #395 may revisit.
5. **(P4)** Record the sccache/`CARGO_BUILD_RUSTC_WRAPPER` trap where the next agent will find it, not only here.

### Disposition

Pass-with-actions. None blocking; all five are honesty-of-claim or follow-up items. Actions 1, 3 and 4 are PR-body edits; 2 is a new issue; 5 is a docs change outside this PR's scope and should not be smuggled into it.

### Residual risk after actions

After this merges, the Roon adapter still has one 10s ambiguity: a request in flight when the Core connection drops. It has the same shape as the defect just fixed, the same fix pattern is now available, and nothing in the codebase points at it. Action 2 is the only thing standing between "noticed" and "forgotten".

### Next gate authorization

Ready for operator review. Not authorized to merge. Do not close #405 on merge while the portability criterion is open.

muness:
## Gate report — SHIP / dissent, retraction of P3

**Gate/role:** Ship gate, dissent
**Verdict on the retracted finding:** withdrawn — the finding was false

I asserted, in the ship-gate dissent above, that a request in flight when the Core connection drops still hangs for the full `BROWSE_TIMEOUT`, and I recommended opening a follow-up issue to drain the pending maps on `CoreEvent::Lost`. I drafted that issue. Before filing it I went to read the `CoreEvent::Lost` arm, and it does not say what I claimed.

**What the code actually does** (`src/adapters/roon.rs:1935-1947`):

```rust
{
    let mut s = state_for_events.write().await;
    s.connected = false;
    …
    s.pending_images.clear();
    s.pending_browses.clear();
    s.pending_loads.clear();
}
```

All three pending maps are cleared **in the `Lost` arm**, immediately, not at teardown. Clearing drops every sender, so each waiting receiver resolves at once with `RecvError`, and the callers already handle that distinctly:

- `browse()` → `Ok(Err(_))` → `"Browse request cancelled"` (`:623`)
- `load()` → `Ok(Err(_))` → `"Load request cancelled"` (`:669`)
- `get_image()` → `Ok(Err(_))` → `"Image request cancelled"` (`:564`)

So on connection loss the behavior is already prompt, already leak-free, and already distinguishable from `"Browse request timed out"`. There is no second 10s ambiguity, and no issue to file. My "the post-join `clear()` is the only clear" reading was simply wrong — I had seen the teardown block and not the one in the `Lost` arm.

**The residual point that survives, at P4:** `"Browse request cancelled"` does not say *why* it was cancelled — connection lost, adapter restart, or shutdown all produce the same string. That is a message-quality nit on an already-prompt path, not a hang, and it is not worth an issue.

**What this says about the gate process.** The finding was plausible, specific, cited a file, and was wrong. It was caught only because filing the follow-up required re-reading the code, and the draft issue would have sent the next agent to fix a bug that does not exist. A dissent's job is to be adversarial, which makes it exactly as capable of manufacturing a defect as a review is of missing one — so a dissent finding that recommends *new work* has to be verified to the same standard as a review finding that blocks work. This one was not, until the last possible moment.

The corresponding paragraph has been removed from the PR body and replaced with the verified statement, so that the next reader does not inherit the error from the comment thread.

Ship-gate verdict is unchanged: **pass-with-actions**, actions 1, 3, 4 applied to the PR body; action 2 withdrawn; action 5 (the sccache/`CARGO_BUILD_RUSTC_WRAPPER` build trap belongs in the repo's build docs) stands, outside this PR.

muness:
## Gate addendum — SHIP / review, correction: CI is a gate here, and it caught something

**Verdict:** unchanged (pass-with-actions), with one premise corrected and one action added.

Both gate reports above, and the PR body as first written, asserted that this repo runs no CI on feature-branch PRs, so local verification was all that was available. **That is wrong for this PR.** The note applies to PRs whose *base* is another feature branch. This one bases on `v3`, so the full workflow set ran: `Lint`, `Build`, `API Guard`.

And it caught a real failure that every local check had passed:

```
Lint / Check formatting → cargo fmt --check → exit 1
  Diff in src/adapters/roon.rs:2428, 2677, 2705, 2778 (+1)
```

Five formatting hunks, all in the new test module. `cargo clippy -- -D warnings` was clean the whole time, and I had treated clippy as the gate because #405 and the epic both name it. **Lint runs `cargo fmt --check` before clippy** (`.github/workflows/build.yml:366-369`), so a fmt failure short-circuits the job in 21s and clippy never runs. Fixed in `96d3ebd`.

Also passing on that run: `Build Linux x64`, `Build WASM Assets`, `Build Summary`, and `API Guard` (which is the route-contract check — independent confirmation that `tests/fixtures/api_routes.txt` is untouched and no `api-change-approved` is needed).

Two things worth carrying forward:

1. **The verification recipe in the epic is incomplete.** `cargo clippy -- -D warnings` plus the test suites is what every issue in #392 names, and it is not sufficient — `cargo fmt --check` is a separate gate that fails first. Any agent that verifies locally per the epic's list and stops there will hand over a red PR.
2. **Where CI is real, use it.** The "zero CI" caveat is load-bearing for stacked PRs and misleading for `v3`-based ones. This PR's evidence is stronger than it claimed: the local runs *and* a green Lint/Build.

Added action, for the epic rather than this PR: state in #392's execution contract that local verification means `cargo fmt --check`, `cargo clippy -- -D warnings`, the test suites, and `dx build` — and that a PR based on `v3` gets real CI whose result should be quoted rather than dismissed.

muness:
## Gate addendum — SHIP, CI evidence

CI is green on head (`96d3ebd`):

| Check | Result |
|---|---|
| Lint (`cargo fmt --check` + `cargo clippy -- -D warnings`) | pass, 47s |
| Test | pass, 1m50s |
| Build Linux x64 | pass, 3m57s |
| Build WASM Assets | pass, 4m44s |
| Check API Changes | pass |
| Smoke Test Binary | pass |
| Synology Package Contract | pass |
| Build Summary | pass |

`Check API Changes` passing is independent confirmation that the route contract is untouched — no `api-change-approved` needed. CodeRabbit reports "Review rate limited", so there is no automated review to address yet.

One thing that came out of this and is useful beyond this PR: **CI's `Test` job is green, including the two `/hqp/discover` tests that fail locally.** Their failure is a local-environment artifact of live UDP SSDP, not something anyone should chase. Every issue brief in this epic warns about those two failures; it is worth knowing they are green on the runner.

Ship-gate verdict stands: **ready for operator review, not authorized to merge.** The portability criterion remains open and needs a host that can reach the Core.

muness:
## Operator decision recorded

**404-vs-500 (§6.4 of the epic hand-off):** approved. Keep the 404 behavior on `/roon/browse` and `/roon/play_item` as shipped. No revert needed.

PR #413, opened 2026-08-01
test(roon): FakeRoonCore, a protocol-level Roon Core fake, and the first Roon success-path assertions (#408)

### LINKED ISSUE #394: MCP foundation: modularize the MCP server and add a contract test harness

### LINKED ISSUE #408: Test infrastructure: MockRoonCore is not a server, so no Roon success path is asserted

muness:
Fixes #408

`MockRoonCore` is not a server. **`FakeRoonCore` is.** It speaks Roon's MOO protocol over a real WebSocket, so `RoonAdapter` can be driven end to end without a Core, a network, or multicast — and for the first time a test asserts a Roon *success* path.

---

## CI is a real gate on this PR

An earlier draft of this section said feature-branch PRs run zero CI. **That was wrong, and it is worth saying why:** the workflow triggers on `pull_request: branches: [master, v3]`, which keys on the PR's **base**, not on the head being a feature branch. This PR targets `v3`, so it gets the full Lint / Test / Build set and a red check is blocking.

The correction cost a real failure: the first push was green on local `cargo clippy -- -D warnings` and still failed CI's **Lint** job on `cargo fmt --check` alone, which clippy does not catch. Both are now run locally before every push.

Consequence for the `/hqp/discover` failures below: they fail on a dev Mac because they do live UDP SSDP, and CI's Test job passes them. So they are a **local-environment artifact**, not pre-existing failures — an earlier draft of this body wrote them off as the latter, which was a second version of the same mistake.

The full gate, now run locally before every push: **Lint** = `cargo fmt --check` + `cargo clippy -- -D warnings`; **Test** = `cargo test --workspace`; `make css` first in both.

Settled on this PR: **Lint pass · Test pass · Plan pass · Build WASM Assets pass · Synology Package Contract pass · CodeRabbit pass.**

## The hole, and the proof it is closed

#394's execute gate found it: *"if the module split had swapped `title` and `subtitle` in Roon's search mapping, none of its 48 contract tests would have noticed."* #396's design gate hit the same wall from the other side — no browse mock, so it could not state an honest Roon test strategy.

There are **four** structurally identical `item.title -> title, item.subtitle -> subtitle` projections in this repo. Three are now covered with asymmetric values, and each was verified to fail when swapped.

### 1. `SearchResultItem::from` — `src/api/mod.rs:402`

```
$ # title: item.subtitle.clone().unwrap_or_default(),  subtitle: Some(item.title),
$ cargo test --test roon_protocol search_results_survive
test search_results_survive_the_http_boundary ... FAILED

thread 'search_results_survive_the_http_boundary' panicked at tests/roon_protocol.rs:263:5:
assertion `left == right` failed
  left: String("Miles Davis")
 right: "Kind of Blue"
```

### 2. `BrowseItemResponse` in `roon_browse_handler` — `src/api/mod.rs:681`

```
$ # same swap
$ cargo test --test roon_protocol browse_items_survive
test browse_items_survive_the_http_boundary ... FAILED

thread 'browse_items_survive_the_http_boundary' panicked at tests/roon_protocol.rs:322:5:
assertion `left == right` failed
  left: String("")
 right: "Library"
```

### 3. The adapter's own navigation — `src/adapters/roon.rs:766`

Not a field swap but the same class of silent-wrong-data bug: a refactor that returns the wrong browse level. `search()` changed to return `source_items.items` instead of `load_result.items`:

```
$ cargo test --test roon_protocol search_maps_title
test search_maps_title_and_subtitle_without_swapping_them ... FAILED

thread 'search_maps_title_and_subtitle_without_swapping_them' panicked at tests/roon_protocol.rs:192:5:
assertion `left == right` failed: expected exactly one match
  left: 3
 right: 1
```

All three reverted; `git diff` clean; 46/46 green again.

### 4. `McpSearchResult` — `src/mcp/mod.rs:483` — **deliberately not covered**

#394 (PR #404) is rewriting all of `src/mcp/` and owns `tests/mcp_contract.rs`. Covering it here would either duplicate that harness or conflict with it. With this fake merged, the assertion is a few lines in #394's harness: start a `FakeRoonCore`, point the harness's existing `AppState` at an adapter connected to it, and assert `hifi_search`'s Roon branch returns `title: "Kind of Blue"`, `subtitle: "Miles Davis"`.

Stated as a gap rather than papered over. It is the only one of the four still uncovered, and every layer beneath it is now covered — so that one `.map()` is the sole remaining place this bug can hide.

## What the fake covers

`tests/mock_servers/roon_core.rs` (1337 lines), `tests/roon_protocol.rs` (25 tests), plus 6 unit tests in the fake and the pre-existing mock self-tests — 49 in the binary, **1.7s**.

| Area | Asserted |
|---|---|
| Handshake | `registry:1/info` at request id 0 → `register` → `CoreEvent::Registered`; Browse becomes available; the pairing token is persisted — and the test proves it landed in a temp dir, not the operator's config |
| Zones | a `Subscribed` payload survives the fork's deserializer and reaches `get_zones()` with volume intact |
| Search | the full six-request sequence; **one** `search_{nanos}` session key throughout; query carried as `input`; `hierarchy=browse` on every request; Library/TIDAL/Qobuz select different branches; empty results are not an error |
| **title/subtitle** | asymmetric values at three layers (above) |
| Browse | two levels down and `pop_all` back to the root, asserted against the Core's own level stack; sessions proven independent; `input_prompt` advertised; unkeyed rows |
| Load | offset/count paging, short final page, total count vs page size, past-the-end empty |
| `play_item` | an `item_key` resolves through the item's action list to `Play Now`; the exact walk is pinned (`Library → Search → Kind of Blue → Play Album → Play Now`); the `roon:` prefix is stripped before it reaches the Core |
| `search_and_play` | navigates into a hit to find a playable action; `play`/`queue` invoke different actions; a missing action reports what *is* available |
| **Errors on demand** | `InvalidItemKey` and `InvalidLevels`, correlated to the right request id and session key, including with three requests in flight and responses arriving out of order |
| Drift | the set of request names the adapter sends is closed; anything else is answered `InvalidRequest` and fails an assertion |

## What it deliberately does NOT cover

Written up in full in `tests/mock_servers/README.md` so the next agent does not over-trust it. The load-bearing entries:

- **It cannot prove the adapter matches a real Roon Core.** The fake's semantics came from the pinned fork's deserializers and from this repo's own adapter — the same source of truth the adapter was written against, and one that has never itself been checked against Roon. **Green means "unchanged", not "correct."**
- **Category-grouped search results.** A real Core groups hits under `Albums`/`Tracks`/`Artists` rows. The default library returns them flat, so `is_category`, `try_category_playable` and the second half of `try_navigate_to_playable` are **untested**. `FakeLibrary` and the `FakeItem` builders are public so a test can build a grouped library — but nobody has verified what the real grouping looks like, so the fake ships no guess.
- **Transport control, images, queue** (`subscribe_queue` / `play_from_here`, which #400 needs), **zone updates over time** (`Changed`, `zones_added`, seek), **reconnection / Core loss**, **the unauthorized-extension state**.
- **Roon's real error bodies.** The fake sends error names with no body.

## Which shapes are recorded, which are documented, and which are still guesses

**No live Roon Core was reachable**, so nothing here was recorded off a Core by this work:

```
$ ls ~/Library/Application\ Support/unified-hifi-control/unified-hifi/
hqp-config.json   knobs.json   lms-config.json          # no roon_state.json
```

No pairing token exists on this machine, and getting one needs a human in Roon's UI.

That is where an earlier draft of this PR stopped, and stopping there was the weak point. **#407 found a real client-visible bug precisely because it recorded a response instead of hand-writing what it expected** — a mock that agrees with a wrong premise is the exact failure #408 exists to end. So the uncertain shapes were chased into published sources rather than left as author-flavoured guesses. Six of them moved out of INFERRED.

### Documented by RoonLabs themselves

From `RoonLabs/node-roon-api-browse` and `RoonLabs/node-roon-api`, quoted at the use sites:

| Shape | RoonLabs' words | Was |
|---|---|---|
| `list.level` | **"increases from 0"** | INFERRED |
| `action` | `message`, `none` ("No action is required"), `list`, `replace_item`, `remove_item` | INFERRED |
| `hint` | `null` ("Unknown"), `action`, `action_list`, `list`, `header` — exactly the fork's five | fork-only |
| `input_prompt` | `prompt`, `action` ("The verb that goes with this action"), `value`, `is_password` | fork-only |
| `InvalidRequest` + `error` string | what `node-roon-api` itself sends for "unknown service" / "unknown request name" | fork-only |

`list.level` mattered: #399's "every browse page reports its position" rests on it, and it is now documented rather than assumed.

### Recorded off real Cores — by third parties, not by me

| Observation | Source |
|---|---|
| `MOO/1 COMPLETE InvalidItemKey` **verbatim** — `Could not play id:122:4, result: MOO/1 COMPLETE InvalidItemKey` (also `130:12`, `133:13`, `109:7`, `135:5`) | `home-assistant/core#137605` |
| real `item_key`s are two colon-separated integers that **"change with each refresh"** — `"115:0"` → `"116:0"` | Roon Labs community thread 23129 |

The first confirms both the error name **and** that the reply is verb-and-name with no body — which is exactly the frame this fake sends. The second confirms the key shape the fake mints. These are real Cores, but other people's logs rather than a controlled run, and they are labelled that way.

### Still guesses, and now with a source saying so

| Shape | Consequence if wrong |
|---|---|
| root list title `"Explore"` | none — never read |
| a real Core sends *no body* with an error | none for the fork; a real body parsing as a `BrowseResult` would read as success |
| `action: "none"` is what a Core picks for an invoked action | none today — the adapter discards it |
| **`InvalidLevels` / `UnexpectedError` / `ZoneNotFound` spellings** | **dropped inside the dependency; caller times out as if the Core were unreachable, every test here still green** |
| search results flat rather than category-grouped | `is_category` / `try_category_playable` untested |
| **`item_key` portability across `multi_session_key`s** | see below — this one now has contrary evidence |

**Only `InvalidItemKey` of the fork's four names is corroborated.** RoonLabs' published browse API documents errors solely as "an error code or false if no error" and never enumerates them; the other three appear nowhere in `node-roon-api`. A community thread shows the prose "Zone not found", which is not evidence of a wire spelling. **#408 does not close the verification #412 assigned it — it narrows it from four unknowns to three and pins the consequence.**

### The epic's `item_key` question: third-party evidence points against the repo's assumption

`RoonAdapter::play_item` mints a *fresh, unrelated* session key and browses the caller's key inside it, so the repo assumes keys are portable across sessions. Two recorded observations point the other way:

* `home-assistant/core#137605` — playing a Roon playlist from a script "works successfully on the first attempt but fails on the second", with `MOO/1 COMPLETE InvalidItemKey`. **That is the same shape as UHC's `play_item`:** hold a key from an earlier lookup, browse it later.
* Community 23129 — item keys "change with each refresh".

That is `ItemKeyScope::PerSession` behaviour. **Not proof** — Home Assistant is not UHC and neither source states the rule — but it is enough that **#396 should not design on the assumption keys are portable**, and enough that `/roon/play_item` should be assumed broken until the operator's rig says otherwise. This is the epic's empirical unknown #1, and it now has evidence rather than only a question mark.

The fake's default stays `ItemKeyScope::Global` because that is what the adapter's code assumes and these tests describe the adapter. The default is explicitly not a claim about Roon, and `a_foreign_item_key_is_rejected_when_keys_are_session_scoped` exercises the other setting.

## Production-code change — one commit, review it as a behaviour change

`707833e`, `src/adapters/roon.rs`, +79/−9, three hunks:

1. `CoreConnect { Discovery, Direct { ip, port } }` — a private enum.
2. `run_roon_loop` takes it; `AdapterLogic::run` passes `Discovery` and the discovery call is otherwise unchanged.
3. `#[doc(hidden)] pub async fn run_event_loop_against_core_for_tests(ip, port)` — no production callers.

**None of it touches the `Parsed::` match arms at `:1860-1915`**, which is what #405 modifies. The hunks are at `:334`, `:1344` and `:1504-1600`. Conflict risk with #405 is a context-line rename at most; if #405 lands first I will rebase.

Why a seam at all: the whole value of the fake is that it drives the **real** event loop and its pending-request correlation maps. A test that stood up its own loop would assert nothing about `roon.rs` — precisely the failure mode that makes mocks dangerous. Rejected alternative: have the fake answer SOOD so no seam is needed. SOOD is UDP multicast, process-global, and already the cause of this repo's two flaky `/hqp/discover` tests.

## Dependency added

`tokio-tungstenite = "0.21"` under `[dev-dependencies]` — the exact version `roon-api` already uses, so the fake and the real client library share one framing implementation. **No new crate enters the tree:**

```
$ git diff v3 -- Cargo.lock
+  "tokio-tungstenite 0.21.0",     # one line, in this package's own dep list
```

## This is #405 / PR #412's end-to-end proof, and it was verified as such

#412's own correlation tests are unit-scoped *precisely* because nothing could drive the wire path. An earlier draft of this PR pinned the pre-#405 hang; that was the wrong instrument, because it would have turned red the moment the defect was fixed.

Four tests drive a Core rejection — bad item key, foreign key under `ItemKeyScope::PerSession`, load with no browse, and one rejection among three concurrent browses. None asserts "it hangs". `classify_rejection()` enforces what must hold on either base:

* a rejected request **never** resolves as success;
* if it resolves at all, it resolves promptly, its message does not read as a timeout, and it names the browse session.

…and reports which adapter it saw. **Verified both ways — same file, same assertions, both green:**

```
$ # on v3 (c47d36a), #405 not present
load without a browse: DroppedByAdapter
rejected browse: DroppedByAdapter
rejected browse among three in flight: DroppedByAdapter
play_item against a foreign key: DroppedByAdapter
test result: ok. 49 passed

$ git merge origin/fix/issue-405-roon-browse-errors      # local, throwaway branch
$ cargo test --test roon_protocol -- --nocapture
load without a browse: RoutedToCaller
rejected browse: RoutedToCaller
rejected browse among three in flight: RoutedToCaller
play_item against a foreign key: RoutedToCaller
test result: ok. 49 passed
$ cargo test --lib && cargo clippy -- -D warnings && cargo fmt --check
100 passed; clippy clean; fmt clean
```

So the fake genuinely drives #412's `Parsed::Error` arm, and this suite needs no inversion on merge.

Pinned unconditionally, on both bases: the Core answered, named the error, and correlated it to the rejected request's own `req_id` **and** session key — exactly the `Parsed::Error(BrowseInvalidItemKey((req_id, multi_session_key)))` payload. `FakeRoonCore::errors_sent()` records refusals as that triple.

**Base stays `v3`, deliberately.** Rebasing onto `fix/issue-405-roon-browse-errors` would make the PR's base a non-`v3` branch, and CI keys on the base — I would lose the entire Lint/Test/Build set. A suite that is correct on both bases is strictly better than one that trades CI for a tighter type assertion. When #412 is on `v3`, tighten `classify_rejection`'s `Ok(Err(_))` arm to `RoonBrowseError::from_error(&e)` and assert `kind == InvalidItemKey` directly; the string checks are the strongest assertions expressible without that type.

## The gap #412 handed to this issue, and what I could and could not do about it

#412's module comment says it plainly:

> `Browse::parse_msg` matches four literal error names against `msg["name"]` (fork `src/browse.rs:174-188`). A Core that spells one differently yields `Parsed::None`, which the fork drops, and the caller times out as it did before — with this module's tests still green. That half is wire-level and is #408's to pin.

**I cannot pin it, and neither can any fake.** This fake sends the fork's own literals, so it agrees with the fork by construction. Verifying `"InvalidItemKey"` / `"InvalidLevels"` / `"UnexpectedError"` / `"ZoneNotFound"` against a real Core is the only way, and no Core was reachable.

What I did instead — pin the *consequence*, so the risk is a test rather than a paragraph:

* `FakeRoonCore::reject_item_key_with_name()` emits an arbitrary error name.
* `an_unrecognised_error_name_degrades_to_an_indistinguishable_timeout` makes the Core answer instantly with `"ItemKeyNoLongerValid"` and asserts the caller still waits out `BROWSE_TIMEOUT`, and that the Core's refusal *was* correlated and logged. It passes on `v3` **and** with #412 merged — the fix does not help when the name is unrecognised.
* `FakeRoonCore::FORK_ERROR_NAMES` plus `the_forks_browse_error_names_are_pinned` make a fork bump that renames one fail here rather than as a mysterious timeout somewhere else.

**Open, needs the operator's rig:** do a real Core's error names match those four literals? If any differs, `an_unrecognised_error_name_degrades_to_an_indistinguishable_timeout` is the failure you will be looking at.

## Designed to fail loudly instead of reading as coverage

The `MockHqpServer` precedent is the whole reason this section exists: it rejected its own adapter's wire format for months and read as coverage. Three defences:

1. **Request log** — tests assert what the adapter *sent*, not only what came back. Same fix #394 applied to `MockLmsServer`.
2. **Response log** — `FakeRoonCore::responses()`, so "the Core never answered" and "the Core answered and the adapter dropped it" are distinguishable.
3. **Closed request set** — anything unmodelled is answered `InvalidRequest`, recorded as unhandled, and `assert_no_unhandled_requests()` fails. `the_fake_models_everything_the_adapter_sends` additionally pins the exact set of five request names, so a new adapter call is a visible edit rather than a silent hang.

And the fake is never validated by a hand-written client — only by the real `roon_api`. If its framing breaks, `roon_core_completes_handshake` fails first.

## Found by the instrument, not fixed here

Per the epic ("anything it finds gets its own issue"):

### 1. Concurrent browses in one session cross-deliver their results — silently

**This one needs its own issue.** Two browses in flight under the same `multi_session_key` hand each caller the *other's* list, with no error. Measured, seven trials in eight:

```
SCRATCH trial 0: asked ["Library", "TIDAL", "Qobuz"] got ["TIDAL", "Qobuz", "Library"]
SCRATCH trial 1: asked ["Library", "TIDAL", "Qobuz"] got ["Library", "TIDAL", "Qobuz"]
SCRATCH trial 2: asked ["Library", "TIDAL", "Qobuz"] got ["Qobuz", "TIDAL", "Library"]
SCRATCH trial 3: asked ["Library", "TIDAL", "Qobuz"] got ["TIDAL", "Library", "Qobuz"]
...
SCRATCH crossed 7/8
```

`pending_browses` is keyed by `req_id`, but the `Parsed::BrowseResult` arm scans by session key — because `Parsed::BrowseResult` carries no `req_id` — and `.find()` returns an arbitrary match. **It survives #405**: #412 fixes the *error* arm's correlation and its own comment says the success arm's gap needs a change to the pinned fork.

Reachability: `search`, `search_and_play` and `play_item` each mint a private key and run sequentially, so they are safe. `POST /roon/browse` takes a **caller-supplied** `session_key` and has no in-repo callers today, so nothing here triggers it — but any external client can, and **#399's navigation handle will hand clients a reusable session identity, which is exactly the shape that triggers it.** Worth resolving before #399 designs around it.

`concurrent_browses_in_one_session_cross_deliver_their_results` asserts the defect is present over eight trials (one trial is right by luck about one time in eight). When it starts failing, the bug is fixed — invert it.

### 3. **`MockRoonCore` has no callers.** Not "used by a few tests" — none:
   ```
   $ grep -rn "MockRoonCore" tests/ | grep -v tests/mock_servers/roon.rs
   tests/mock_servers/mod.rs:19:pub use roon::MockRoonCore;
   ```
Its only exercisers are its own six unit tests. Left in place; flagged as a removal candidate.

## What I ran locally

```
$ cargo test --test roon_protocol
test result: ok. 49 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 1.72s

$ cargo clippy -- -D warnings            # CI's Lint job, part 1
Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.59s      (exit 0)

$ cargo fmt --check                      # CI's Lint job, part 2 - clippy does not catch this
(no output)

$ make css && cargo test --workspace     # CI's Test job, exactly
... one failure: client_harness shared_endpoints::get_hqp_discover (live UDP SSDP)
    CI's Test job passes it - see the CI note at the top

$ cargo clippy --test roon_protocol      # zero findings in either new file
(only the pre-existing unused-re-export warnings in tests/mock_servers/mod.rs)

$ cargo test --lib
test result: ok. 86 passed; 0 failed

$ make css && dx build --release --platform web --features web
#   rustup toolchain ahead of Homebrew's on PATH, RUSTC_WRAPPER unset
 62.99s  INFO Client build completed successfully! 🚀
189.45s  INFO Server build completed successfully! 🚀
```

Every other test target, run individually — all green, all unchanged:

```
adapter_integration    47      protocol_schema        41
zones_sha_integration  25      volume_safety          14
volume_step             9      architecture_lint       8
arch_packaging          6      await_in_lock_lint      4
ignored_send_lint       4      arbitrary_find_lint     3
release_hardening       3      aggregator_lint         2
api_contract            2      oneshot_leak_lint       1
spawn_cancellation_lint 1      unbounded_channel_lint  1
client_harness         41  (+1 local-only failure)
protocol_integration   32  (+1 local-only failure)
```

The two failures are the `/hqp/discover` ones, which do live UDP SSDP broadcast and fail on a dev Mac. **They are a local-environment artifact, not this branch's and not pre-existing repo breakage** — CI's Test job passes them, including on this PR. Confirmed unrelated to the branch by stashing it and reproducing at `c47d36a`:

```
$ git stash push --include-untracked
$ cargo test --test protocol_integration api_endpoints::hqp_discover_returns_json
test result: FAILED. 0 passed; 1 failed; 32 filtered out
$ cargo test --test client_harness shared_endpoints::get_hqp_discover
test result: FAILED. 0 passed; 1 failed; 41 filtered out
```

Flakiness check, because these are timing- and socket-dependent:

```
$ for i in 1 2 3 4 5; do cargo test --test roon_protocol; done
ok  (1.60s / 1.61s / 1.58s / 1.59s / 1.93s)

$ cargo test --test roon_protocol -- --test-threads=1     # ok, 9.65s
$ cargo test --test roon_protocol -- --test-threads=16    # ok, 1.86s
```

`tests/fixtures/api_routes.txt` untouched — no `api-change-approved` label needed.

## Deliberate calls I made

- **Base stays `v3`, and the rejection tests are written to be correct with or without #405.** Rebasing onto `fix/issue-405-roon-browse-errors` would have let me assert `RoonBrowseError` by type, but the PR's base determines whether CI runs at all — so that trade buys a stronger assertion with the entire Lint/Test/Build gate. Verified against both bases instead.
- **A separate fake, not an extended `MockRoonCore`.** #408 left this open. `MockRoonCore`'s state model is a different shape, it has no callers whose compatibility needs preserving, and conflating "in-memory state holder" with "protocol server" is how the confusion arose. Both now exist side by side, with the README saying which is which.
- **The library is data the test supplies; only the protocol is the fake's.** `FakeLibrary` and `FakeItem` are public builders. This is what keeps the fake from becoming a second implementation of Roon's semantics that can drift from the fork — the constraint #408 named.
- **The unverifiable question is configurable, not decided.** `ItemKeyScope` exists so #405 can pin either answer rather than inheriting my guess.
- **Item keys are minted by the Core and unguessable by the test.** Nonce-scoped, so a test must read keys out of a load result the way a client does. `key_for_title` / `title_for_key` exist for assertions; a test cannot construct a key. A key from one fake instance does not resolve in another, and that is asserted.
- **Errors are injected by key, not by call count.** `reject_item_key` survives concurrency; "fail the Nth call" would not.
- **Per-request tasks and per-key delays.** A real Core pipelines. Serialising responses would make it impossible to have several requests genuinely in flight, which is the only way to test correlation.
- **A 3s keepalive ping.** The fork's reader times out after 10s of silence and the client then reports a lost Core, so without this any test slower than 10s would fail for the wrong reason. The client's built-in Ping service answers it, which exercises the provided-service path for free.
- **Assertions against the Core's own level stack**, not only against adapter return values. `pop_all` is only really tested if you can see the stack it emptied.
- **`UHC_CONFIG_DIR` set once inside a `OnceLock` initializer**, so `set_var` runs exactly once and every later read is ordered after it. And the handshake test asserts the state file landed in the temp dir — if a refactor starts writing to the operator's real config, a test fails.

## Rollback

Self-contained. Nothing outside `src/adapters/roon.rs` (one reviewable commit), two new test files, one README, and a dev-dependency line.

- **Whole branch:** revert `a550674`, `cc74ca0`, `a64b745`, `19bed47`, `707833e`, `e604c0c` — or don't merge. No migration, no config, no persisted state, no HTTP surface, no route changes.
- **Production seam only, keeping the fake:** revert `707833e`. `tests/roon_protocol.rs` then fails to compile, so that is not a useful intermediate state — the fake and its seam belong together.
- **Runtime:** nothing ships. The seam has no production callers and is `#[doc(hidden)]`.

## Residual risk

1. **The fake's semantics are derived from the same repo as the adapter.** This is the ceiling on what any of it proves. It is stated at the top of both new files and in the README, and it cannot be lifted without the operator's Core. If the fork's understanding of Roon is wrong, these tests agree with it confidently.
2. **`#[doc(hidden)] pub` is not enforcement.** Nothing stops production code calling `run_event_loop_against_core_for_tests`. A cargo feature gate would remove the risk but also remove the function from `cargo test` and from `cargo clippy -- -D warnings`, which is worse. An `architecture_lint` rule asserting zero non-test callers would close it properly; follow-up, not this PR.
3. **One test asserts a defect is present** (`concurrent_browses_in_one_session_cross_deliver_their_results`) and is probabilistic — eight trials, each right by luck about one time in eight, so a false failure is around 1 in 10^7. If someone "fixes" it by deleting rather than inverting, the evidence for the cross-delivery bug is lost. Says so at the assertion and in the README.
4. **Three of the four browse error names remain unverified.** `InvalidItemKey` is corroborated off real Cores; `InvalidLevels`, `UnexpectedError` and `ZoneNotFound` are not, and RoonLabs does not publish them. If Roon spells one differently, every rejection test here stays green and real callers still time out. Mitigated only by pinning the consequence.
5. **`std::env::set_var` in a threaded test binary** is correctly ordered for its own reads but still races with unrelated `getenv` from other threads in principle. Unavoidable while `get_roon_state_path()` reads the environment.
6. **The category-grouped search shape is untested and unverified**, so `try_category_playable` remains dead as far as tests are concerned — and a real Core probably exercises it on every search.

## Gate reports

Solution / execute / ship, each with a review and a dissent, in the comments.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4607,6 +4607,7 @@ dependencies = [
  "tokio",
  "tokio-stream",
  "tokio-test",
+ "tokio-tungstenite 0.21.0",
  "tokio-util",
  "tower",
  "tower-http",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -151,6 +151,10 @@ serial_test = "3"
 syn = { version = "2", features = ["full", "parsing", "visit"] }
 walkdir = "2"
 tempfile = "3"
+# WebSocket *server* for tests/mock_servers/roon_core.rs (the Roon Core protocol fake).
+# Pinned to the same version roon-api uses as its client, so the fake and the real
+# client library share one framing implementation and no new crate is compiled.
+tokio-tungstenite = "0.21"
 
 [profile.release]
 lto = true
--- a/src/adapters/roon.rs
+++ b/src/adapters/roon.rs
@@ -334,6 +334,49 @@ impl RoonAdapter {
         Ok(())
     }
 
+    /// Test seam (issue #408): run the Roon event loop against a Core at a known
+    /// address, skipping SOOD multicast discovery.
+    ///
+    /// **Not for production use** — `AdapterLogic::run` is the production entry
+    /// point and always discovers. This exists so `tests/roon_protocol.rs` can
+    /// drive the *real* event loop (including the pending-browse/load correlation
+    /// maps) against `tests/mock_servers/roon_core.rs`. Driving the real loop is
+    /// the whole point: a test that reimplemented the loop would assert nothing
+    /// about this file.
+    ///
+    /// SOOD-based discovery against a fake was rejected deliberately: it needs UDP
+    /// multicast, which is process-global and already the cause of this repo's two
+    /// flaky `/hqp/discover` tests.
+    ///
+    /// Returns when the Core is lost or `stop()` is called.
+    #[doc(hidden)]
+    pub async fn run_event_loop_against_core_for_tests(
+        &self,
+        ip: std::net::IpAddr,
+        port: &str,
+    ) -> Result<()> {
+        let base_url = self
+            .base_url
+            .read()
+            .await
+            .clone()
+            .unwrap_or_else(|| "http://test.invalid".to_string());
+        let shutdown = self.shutdown.read().await.clone();
+
+        run_roon_loop(
+            self.state.clone(),
+            self.bus.clone(),
+            base_url,
+            shutdown,
+            self.knob_store.clone(),
+            CoreConnect::Direct {
+                ip,
+                port: port.to_string(),
+            },
+        )
+        .await
+    }
+
     /// Stop the Roon adapter (internal - use Startable trait)
     async fn stop_internal(&self) {
         use std::sync::atomic::Ordering;
@@ -1301,6 +1344,7 @@ impl AdapterLogic for RoonAdapter {
             base_url,
             ctx.shutdown,
             self.knob_store.clone(),
+            CoreConnect::Discovery,
         )
         .await
     }
@@ -1460,6 +1504,17 @@ fn roon_zone_to_bus_zone(zone: &Zone) -> BusZone {
     }
 }
 
+/// How the event loop obtains its connection to a Roon Core.
+///
+/// Production always uses [`CoreConnect::Discovery`]; `Direct` exists only for
+/// `RoonAdapter::run_event_loop_against_core_for_tests` (issue #408).
+enum CoreConnect {
+    /// SOOD multicast discovery, then WebSocket to whatever answers.
+    Discovery,
+    /// WebSocket straight to a known address, no discovery.
+    Direct { ip: std::net::IpAddr, port: String },
+}
+
 /// Main Roon event loop
 async fn run_roon_loop(
     state: Arc<RwLock<RoonState>>,
@@ -1467,6 +1522,7 @@ async fn run_roon_loop(
     base_url: String,
     shutdown: CancellationToken,
     knob_store: Option<KnobStore>,
+    connect: CoreConnect,
 ) -> Result<()> {
     tracing::info!("Starting Roon discovery...");
 
@@ -1518,15 +1574,29 @@ async fn run_roon_loop(
     let state_path_clone = state_path_str.clone();
     let get_roon_state = move || RoonApi::load_roon_state(&state_path_clone);
 
-    // Start discovery
-    let (mut handles, mut core_rx) = roon
-        .start_discovery(Box::new(get_roon_state), provided, Some(services))
-        .await
-        .ok_or_else(|| anyhow::anyhow!("Failed to start Roon discovery"))?;
-
-    tracing::info!(
-        "Roon discovery started, waiting for core (authorize in Roon → Settings → Extensions)..."
-    );
+    // Start discovery (or, for tests, connect straight to a known Core)
+    let (mut handles, mut core_rx) = match connect {
+        CoreConnect::Discovery => {
+            let started = roon
+                .start_discovery(Box::new(get_roon_state), provided, Some(services))
+                .await
+                .ok_or_else(|| anyhow::anyhow!("Failed to start Roon discovery"))?;
+            tracing::info!(
+                "Roon discovery started, waiting for core (authorize in Roon → Settings → Extensions)..."
+            );
+            started
+        }
+        CoreConnect::Direct { ip, ref port } => {
+            let connected = roon
+                .ws_connect(Box::new(get_roon_state), provided, Some(services), &ip, port)
+                .await
+                .ok_or_else(|| {
+                    anyhow::anyhow!("Failed to connect to Roon Core at {}:{}", ip, port)
+                })?;
+            tracing::info!("Connected directly to Roon Core at {}:{}", ip, port);
+            connected
+        }
+    };
 
     // Event processing task
     let state_for_events = state.clone();
--- a/src/adapters/roon.rs
+++ b/src/adapters/roon.rs
@@ -1588,7 +1588,13 @@ async fn run_roon_loop(
         }
         CoreConnect::Direct { ip, ref port } => {
             let connected = roon
-                .ws_connect(Box::new(get_roon_state), provided, Some(services), &ip, port)
+                .ws_connect(
+                    Box::new(get_roon_state),
+                    provided,
+                    Some(services),
+                    &ip,
+                    port,
+                )
                 .await
                 .ok_or_else(|| {
                     anyhow::anyhow!("Failed to connect to Roon Core at {}:{}", ip, port)
--- a/src/adapters/roon.rs
+++ b/src/adapters/roon.rs
@@ -345,8 +345,9 @@ impl RoonAdapter {
     /// about this file.
     ///
     /// SOOD-based discovery against a fake was rejected deliberately: it needs UDP
-    /// multicast, which is process-global and already the cause of this repo's two
-    /// flaky `/hqp/discover` tests.
+    /// multicast, which is process-global, cannot be scoped to one test, and is why
+    /// this repo's `/hqp/discover` tests already pass in CI while failing on a
+    /// developer machine.
     ///
     /// Returns when the Core is lost or `stop()` is called.
     #[doc(hidden)]
--- a/tests/mock_servers/README.md
+++ b/tests/mock_servers/README.md
@@ -68,7 +68,7 @@ adapter.
 | Load | offset/count paging, short final page, total count vs page size, past-the-end returns empty |
 | play_item | an `item_key` resolves through the item's action list to `Play Now`; the `roon:` prefix is stripped before it reaches the Core |
 | search_and_play | navigates into a search hit to find a playable action; `play` / `queue` invoke different actions; a missing action reports what *is* available |
-| Errors on demand | `InvalidItemKey` and `InvalidLevels`, correlated to the right request id and session key, including with several requests in flight and responses arriving out of order |
+| Errors on demand | `InvalidItemKey` and `InvalidLevels`, correlated to the right request id and session key, including with several requests in flight and responses arriving out of order; and an *unrecognised* error name, to pin what happens when the fork's four literals are wrong |
 | Drift | every request name the adapter sends is a closed set; anything else is recorded as unhandled and answered `InvalidRequest` |
 
 ### Does NOT cover — do not assume otherwise
@@ -127,7 +127,19 @@ that each contains a `Search` item, that an action list contains
 | `action: "none"` in reply to invoking an action item | the adapter discards this BrowseResult | none today; would matter if a caller started reading it |
 | `InvalidItemKey` carries no body | the fork keys only off the name | none for the fork; a real body that parsed as a `BrowseResult` would be read as success |
 | `item_key` format | opacity is all this repo needs | none |
+| **the four browse error *names*** | only evidence is the fork's own literals | **a Core that spells one differently is dropped inside the dependency, and the caller times out as if unreachable — with every test here still green** |
 | **`item_key` portability across `multi_session_key`s** | **this is the epic's open question** | if keys are *not* portable, `/roon/play_item` is broken and #396's ref design changes |
+
+The error-name row is the one #405's PR (#412) explicitly handed to this issue: the
+fork matches `"InvalidItemKey"`, `"InvalidLevels"`, `"UnexpectedError"` and
+`"ZoneNotFound"` against `msg["name"]`, and anything else becomes `Parsed::None`,
+which it drops. This fake sends the fork's own literals, so it can never catch a
+mismatch. What it *can* do — and does — is pin the consequence:
+`an_unrecognised_error_name_degrades_to_an_indistinguishable_timeout` makes the Core
+answer instantly with an unrecognised name and asserts the caller still times out.
+`FakeRoonCore::FORK_ERROR_NAMES` pins the four literals so a fork bump that renames
+one fails here rather than as a mysterious timeout. **Verifying the names needs a
+real Core.**
 
 That last row is the one that matters. `RoonAdapter::play_item` mints a fresh,
 unrelated session key and browses the caller's key inside it, so the repo already
@@ -137,19 +149,60 @@ Core reject a foreign key, so a test can pin either answer. #405 must settle it
 against the operator's rig — see
 `a_foreign_item_key_is_rejected_when_keys_are_session_scoped`.
 
-### Tests that deliberately pin defects
+### Rejection tests work on both sides of #405, by design
 
-Two tests assert today's *wrong* behaviour, and say so at the assertion:
+Four tests drive a Core rejection: a bad item key, a foreign key under
+`ItemKeyScope::PerSession`, a load with no browse, and one rejection among three
+concurrent browses. None of them asserts "it hangs".
 
-* `browse_error_is_delivered_but_the_adapter_drops_it`
-* `concurrent_browses_one_rejected_leaves_the_rejected_caller_hanging`
+`classify_rejection()` in `tests/roon_protocol.rs` enforces what must hold whether
+or not #405 (PR #412) has landed — a rejected request **never** resolves as success,
+and if it resolves at all it resolves promptly, with a message that neither reads as
+a timeout nor omits the browse session — and reports which of the two it observed.
+Verified both ways:
 
-The adapter's event loop discards `Parsed::Error` in its catch-all, so a
-Core-rejected item key resolves nothing and the caller waits out the 10s
-`BROWSE_TIMEOUT`. The fake proves the Core answered immediately — that is what
-`FakeRoonCore::responses()` is for — so the ten seconds are demonstrably UHC's, not
-Roon's. **#405 fixes this. When it lands, invert the assertions; do not delete the
-tests.**
+```
+$ # on v3 (c47d36a), #405 not present
+load without a browse: DroppedByAdapter
+rejected browse: DroppedByAdapter
+rejected browse among three in flight: DroppedByAdapter
+play_item against a foreign key: DroppedByAdapter
+ok. 49 passed
+
+$ # same file, merged with fix/issue-405-roon-browse-errors
+load without a browse: RoutedToCaller
+rejected browse: RoutedToCaller
+rejected browse among three in flight: RoutedToCaller
+play_item against a foreign key: RoutedToCaller
+ok. 49 passed
+```
+
+So this is the end-to-end proof #412 asked for — its own correlation tests are
+unit-scoped precisely because nothing could drive the wire path — and it does not
+have to be inverted on merge. What is pinned unconditionally is the wire half: the
+Core answered, named the error, and correlated it to the rejected request's own
+`req_id` **and** session key, which is exactly the `Parsed::Error` payload.
+
+### One test does pin a defect: same-session cross-delivery
+
+`concurrent_browses_in_one_session_cross_deliver_their_results` asserts a bug is
+**present**, and it survives #405.
+
+Two browses in flight under the *same* `multi_session_key` cross-deliver their
+results. `pending_browses` is keyed by `req_id` but the `Parsed::BrowseResult` arm
+scans by session key — `Parsed::BrowseResult` carries no `req_id`, so it cannot say
+which request in a session answered — and `.find()` returns an arbitrary match.
+Measured: seven trials in eight hand the caller someone else's list, with no error.
+
+`search`, `search_and_play` and `play_item` each mint a private session key and run
+sequentially, so they are safe. `POST /roon/browse` takes a **caller-supplied**
+session key and has no in-repo callers today — but any external client can trigger
+it, and #399's navigation handle will give clients a reusable session identity,
+which is exactly the shape that triggers it. #412's own comment says closing the gap
+needs a change to the pinned fork.
+
+When that assertion starts failing, the bug is fixed: invert it to compare each
+trial against `expected`.
 
 ### Adding to the fake
 
--- a/tests/mock_servers/README.md
+++ b/tests/mock_servers/README.md
@@ -1,0 +1,163 @@
+# Mock backends
+
+Test doubles for the backends UHC talks to. They exist so adapter behaviour can be
+asserted without hardware, a network, or the operator's rig.
+
+**Read the "does not cover" column before you trust one.** A mock nothing asserts
+*against* is worse than no mock, because it reads as coverage: `MockHqpServer`
+rejected its own adapter's wire format until #394 noticed
+(`HqpAdapter::build_command` sends the XML declaration and the command on one line;
+the mock discarded any line starting with `<?xml`).
+
+| Module | Kind | Drives the real adapter? |
+|---|---|---|
+| `roon_core.rs` — `FakeRoonCore` | WebSocket server, MOO protocol | **yes** |
+| `roon.rs` — `MockRoonCore` | in-memory state holder, no protocol | no — and no test uses it |
+| `lms.rs` — `MockLmsServer` | HTTP server, JSON-RPC `slim.request` | yes |
+| `hqplayer.rs` — `MockHqpServer` | TCP server, XML lines | yes |
+| `openhome.rs` — `MockOpenHomeDevice` | HTTP server, SOAP | discovery only |
+| `upnp.rs` — `MockUpnpRenderer` | HTTP server, SOAP | discovery only |
+
+---
+
+## `FakeRoonCore` (issue #408)
+
+A real WebSocket server that speaks Roon's MOO framing and enough of
+`com.roonlabs.registry:1` / `transport:2` / `browse:1` for `RoonAdapter` to run
+against it end to end. Tests live in `tests/roon_protocol.rs`.
+
+`MockRoonCore` in `roon.rs` is **not** this. It is an in-memory zone-state holder
+that nothing speaks a protocol to — which is why, before #408, no test asserted a
+Roon success path. It is left in place unchanged, but note what a grep shows:
+
+```
+$ grep -rn "MockRoonCore" tests/ | grep -v tests/mock_servers/roon.rs
+tests/mock_servers/mod.rs:19:pub use roon::MockRoonCore;
+```
+
+**Its only callers are its own six unit tests.** No adapter test uses it. Prefer
+`FakeRoonCore` for anything new; `MockRoonCore` is a removal candidate, filed
+separately rather than deleted here.
+
+### How a test uses it
+
+```rust
+let core = FakeRoonCore::start().await;              // random loopback port
+let adapter = connected(&core).await;                // real event loop, no SOOD
+let hits = adapter.search("kind of blue", None, Some(10), SearchSource::Library).await?;
+core.assert_no_unhandled_requests().await;           // drift guard
+```
+
+`connected()` calls `RoonAdapter::run_event_loop_against_core_for_tests`, the one
+production-code seam this added: `run_roon_loop` gained a `CoreConnect` parameter so
+tests can connect to a known address instead of doing SOOD multicast discovery.
+Everything after the connection — the event loop, the `pending_browses` /
+`pending_loads` maps, the zone conversion — is the code production runs. That is
+the point. A test that reimplemented the loop would assert nothing about the
+adapter.
+
+### Covered
+
+| Area | What is asserted |
+|---|---|
+| Handshake | `registry:1/info` at request id 0, then `register`; `CoreEvent::Registered`; Browse becomes available |
+| Zones | a `Subscribed` payload survives the fork's deserializer and reaches `get_zones()` — a missing required field would otherwise be swallowed silently |
+| Search | full six-request sequence, one `search_{nanos}` session key throughout, query carried as `input`, source selection (Library / TIDAL / Qobuz), empty result sets, `hierarchy=browse` on every request |
+| **title / subtitle mapping** | asymmetric values at three layers: adapter return value, `GET /roon/search`, `POST /roon/browse`. This is the hole #408 exists to close |
+| Browse | two levels down and `pop_all` back to the root; per-session level stacks proven independent; `input_prompt` advertised; unkeyed rows |
+| Load | offset/count paging, short final page, total count vs page size, past-the-end returns empty |
+| play_item | an `item_key` resolves through the item's action list to `Play Now`; the `roon:` prefix is stripped before it reaches the Core |
+| search_and_play | navigates into a search hit to find a playable action; `play` / `queue` invoke different actions; a missing action reports what *is* available |
+| Errors on demand | `InvalidItemKey` and `InvalidLevels`, correlated to the right request id and session key, including with several requests in flight and responses arriving out of order |
+| Drift | every request name the adapter sends is a closed set; anything else is recorded as unhandled and answered `InvalidRequest` |
+
+### Does NOT cover — do not assume otherwise
+
+* **It cannot prove the adapter matches a real Roon Core.** The fake's semantics
+  came from the pinned fork's deserializers and from this repo's own adapter, so
+  green means *unchanged*, not *correct*. Both were derived from the same source of
+  truth, and that source has never been checked against Roon.
+* **Category-grouped search results.** A real Core returns search hits grouped
+  under `Albums` / `Tracks` / `Artists` rows. The default library returns hits
+  flat, so `is_category`, `try_category_playable` and the second half of
+  `try_navigate_to_playable` (`src/adapters/roon.rs`) are **untested**. A test
+  can build a grouped library itself — `FakeLibrary`'s fields and the `FakeItem`
+  builders are public — but nobody has verified what the real grouping looks like,
+  so the fake does not ship a guess.
+* **Transport control.** `play` / `pause` / `next` / volume / mute go through
+  `com.roonlabs.transport:2/control` and `change_volume`; the fake does not model
+  them. It answers `InvalidRequest` and records the call as unhandled — those
+  adapter methods do not await a reply, so they would appear to succeed; only
+  `assert_no_unhandled_requests()` catches it. Call that assertion in any test you
+  add.
+* **Images.** `com.roonlabs.image:1/get_image` is not modelled.
+* **Queue.** `subscribe_queue` / `play_from_here` (which #400 needs) are not
+  modelled.
+* **Zone updates over time.** Zones are sent once at subscribe. No `Changed`,
+  `zones_added`, `zones_removed` or `zones_seek_changed` events.
+* **Reconnection and Core loss.** Dropping the connection is not exercised.
+* **Roon's real error bodies.** The fake sends error names with no body.
+* **Authorization.** A real extension must be authorized in Roon → Settings →
+  Extensions before `register` succeeds. The fake always registers, so the
+  unauthorized state is untested.
+
+### Which shapes are inferred rather than recorded
+
+**No live Roon Core was reachable when this was written** — this machine's config
+directory has no `roon_state.json`, so there is no pairing token, and obtaining one
+requires a human to authorize the extension in Roon's UI. So nothing here is
+recorded from a Core. Two pedigrees, unequal confidence:
+
+*From the pinned fork* (`~/.cargo/git/checkouts/rust-roon-api-*/06dd807`) — which
+fields exist, which are required, enum spellings, the four browse error names, the
+handshake order. High confidence: a shape the fork accepts is a shape the adapter
+can consume.
+
+*From this repo's adapter* — that the root contains `Library` / `TIDAL` / `Qobuz`,
+that each contains a `Search` item, that an action list contains
+`Play Now` / `Queue` / `Start Radio`. High confidence as *expectations*:
+`src/adapters/roon.rs` will not work against anything else.
+
+*Inferred, unverified* — each marked `INFERRED:` at its use site in `roon_core.rs`:
+
+| Shape | Why it is a guess | Consequence if wrong |
+|---|---|---|
+| root list title `"Explore"` | never read by this repo | none — cosmetic |
+| `list.level` numbered from 0 at the root | plausible, unchecked | #399's "report your position" would be off by one |
+| `action: "none"` in reply to invoking an action item | the adapter discards this BrowseResult | none today; would matter if a caller started reading it |
+| `InvalidItemKey` carries no body | the fork keys only off the name | none for the fork; a real body that parsed as a `BrowseResult` would be read as success |
+| `item_key` format | opacity is all this repo needs | none |
+| **`item_key` portability across `multi_session_key`s** | **this is the epic's open question** | if keys are *not* portable, `/roon/play_item` is broken and #396's ref design changes |
+
+That last row is the one that matters. `RoonAdapter::play_item` mints a fresh,
+unrelated session key and browses the caller's key inside it, so the repo already
+assumes keys are global. The fake refuses to decide: `ItemKeyScope::Global` is the
+default (matching the repo's assumption) and `ItemKeyScope::PerSession` makes the
+Core reject a foreign key, so a test can pin either answer. #405 must settle it
+against the operator's rig — see
+`a_foreign_item_key_is_rejected_when_keys_are_session_scoped`.
+
+### Tests that deliberately pin defects
+
+Two tests assert today's *wrong* behaviour, and say so at the assertion:
+
+* `browse_error_is_delivered_but_the_adapter_drops_it`
+* `concurrent_browses_one_rejected_leaves_the_rejected_caller_hanging`
+
+The adapter's event loop discards `Parsed::Error` in its catch-all, so a
+Core-rejected item key resolves nothing and the caller waits out the 10s
+`BROWSE_TIMEOUT`. The fake proves the Core answered immediately — that is what
+`FakeRoonCore::responses()` is for — so the ten seconds are demonstrably UHC's, not
+Roon's. **#405 fixes this. When it lands, invert the assertions; do not delete the
+tests.**
+
+### Adding to the fake
+
+* Do not encode a Roon semantic this repo does not already depend on. If a shape
+  is uncertain, make it configurable and mark it `INFERRED:`, the way
+  `ItemKeyScope` is.
+* Library *shape* is data a test supplies (`FakeLibrary`, `FakeItem`); protocol
+  behaviour is the fake's. Keep that line.
+* If you teach it a new request, `the_fake_models_everything_the_adapter_sends`
+  pins the closed set of request names — update it deliberately, and never by
+  deleting the assertion.
--- a/tests/mock_servers/README.md
+++ b/tests/mock_servers/README.md
@@ -62,7 +62,7 @@ adapter.
 |---|---|
 | Handshake | `registry:1/info` at request id 0, then `register`; `CoreEvent::Registered`; Browse becomes available |
 | Zones | a `Subscribed` payload survives the fork's deserializer and reaches `get_zones()` — a missing required field would otherwise be swallowed silently |
-| Search | full six-request sequence, one `search_{nanos}` session key throughout, query carried as `input`, source selection (Library / TIDAL / Qobuz), empty result sets, `hierarchy=browse` on every request |
+| Search | full six-request sequence, one `search_{nanos}` session key throughout, query carried as `input`, source selection (Library / TIDAL / Qobuz), empty result sets, `hierarchy=browse` on every request. **Flat results only** — see the category-grouping entry below — and against a hierarchy taken from the adapter's own expectations, so this cannot detect the adapter expecting the wrong hierarchy |
 | **title / subtitle mapping** | asymmetric values at three layers: adapter return value, `GET /roon/search`, `POST /roon/browse`. This is the hole #408 exists to close |
 | Browse | two levels down and `pop_all` back to the root; per-session level stacks proven independent; `input_prompt` advertised; unkeyed rows |
 | Load | offset/count paging, short final page, total count vs page size, past-the-end returns empty |
@@ -77,13 +77,19 @@ adapter.
   came from the pinned fork's deserializers and from this repo's own adapter, so
   green means *unchanged*, not *correct*. Both were derived from the same source of
   truth, and that source has never been checked against Roon.
-* **Category-grouped search results.** A real Core returns search hits grouped
-  under `Albums` / `Tracks` / `Artists` rows. The default library returns hits
-  flat, so `is_category`, `try_category_playable` and the second half of
-  `try_navigate_to_playable` (`src/adapters/roon.rs`) are **untested**. A test
-  can build a grouped library itself — `FakeLibrary`'s fields and the `FakeItem`
+* **Category-grouped search results — probably the most-travelled real path.** A
+  real Core returns search hits grouped under `Albums` / `Tracks` / `Artists` rows;
+  `is_category` and `try_category_playable` exist in `src/adapters/roon.rs` because
+  someone saw that. The default library returns hits flat, so those two and the
+  second half of `try_navigate_to_playable` are **untested**, which means the
+  branch a real search most likely takes is the one with no coverage. A test can
+  build a grouped library itself — `FakeLibrary`'s fields and the `FakeItem`
   builders are public — but nobody has verified what the real grouping looks like,
-  so the fake does not ship a guess.
+  so the fake ships no guess.
+* **The hierarchy itself.** `Library` / `TIDAL` / `Qobuz` → `Search` → results is
+  taken from what `src/adapters/roon.rs` requires, so a fake built from it can never
+  disagree with the adapter about navigation. If Roon renamed `Search`, `search()`
+  would break in production and this suite would stay green.
 * **Transport control.** `play` / `pause` / `next` / volume / mute go through
   `com.roonlabs.transport:2/control` and `change_volume`; the fake does not model
   them. It answers `InvalidRequest` and records the call as unhandled — those
@@ -203,6 +209,21 @@ needs a change to the pinned fork.
 
 When that assertion starts failing, the bug is fixed: invert it to compare each
 trial against `expected`.
+
+### Two scheduled edits — do these, do not skip them
+
+1. **When #405 (PR #412) is on `v3`:** tighten `classify_rejection` so
+   `RejectionOutcome::DroppedByAdapter` is a hard failure, and replace the string
+   checks with `RoonBrowseError::from_error(&e)` asserting
+   `kind == RoonBrowseErrorKind::InvalidItemKey` and the session key. As written the
+   helper accepts both outcomes, which means it cannot catch a regression *back* to
+   dropping the error — the exact bug #405 fixed.
+2. **`concurrent_browses_in_one_session_cross_deliver_their_results` will rot.** It
+   asserts a defect is present, so it turns red when the defect is fixed, and the
+   person who sees that red will not know it is good news. The instruction is at the
+   assertion site: invert it, compare each trial against `expected`. It is also the
+   only probabilistic assertion in this repo's suite (eight trials, roughly 1 in
+   10^7 for a false failure).
 
 ### Adding to the fake
 
--- a/tests/mock_servers/README.md
+++ b/tests/mock_servers/README.md
@@ -109,43 +109,98 @@ adapter.
 
 ### Which shapes are inferred rather than recorded
 
-**No live Roon Core was reachable when this was written** — this machine's config
-directory has no `roon_state.json`, so there is no pairing token, and obtaining one
-requires a human to authorize the extension in Roon's UI. So nothing here is
-recorded from a Core. Two pedigrees, unequal confidence:
+**No live Roon Core was reachable** — this machine's config directory has no
+`roon_state.json`, so there is no pairing token, and obtaining one requires a human
+to authorize the extension in Roon's UI. So nothing here was recorded off a Core *by
+this work*.
 
-*From the pinned fork* (`~/.cargo/git/checkouts/rust-roon-api-*/06dd807`) — which
-fields exist, which are required, enum spellings, the four browse error names, the
-handshake order. High confidence: a shape the fork accepts is a shape the adapter
-can consume.
+The alternative to recording is not hand-writing shapes and letting the fake agree
+with its author — that is exactly the failure #408 exists to end, and #407 found a
+real client-visible bug precisely because it recorded rather than assumed. So the
+uncertain shapes were chased into published sources instead. **Five pedigrees,
+unequal confidence:**
 
-*From this repo's adapter* — that the root contains `Library` / `TIDAL` / `Qobuz`,
-that each contains a `Search` item, that an action list contains
-`Play Now` / `Queue` / `Start Radio`. High confidence as *expectations*:
-`src/adapters/roon.rs` will not work against anything else.
+**1. From the pinned fork** (`~/.cargo/git/checkouts/rust-roon-api-*/06dd807`) —
+which fields exist, which are required, the handshake order, the 10s read timeout. A
+shape the fork's `serde` accepts is a shape the adapter can consume.
 
-*Inferred, unverified* — each marked `INFERRED:` at its use site in `roon_core.rs`:
+**2. From RoonLabs' own published API** (`RoonLabs/node-roon-api-browse`,
+`RoonLabs/node-roon-api`) — quoted at the use sites:
 
-| Shape | Why it is a guess | Consequence if wrong |
-|---|---|---|
-| root list title `"Explore"` | never read by this repo | none — cosmetic |
-| `list.level` numbered from 0 at the root | plausible, unchecked | #399's "report your position" would be off by one |
-| `action: "none"` in reply to invoking an action item | the adapter discards this BrowseResult | none today; would matter if a caller started reading it |
-| `InvalidItemKey` carries no body | the fork keys only off the name | none for the fork; a real body that parsed as a `BrowseResult` would be read as success |
-| `item_key` format | opacity is all this repo needs | none |
-| **the four browse error *names*** | only evidence is the fork's own literals | **a Core that spells one differently is dropped inside the dependency, and the caller times out as if unreachable — with every test here still green** |
-| **`item_key` portability across `multi_session_key`s** | **this is the epic's open question** | if keys are *not* portable, `/roon/play_item` is broken and #396's ref design changes |
+| Shape | RoonLabs' words |
+|---|---|
+| `list.level` | "increases from 0" — so root = 0 is **documented**, not inferred |
+| `action` values | `message`, `none` ("No action is required"), `list`, `replace_item`, `remove_item` |
+| `hint` values | `null` ("Unknown"), `action`, `action_list`, `list`, `header` |
+| `input_prompt` | `prompt`, `action` ("The verb that goes with this action"), `value`, `is_password` |
+| `InvalidRequest` | what `node-roon-api` itself sends for "unknown service" / "unknown request name" — the shape this fake uses for unmodelled calls |
 
-The error-name row is the one #405's PR (#412) explicitly handed to this issue: the
-fork matches `"InvalidItemKey"`, `"InvalidLevels"`, `"UnexpectedError"` and
-`"ZoneNotFound"` against `msg["name"]`, and anything else becomes `Parsed::None`,
-which it drops. This fake sends the fork's own literals, so it can never catch a
-mismatch. What it *can* do — and does — is pin the consequence:
+**3. Recorded off real Cores, by third parties** — real Cores, but other people's
+logs rather than a controlled run:
+
+* `MOO/1 COMPLETE InvalidItemKey`, verbatim, in `home-assistant/core#137605`:
+  `Could not play id:122:4, result: MOO/1 COMPLETE InvalidItemKey` (also `130:12`,
+  `133:13`, `109:7`, `135:5`). Confirms the name **and** that the reply is just verb
+  and name.
+* Real `item_key`s are two colon-separated integers and "change with each refresh" —
+  `"115:0"` becoming `"116:0"` (Roon Labs community thread 23129).
+
+**4. From this repo's adapter** — the `Library`/`TIDAL`/`Qobuz` → `Search` hierarchy
+and the `Play Now`/`Queue`/`Start Radio` action titles. High confidence as
+*expectations*; `src/adapters/roon.rs` will not work against anything else. **That is
+not the same as Roon being that way.**
+
+**5. Inferred and unverified** — each marked `INFERRED:` at its use site:
+
+| Shape | Consequence if wrong |
+|---|---|
+| root list title `"Explore"` | none — never read |
+| a real Core sends *no body* with an error | none for the fork; a real body that parsed as a `BrowseResult` would read as success |
+| `action: "none"` is the reply a Core picks for an invoked action | none today — the adapter discards that `BrowseResult` |
+| **`InvalidLevels` / `UnexpectedError` / `ZoneNotFound` spellings** | **dropped inside the dependency; the caller times out as if the Core were unreachable, with every test here still green** |
+| search results are flat rather than category-grouped | `is_category` / `try_category_playable` untested |
+| **`item_key` portability across `multi_session_key`s** | if keys are *not* portable, `/roon/play_item` is broken and #396's ref design changes |
+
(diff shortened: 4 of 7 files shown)


muness:
# Gate 1 — Solution space · REVIEW

**Gate/role:** solution-space gate, reviewer
**Issue:** #408 · **PR:** #413 · **Base:** `v3`

## Aim and scope checked

Aim: make Roon's wire behaviour assertable against something that answers like Roon, so #396/#399/#405 stop choosing between unit-scoping and asserting nothing.

In scope: a test-only fake, sufficient for search (incl. the `multi_session_key` mint/consume sequence), browse+load across levels with `pop_all`, `play_item`, and on-demand `Parsed::Error(BrowseInvalidItemKey)`. Out of scope: reimplementing Roon Core, CI against a live Core, changing adapter behaviour.

## Evidence inspected

- `src/adapters/roon.rs` — `search()` `:675-780`, `browse()`/`load()` `:580-675`, `play_item()` `:1094-1160`, `execute_play_action` `:1223-1300`, event loop and pending maps `:1860-1915`, `run_roon_loop` `:1464-1540`.
- Pinned fork `06dd807`: `moo.rs` (framing, the 10s read timeout at `:245-249`), `lib.rs` (`ws_connect` `:170-200`, `start_discovery` `:204-227`, registration `:405-524`, the service dispatch chain `:597-645`), `browse.rs` (`BrowseOpts`/`LoadOpts`/`BrowseResult`/`LoadResult`, `parse_msg` `:150-188`), `transport.rs` (`Zone` required fields `:94-131`, `parse_msg` `:451-541`), `status.rs`.
- `tests/mock_servers/*` — all five existing mocks, for shape precedent.
- PR #404's body and its three gate pairs.
- `~/Library/Application Support/unified-hifi-control/unified-hifi/` — no `roon_state.json`.

## Approaches compared

**A. Extend `MockRoonCore` and stand up an event loop inside the test.**
No production change. Rejected: the test would have to re-implement the `Parsed::` dispatch and the `pending_browses`/`pending_loads` correlation, which is the code #405 changes and the code most worth covering. A fake whose tests exercise a *copy* of the logic asserts nothing about production — the exact "looks like coverage" failure this issue was raised to end.

**B. Record a real Core's traffic and replay a fixed transcript.**
Highest fidelity in principle. Rejected as the primary mechanism on two grounds: (i) `req_id`s and `multi_session_key`s vary per run, so a transcript needs a matcher, at which point it is a fake with worse ergonomics; (ii) it cannot emit an error correlated to an arbitrary in-flight request, which is the #405 requirement. Also blocked in practice — no Core reachable.

**C. Protocol fake over WebSocket + a minimal `ws_connect` seam.** Selected.
`RoonApi::ws_connect(.., ip, port)` is already `pub` in the fork, so the seam is a parameter, not new transport. Tests then drive the real loop.

**D. Fake SOOD discovery so no production change is needed.**
Rejected: UDP multicast is process-global, cannot be scoped to one test, and is already the cause of this repo's two flaky `/hqp/discover` failures. Trading a 20-line reviewable seam for the repo's known worst flakiness source is a bad trade.

Selected: **C**, with B folded in where it can be: response *shapes* taken from the fork's deserializers rather than invented, and library *structure* taken from the adapter's own expectations.

## Findings

**P1 — The recording criterion cannot be met, and that must be stated, not finessed.**
#408 says "where a response shape was uncertain, it is recorded from a real Core and the provenance noted — not guessed." No Core is reachable: no pairing token exists, and pairing requires a human in Roon's UI. Any shape not derivable from the fork or the adapter is therefore a guess. **Required:** enumerate the guesses explicitly, mark each at its use site, and state the consequence if each is wrong. Do not let plausible-sounding recall pass as fact.

**P2 — The fake must not become a second implementation of Roon's semantics.**
This is #408's named constraint and the main design risk. **Required:** the fake owns *protocol behaviour* only; library shape is data a test supplies. Where the design would otherwise need a semantic decision (notably `item_key` scoping), make it configurable rather than choosing.

**P2 — "It emits the error" is not the same as "the error is asserted."**
The adapter currently discards `Parsed::Error`, so a test that only checks a return value will show a 10s timeout and cannot distinguish "the Core never answered" from "the Core answered and we dropped it". **Required:** the fake records what it *sent*, not only what it received.

**P3 — Test runtime.** `BROWSE_TIMEOUT` is 10s and not configurable. A test that waits it out costs 10s. **Required:** assert "still pending after a short bound" instead, which proves the same thing in under a second.

**P3 — The `title`/`subtitle` hole lives in four places, not one.** `src/api/mod.rs:402` and `:681`, `src/mcp/mod.rs:483`, plus the adapter's navigation. The issue's AC names the search mapping; covering only the adapter layer would satisfy the letter and miss the defect. **Required:** cover every layer reachable without colliding with #394, and name the one that is not.

**P4 — A literal `title: item.subtitle` swap does not type-check** (`String` vs `Option<String>`). The demonstration must use the swap a careless refactor would actually produce.

## Required actions

1. Provenance table: FROM FORK / from this repo's adapter / INFERRED, marked at each use site, with consequences.
2. Protocol in the fake, library shape in the test.
3. `ItemKeyScope` configurable; default matches what the repo already assumes.
4. Record outbound responses as well as inbound requests.
5. Bounded-wait assertions instead of 10s waits.
6. Cover the two HTTP projections; state the MCP one as #394's.
7. Production seam confined to `run_roon_loop`'s connect step; nothing in the `Parsed::` arms (#405's territory).

## Disposition

Approach C authorized with the seven actions above.

## Residual risk

The fake's semantics are derived from the same repo as the adapter, so it can only prove self-consistency. That ceiling is inherent to building this without a Core and must be documented rather than engineered away.

## Next gate

Execute gate authorized.

muness:
# Gate 1 — Solution space · DISSENT

**Gate/role:** solution-space gate, dissent
**Verdict:** block (two blocks; both must be resolved before execute)
**Issue:** #408 · **PR:** #413

The thing that makes mocks dangerous is not that they are wrong. It is that they are *reassuring*. My job at this gate is to ask whether the selected approach produces assertions about real behaviour or a well-built simulator that agrees with itself. The precedent is not hypothetical: `MockHqpServer` rejected `HqpAdapter`'s own wire format for months and nobody noticed, because nothing asserted against it and it read as coverage.

## The contrary case for doing nothing

Worth stating because it is not weak. `#405` was told to unit-scope its tests and not build a mock, and #396 can unit-scope its ref table. If both can proceed without this, the fake's value is *confidence*, and confidence built on a fake derived from the same repo as the adapter is the most expensive kind of wrong. A cheaper alternative exists: extract the `Parsed::` dispatch into a pure function and unit-test it with hand-built values. Twenty lines, no seam, no dev-dependency, no WebSocket server.

Why I do not press it to a block: that alternative cannot cover `search()`'s six-request sequence, cannot show that `pop_all` resets a level stack, and cannot show `play_item` walking to an action. Those are integration facts about a stateful protocol, and the epic depends on all three. But it does mean the fake must earn its keep on *assertions*, not on existence. Which produces my first block.

## BLOCK 1 — As specified, this fake would pass its own tests while the adapter is broken

Concretely, here are three real defects the design as described at the gate would not catch:

1. **The adapter sends the query in the wrong field.** If `search()` put the query in `input` on the *source* item rather than the Search item, a fake that returns matches for any input-carrying browse would answer happily. Nothing would fail.
2. **The adapter forgets to strip the `roon:` prefix from a zone id.** The fake ignores `zone_or_output_id`. Green.
3. **The adapter stops sending `pop_all` and starts sending `pop_levels: 99`.** Both reset the fake to the root. Green.

Each is silent because the assertions all live on the *return* path. The fix is structural, not extra tests: the fake must record every inbound request and every outbound response, and the tests must assert on the recorded requests. #394 hit exactly this and fixed it by adding a command log to `MockLmsServer`; the same fix, or this fake is decoration.

**Also required, same block:** the fake must answer `InvalidRequest` to anything it does not model *and* record it as unhandled, with an `assert_no_unhandled_requests()` that tests call. Otherwise an adapter that grows a new Roon call hangs for ten seconds and someone "fixes" it by raising a timeout. The set of request names the adapter sends must be pinned as a closed set, so growth is a visible edit.

**Unblocks when:** request log, response log, unhandled-request assertion, and closed request-name set all exist and are used by tests.

## BLOCK 2 — The title/subtitle demonstration as scoped proves the least interesting layer

#408's AC says "a regression test that would have caught a swapped `title`/`subtitle` in Roon's search mapping." Read carelessly, an adapter-level test satisfies it. It should not, because **there is no `title`/`subtitle` mapping in the adapter** — the wire-to-struct step is `serde` inside the pinned fork. A test asserting that `serde` puts `title` into `title` is asserting that a third-party derive macro works.

The actual mapping #394 was talking about is in `src/mcp/mod.rs:483`, and there are two more of the same shape in `src/api/mod.rs` (`:402`, `:681`). Shipping an adapter-level assertion and calling the hole closed would be the paper-over version of this issue.

I accept that `src/mcp/` is #394's and that duplicating its harness is worse than leaving a stated gap. But then:

**Unblocks when:** (a) both `src/api/mod.rs` projections are covered end to end through their HTTP handlers, (b) each is demonstrated failing when swapped — with pasted output, not a claim — and (c) the MCP one is named as uncovered, with the specific reason and the specific way #394 closes it. If only the adapter layer is covered, this issue has not done its job regardless of how many tests it ships.

## Concerns short of a block

**The seam is a permanent `pub` hole.** `#[doc(hidden)]` is documentation, not enforcement. I considered demanding a cargo feature gate and withdraw it: a gated seam is invisible to `cargo test` and to `cargo clippy -- -D warnings`, which trades a small API-surface risk for a real loss of lint coverage on the seam itself. The correct fix is an `architecture_lint` rule asserting no non-test callers. **Not a block** — it is new lint infrastructure, and #408 is not a lint issue. Record it as residual risk and as a follow-up, and do not pretend `#[doc(hidden)]` is a control.

**`ItemKeyScope` could become a way to dodge the empirical question.** Making the unknown configurable is right. But a configurable unknown with a default is still a default, and defaults get quoted as behaviour. Acceptable only if the default is documented as *matching what `play_item` already assumes* rather than as *what Roon does*, and if a test exercises the other setting so the alternative is not theoretical.

**The keepalive ping hides a real failure mode.** The fork drops the connection after 10s of silence and reports a lost Core. A fake that pings every 3s means no test will ever see that. Right call for test stability, but it must be in the not-covered list: reconnection and Core-loss handling stay untested, and they are not hypothetical — `restart_needed` and the `CoreEvent::Lost` path are live production code with no coverage at all.

**"1337 lines of test infrastructure" is itself a liability.** It is roughly 65% of the adapter it tests. Every line is a line that can be wrong in the direction of the adapter's own bugs. The mitigation is that no semantic decision may be hardcoded where it could be data — which is why I support the review's action 2 and would escalate if the library shape ended up baked into the protocol code.

## Contrary evidence I looked for and did not find

- **Any recorded Roon browse payload in this repo or on this machine.** Searched the repo (`item_key` appears only in `AGENTS.md`, `src/adapters/roon.rs`, `src/api/mod.rs`), the legacy `mcp/index.js` (no browse code), and the Node reference path in `AGENTS.md` (`/Users/muness1/...` does not exist). Nothing. So "recorded from a real Core" is not merely inconvenient here — it is unavailable, and the honest output is a labelled guess.
- **A precedent in this repo for a feature-gated test seam.** None. So the unconditional-`pub` choice is not a departure from an established pattern; it is the first of its kind, which is why it needs saying out loud in the PR rather than burying in a doc comment.

## Required actions

1. Request log, response log, unhandled-request assertion, closed request-name set. (Block 1)
2. Both `src/api/mod.rs` projections covered *and demonstrated biting with pasted output*; MCP named as the remaining gap with its reason. (Block 2)
3. `ItemKeyScope` default documented as the repo's assumption, and the other setting exercised by a test.
4. Not-covered list must include reconnection/Core-loss, and say the keepalive is why.
5. Seam risk recorded as residual with the lint follow-up named; `#[doc(hidden)]` not presented as enforcement.

## Disposition

Blocked pending 1 and 2. Approach C is otherwise the right choice and A/B/D are correctly rejected.

## Next gate

Not authorized until Blocks 1 and 2 clear.

muness:
# Gate 2 — Execute · REVIEW

**Gate/role:** execute gate, reviewer
**Issue:** #408 · **PR:** #413 · **Commits:** `e604c0c`, `707833e`, `19bed47`, `a64b745`

## Aim and scope checked

Build the instrument, do not fix what it finds. Support search including the session-key mint/consume sequence, browse+load across levels with `pop_all`, `play_item`, and on-demand `Parsed::Error(BrowseInvalidItemKey)`. Prove the `title`/`subtitle` hole is closed. No production change beyond what testability requires.

## Evidence inspected

- `tests/mock_servers/roon_core.rs` — the framing codec against `moo.rs`, the handshake against `lib.rs:405-524`, the zone payload against `transport.rs:94-131`, the browse/load bodies against `browse.rs:57-98`, the four error names against `browse.rs:169-183`.
- `tests/roon_protocol.rs` — 25 tests.
- `src/adapters/roon.rs` diff: +79/−9, three hunks, none in the `Parsed::` arms.
- PR #412 (#405) in full, including its module comment assigning the wire-level error-name question here.
- Local runs pasted in the PR body; CI on #413.

## Gate-1 required actions — disposition

| # | Action | Status |
|---|---|---|
| 1 | Provenance table, marked at use sites, with consequences | done — module header + README, `FROM FORK` / `INFERRED` at every site |
| 2 | Protocol in the fake, library shape in the test | done — `FakeLibrary`/`FakeItem` public; `a_custom_library_is_browsable` proves a test can supply its own |
| 3 | `ItemKeyScope` configurable, default = the repo's assumption | done, and both settings exercised |
| 4 | Record outbound responses | done — `responses()`, and `errors_sent()` records the `(req_id, name, session_key)` triple |
| 5 | Bounded waits, not 10s | done — 750 ms bound; whole suite runs in 1.7 s |
| 6 | Cover both HTTP projections; name the MCP one | done, and both demonstrated biting with pasted output |
| 7 | Seam confined to the connect step | done |

Gate-1 dissent Blocks 1 and 2: both cleared. Block 1 by the request log, response log, `assert_no_unhandled_requests()` and the closed request-name set. Block 2 by `search_results_survive_the_http_boundary` and `browse_items_survive_the_http_boundary`, each shown failing under a swap.

## Findings

**P1 — The rejection tests were pointed at the wrong target, and this is now fixed.**
The first implementation pinned the pre-#405 hang. #405 has since landed as PR #412 and is green against `v3`; those assertions would have turned red on merge — a test whose job is to prove a fix cannot be built to fail when the fix arrives. `a64b745` replaces them with `classify_rejection`, which enforces the base-independent invariants and reports which adapter it saw. **Verified both ways, same file:** `DroppedByAdapter` ×4 on `v3`, `RoutedToCaller` ×4 merged with #412, 49 passed each time, with `cargo test --lib`, clippy and fmt clean on the merged state. This is the strongest evidence in the PR and it should not be diluted.

**P1 — `cargo fmt --check` was missing from the local gate and CI's Lint job caught it.**
Clippy does not catch formatting. First push failed Lint on fmt alone. Both now run before every push; Lint is green on `a64b745`.

**P2 — A new defect surfaced and is correctly not fixed here.**
Concurrent browses in one `multi_session_key` cross-deliver results, 7 trials in 8, silently, and it survives #405. Measured, not inferred. Correctly pinned as present rather than fixed, and flagged as needing its own issue before #399 designs a reusable navigation handle on top of it. This is the instrument doing precisely what it was built for.

**P2 — The error-name gap is handled as well as it can be from here.**
#412 assigned it to #408; no Core is reachable, so it cannot be verified. Pinning the *consequence* — `an_unrecognised_error_name_degrades_to_an_indistinguishable_timeout`, plus `FORK_ERROR_NAMES` — converts an unverifiable claim into a test. That is the right substitution and it is labelled as a substitution, not a verification.

**P3 — Three robustness fixes found by inspection, all real.**
Per-instance `core_id` plus a retrying read of the shared pairing-state file (every test in the binary rewrites it, so a read could catch a truncate); `app_state_with` made `async` instead of calling `futures::executor::block_on` inside the tokio runtime.

**P3 — `play_item`'s message names the action container** (`"Play Now: Play Album"`). Recorded as behaviour, filed as a finding, not fixed. Correct call.

**P4 — `search_source_selects_a_different_branch` indexes `browses[1]` and `browses[4]`.** Positional and brittle if the request sequence changes. Acceptable: a change in that sequence *should* fail a test, and `search_mints_one_session_key_and_uses_it_throughout` pins the sequence explicitly.

## Required actions

1. Re-verify against #412 if that PR's head moves before merge. The verification is a point-in-time claim about commit `96d3ebd`.
2. Ship gate must state the exact tightening #412's merge enables (`RoonBrowseError::from_error`) so it is not lost.
3. The cross-delivery defect needs an issue; it is out of scope here but must not evaporate into a PR paragraph.

## Disposition

Pass with the three actions. The instrument now asserts real behaviour at every layer it claims to cover, and its limits are stated where a reader will hit them.

## Residual risk

The fake agrees with the fork by construction. Every unverifiable question is either configurable or has its consequence pinned, which is the most that can be done without the operator's Core — and it means no green run here should ever be quoted as evidence about Roon itself.

## Next gate

Ship gate authorized.

muness:
# Gate 2 — Execute · DISSENT

**Gate/role:** execute gate, dissent
**Verdict:** block (one block), then pass-with-actions on the rest
**Issue:** #408 · **PR:** #413

Gate 1's question was whether the design could produce honest assertions. This gate's question is narrower and harsher: **now that the fake exists and everything is green, what is it lying about?**

## BLOCK — Two claims in this PR were asserted before they were checked

Both were wrong in the same direction: assuming, then presenting the assumption as fact. That is the exact failure this issue exists to end, committed in the PR body of the issue that exists to end it.

1. **"Feature-branch PRs run zero CI."** Inherited from #404's body, repeated without checking the workflow triggers. The triggers key on the PR's *base*. This PR targets `v3`. It gets the full gate — and the first push **failed Lint** on `cargo fmt --check`, a gate that was never run locally because the brief named clippy and I did not go read the workflow.

2. **"Two pre-existing test failures."** The `/hqp/discover` tests fail on a dev Mac because they do live UDP SSDP. CI passes them. "Pre-existing failure" and "fails in my environment" are different claims, and writing the second as the first is how a real regression eventually gets waved through.

Neither was a code defect, which is why it would be easy to file them as cosmetic. I do not accept that. A PR whose central argument is *provenance discipline* cannot carry two unverified inherited claims in its own evidence section. **Unblocks when** both are corrected in the body **with the reasoning shown**, not silently edited — a reader needs to see that the earlier claim was wrong and why, or the correction teaches nothing.

*Resolved:* the body now opens with "CI is a real gate on this PR" and explains the base-vs-head confusion and the fmt miss; the `/hqp/discover` entry is relabelled a local-environment artifact with CI's contrary result cited. Block cleared.

## The near-miss that concerns me most

The rejection tests originally pinned the pre-#405 hang. That was internally consistent, well documented, and **would have broken `v3` on #412's merge** — a test whose stated purpose is to be the end-to-end proof of a fix, built so that the fix makes it red. It only got caught because the coordinator said #405 had landed.

That is not a lucky catch to be relieved about. It is a warning about the whole genre: *defect-pinning tests rot silently, and the person who wrote them is the least likely to notice.* The fix — `classify_rejection`, correct on both bases, verified on both bases — is right. But note the one remaining defect-pinning test (`concurrent_browses_in_one_session_cross_deliver_their_results`) has the identical failure mode, and its inversion instruction lives in a doc comment. That is the best available answer and it is not a good one.

**Required:** the ship gate must name this test explicitly as a rot risk, not bury it in a residual-risk list.

## Where the fake is still agreeing with itself

Not blocks, because each is documented — but each should be read as a limit, not a caveat.

**The four error names.** The fake sends the fork's literals, so `a_rejected_item_key_is_answered_immediately_and_correlated` proves the adapter handles the names *the fork already expects*. If Roon says `"InvalidItemKeyError"`, every rejection test here stays green and every real caller times out. `an_unrecognised_error_name_degrades_to_an_indistinguishable_timeout` is a good substitution — it makes the failure mode a test rather than a sentence — but substitution is not verification, and #412's comment assigned this issue the verification. **#408 does not close that; say so plainly and stop there.**

**The library structure.** `Library` / `TIDAL` / `Qobuz` → `Search` → results is the adapter's expectation, so a fake built from it can never disagree with the adapter about navigation. If Roon renamed `Search`, `search()` breaks in production and this suite stays green. Unfixable without a Core. Must not be described as "search is covered" without that qualifier.

**Category-grouped results.** `is_category` and `try_category_playable` exist because someone saw a real Core group search hits under `Albums`/`Tracks`. The fake returns hits flat, so those paths are dead in tests — and a real Core probably exercises them on *every* search. So the most-travelled real-world search path is the one with no coverage. Correctly listed as not-covered; worth restating that "search is covered" means the flat shape only.

## Contrary evidence I went looking for

- **Does the same-session cross-delivery finding hold up, or is it an artifact of the fake pipelining?** It is not an artifact. A real Core pipelines too, the delays only control arrival order, and the defect is in `.find()` over a map keyed by something other than what the response identifies. Verified it survives #412, whose own comment predicts exactly this gap and says closing it needs a fork change. The finding is real and it is additive to #412.
- **Is the seam actually exercising the production loop, or a lookalike?** Checked: `run_event_loop_against_core_for_tests` calls the same `run_roon_loop` with the same state, bus and knob store; only the connect step differs. The `Parsed::` arms, the pending maps and `convert_zone` are all the production ones. Confirmed by the #412 experiment — the tests changed behaviour purely because the adapter changed, which is only possible if they are driving it.
- **Can the 46→49 test count be padding?** Of the three added, one is the cross-delivery finding, one is the unrecognised-error-name consequence, one pins the fork's four literals. None is decorative.
- **Does `assert_no_unhandled_requests()` get called where it matters?** Called in 6 tests including the drift guard. Not called in the paging or HTTP tests, where it would add nothing. Acceptable, but note the README's claim that transport control "fails loudly" is only true in tests that call it — which the README now says.

## Concerns short of a block

**`classify_rejection` accepts two outcomes, and that is a real cost.** It cannot catch a regression *from* `RoutedToCaller` *back to* `DroppedByAdapter`. Once #412 is on `v3`, `DroppedByAdapter` must become a hard failure. If that tightening is forgotten, this suite silently tolerates the reintroduction of the exact bug #405 fixed. **Required:** ship gate states it as a scheduled action with the precise edit, not as a suggestion.

**1337 lines of fake for a 2055-line adapter.** I raised this at gate 1 and it has grown. Mitigated as agreed — no semantic decision is hardcoded where it could be data — but the ratio is the honest measure of how much scaffolding "assert one Roon success path" actually costs, and the PR should not present it as free.

**The probabilistic assertion.** Eight trials at ~1/8 each gives about 1 in 10^7 for a false failure. Fine. But it is the only non-deterministic assertion in the repo's test suite, and the next person to see it fail will not know that. It says so at the assertion site; that has to be enough.

## Required actions

1. Correct both inherited claims in the body, showing the reasoning. (Block — resolved)
2. Ship gate names `concurrent_browses_in_one_session_cross_deliver_their_results` as a rot risk and `classify_rejection`'s tightening as a scheduled edit with the exact change.
3. Do not describe search as "covered" without the flat-results and structure-from-the-adapter qualifiers.
4. State that #408 does **not** close the error-name verification #412 assigned it, and that it needs the rig.

## Disposition

Block resolved; pass with actions 2–4.

## Next gate

Ship gate authorized.

muness:
# Gate 3 — Ship · REVIEW

**Gate/role:** ship gate, reviewer
**Verdict:** pass-with-actions (no merge — operator approval required)
**Issue:** #408 · **PR:** #413 · **Base:** `v3` · **Head:** `a550674`

## Aim and scope checked

Deliver a test-only Roon protocol fake sufficient for the epic's dependent issues, prove the `title`/`subtitle` hole is closed, document what is and is not covered, keep production changes to what testability requires, and do not fix what the instrument finds.

## Verifiable evidence

**CI is a real gate on this PR** (base `v3` matches `pull_request: branches: [master, v3]`). All checks settled green: **Lint, Test, Plan, Build WASM Assets, Synology Package Contract, CodeRabbit.** Notably CI's Test job **passes** the `/hqp/discover` tests that fail on a dev Mac, which settles them as a local-environment artifact rather than repo breakage.

The full gate is Lint = `cargo fmt --check` + `cargo clippy -- -D warnings`, Test = `cargo test --workspace`, `make css` first in both. All four now run locally before each push; `cargo test --workspace` was added after the gate table was corrected mid-task and shows the one local `/hqp/discover` failure and nothing else.

Local, pasted in the body rather than summarised:

```
cargo test --test roon_protocol   49 passed, 1.72s
cargo test --lib                  86 passed
cargo clippy -- -D warnings       clean
cargo fmt --check                 clean
make css && dx build --release --platform web --features web    both stages OK
```

Every other test target run individually, all green and unchanged. Flakiness: five consecutive runs, plus `--test-threads=1` and `--test-threads=16`.

**The two demonstrations the AC required, with pasted failure output:**
`SearchResultItem::from` swapped → `search_results_survive_the_http_boundary` fails.
`BrowseItemResponse` swapped → `browse_items_survive_the_http_boundary` fails.
Plus a third: `search()` returning the wrong browse level → `search_maps_title_and_subtitle_without_swapping_them` fails. All reverted, `git diff` clean.

**The cross-base verification**, which is the strongest single piece of evidence here: the same 49 tests, unchanged, report `DroppedByAdapter` ×4 on `v3` and `RoutedToCaller` ×4 when merged locally with `fix/issue-405-roon-browse-errors` — with `cargo test --lib`, clippy and fmt clean on the merged state. The fake demonstrably drives #412's `Parsed::Error` arm.

## Installability / verifiability

Nothing ships to users. The only production change is a `#[doc(hidden)]` function with no production callers plus a private enum on an internal function's signature. Runtime behaviour is byte-identical: `AdapterLogic::run` passes `CoreConnect::Discovery` and the discovery call is the same call it was. Verified by the unchanged pass of every existing suite and by the release build.

## Findings

**P2 — `classify_rejection` accepts two outcomes and must be tightened.** Today it cannot catch a regression from `RoutedToCaller` back to `DroppedByAdapter` — the exact defect #405 fixed. Recorded as a scheduled edit in `tests/mock_servers/README.md` ("Two scheduled edits") with the precise change: make `DroppedByAdapter` a hard failure and assert `RoonBrowseError::from_error(&e).kind == InvalidItemKey`. This is a real, dated debt, not a nice-to-have.

**P2 — `concurrent_browses_in_one_session_cross_deliver_their_results` will rot.** It asserts a defect is *present*, so it turns red when the defect is fixed and the person seeing that red will not know it is good news. Named explicitly in the README's scheduled-edits section rather than buried, per the execute-gate dissent.

**P2 — The cross-delivery defect needs its own issue.** Measured 7/8; survives #405; reachable through the caller-supplied `session_key` on `POST /roon/browse`; and #399's navigation handle is exactly the shape that triggers it. Not created here (creating issues was not authorized), so it is written up in the PR body with the measurements. **It should be filed before #399 designs around it.**

**P2 — The provenance work was reopened and materially improved after the recording criterion was restated.** An earlier revision flagged six shapes INFERRED and stopped there, which is the "hand-write a shape and let the fake agree with you" failure #408 exists to end. Chasing published sources moved five of them to documented (`list.level` "increases from 0", the `action` and `hint` value sets, `input_prompt`, `InvalidRequest`) and corroborated two off real Cores (`MOO/1 COMPLETE InvalidItemKey` verbatim, and the `"<int>:<int>"` key shape). `list.level` was load-bearing for #399. This should have been done in the first pass.

**P1 — The epic's `item_key` question now has evidence, and it points against the repo.** `home-assistant/core#137605` reports a Roon playlist that "works on the first attempt but fails on the second" with `MOO/1 COMPLETE InvalidItemKey` — the same shape as UHC's `play_item` — and community thread 23129 says keys "change with each refresh". That is `PerSession` behaviour. Not proof, and labelled as such, but **#396 should not design on the assumption keys are portable, and `/roon/play_item` should be assumed broken until the rig says otherwise.** This is the single most consequential finding in the PR and it is not something the fake could have produced on its own.

**P3 — The error-name verification is narrowed, not closed.** `InvalidItemKey` is corroborated; the other three appear nowhere in RoonLabs' published API, which documents browse errors only as "an error code or false if no error". #408 takes it from four unknowns to three and pins the consequence. Still needs the rig.

**P3 — `MockRoonCore` is dead code.** Six self-tests, no other callers. Left in place; removal candidate.

**P4 — 1337 lines of fake for a 2055-line adapter.** The honest price of one asserted Roon success path. Mitigated by keeping library shape as data, not as protocol code.

## Required actions

1. File the cross-delivery defect as an issue before #399 starts.
2. Do the two scheduled edits when their triggers land.
3. When a Core is next reachable: verify `InvalidLevels` / `UnexpectedError` / `ZoneNotFound`, the browse-root hierarchy names, whether search results are category-grouped, and `item_key` portability. All are marked in the code.
4. Carry the `item_key` finding to #396 **before** it designs its ref table, and to #405/#412's follow-up. A ref table that assumes portable Roon keys inherits a defect.

## Rollback

Revert `a550674`, `cc74ca0`, `a64b745`, `19bed47`, `707833e`, `e604c0c`, or simply don't merge. No migration, no config, no persisted state, no HTTP route change, `tests/fixtures/api_routes.txt` untouched so no `api-change-approved` label. Reverting the seam alone (`707833e`) leaves `tests/roon_protocol.rs` uncompilable, so the fake and its seam revert together.

## Disposition

Ready for operator review. **Not merged, no labels added, no force-push.**

## Residual risk

Green here means "unchanged", not "correct" — the fake was derived from the same fork and adapter it tests. Stated at the top of both new files, in the README, and in the PR body, because the failure mode of this deliverable is someone quoting it as evidence about Roon.

muness:
# Gate 3 — Ship · DISSENT

**Gate/role:** ship gate, dissent
**Verdict:** pass-with-actions (do not merge without the operator, and read action 1 first)
**Issue:** #408 · **PR:** #413

Everything is green, which is exactly when this role earns its keep. So: **what will someone believe because of this PR that is not true?**

## The one thing most likely to go wrong

Not a bug in the fake. It is this sentence, which will be quoted:

> "Roon's success path is now tested."

It is not. What is tested is that `RoonAdapter` behaves consistently against a fake built from the same fork the adapter was written against, following a library hierarchy read out of the adapter's own source. If Roon renamed `Search` tomorrow, `search()` would break in production and all 49 tests would stay green. If Roon spells `InvalidItemKey` differently, every rejection test stays green and every real caller times out.

#408's own words were "Roon's wire behaviour is asserted against something that answers like Roon." The honest version is: *against something that answers the way this repo believes Roon answers.* Those differ, and the difference is the entire unverified surface.

The PR says this in four places. **Required action 1:** whoever closes #408 must not write "retires the standing caveat that Roon's success path is unverified" — because it does not. It replaces "unverified and untestable" with "unverified but now instrumented, and here is the list of what to check when you have a Core." That is a large improvement and a smaller claim.

## What I attacked and could not break

- **Is the seam a lookalike loop?** No. The #412 experiment settles it: the same test file changed behaviour purely because the adapter changed, which is only possible if the tests drive the production loop.
- **Are the three swap demonstrations real?** Yes — pasted failure output with actual values, and the swaps are the type-checking form a careless refactor produces, not `title: item.subtitle` which would not compile.
- **Is the cross-delivery finding a fake artifact?** No. Verified it survives #412, whose own comment predicts the gap. Measured 7/8 across independent sessions.
- **Is the test count padded?** 46→49: the cross-delivery finding, the unrecognised-error-name consequence, and the fork's four literals. None decorative.
- **Does anything reach the operator's real config directory?** No, and it is asserted: the handshake test checks the pairing state landed in the temp dir.

## Concerns I am not withdrawing

**The two scheduled edits are the weakest part of this delivery.** `classify_rejection` tolerating `DroppedByAdapter`, and a test that asserts a defect is present, both depend on a human reading a README section at the right future moment. That is not a control. I considered demanding a compile-time forcing function — a `cfg` probe on `RoonBrowseError`'s existence — and rejected it: conditional compilation on a type's presence is worse than the debt it prevents. So the debt stands, named and dated. **If #412 merges and nobody tightens `classify_rejection`, this suite will silently tolerate a regression to the exact bug #405 fixed.**

**Filing the cross-delivery issue is the highest-value follow-up in this PR and it is not done.** Issue creation was not authorized, so it lives in a PR body — the least durable place for a finding. #399 is the issue that will build a reusable navigation handle straight onto this defect. **Required action 2:** file it before #399 starts, with the 7/8 measurement.

**A 1337-line fake is now load-bearing test infrastructure that nothing tests except itself.** Six unit tests cover the framing codec and the key arena. The browse/load state machine — level stacks, paging arithmetic, per-session isolation, the search filter — is covered only transitively, by the adapter tests passing. If the fake has an off-by-one in its paging window, `load_pages_within_a_level_and_reports_the_total` asserts the fake's arithmetic against itself. That is genuinely circular, and it is the residual form of the exact problem #408 was raised to fix, one level down. Not a block: the alternative is testing the test, indefinitely. But the next agent should know that the paging and level-stack semantics are the least independently verified thing in this PR.

**The `#[doc(hidden)] pub` seam is still unenforced.** Unchanged from gate 1. `architecture_lint` could assert no non-test callers; that is a follow-up and this PR should not claim `#[doc(hidden)]` is a control. It does not.

## Contrary case for not merging at all

Worth stating: the epic could proceed without this. #405 shipped with unit-scoped tests; #396 can too. Merging adds 2700 lines of test surface, one dev-dependency, and a permanent `pub` seam in the adapter, in exchange for assertions whose ceiling is self-consistency.

I do not press it. The three swap demonstrations are real defects caught at three real layers; the cross-delivery finding is a genuine bug that no unit-scoped test would have surfaced; and #412's own author documented that the wire path could not be driven, which this now drives — proven, not asserted. That is more than confidence. But the case for merging rests on those specific artefacts, not on "Roon is now covered."

## Required actions

1. Whoever closes #408 states the smaller, true claim — instrumented, not verified — and does not retire the unverified-success-path caveat.
2. File the same-session cross-delivery defect as an issue before #399 starts, with the 7/8 measurement.
3. Tighten `classify_rejection` when #412 lands on `v3`. If it is not tightened, the suite tolerates a regression to #405's bug.
4. On the next contact with a real Core, check the three marked items: the four error names, the browse-root hierarchy names, `item_key` portability.

## Disposition

Pass with actions. **No merge** — operator approval required per AGENTS.md and #392. No labels, no force-push.

## Residual risk

The single highest risk in this deliverable is not technical. It is that a well-documented fake reads as verification, and every later Roon issue inherits that comfort. The README's "green means unchanged, not correct" line is the only defence, and it works only for people who read it.

muness:
# Gate 3 — Ship · DISSENT, addendum

The dissent above was posted from a stale file — a scripted edit failed its assertion and the pre-edit version went out. The missing section is the most important one, so here it is separately rather than as a silent re-post.

## The correction that should have come from inside this task

The recording criterion was in #408 from the start: *"where a response shape was uncertain, it is recorded from a real Core and the provenance noted — not guessed."* My first pass read "no Core reachable" as licence to mark six shapes `INFERRED` and move on. That is technically honest labelling and it is still the wrong instinct, because **"I could not record it" is not the same as "there is nothing to check."**

There was plenty to check, and it took twenty minutes:

* RoonLabs publishes the browse API's semantics. `list.level` **"increases from 0"** — I had that as "plausible, unchecked", and it is load-bearing for #399's "every browse page reports its position". Also documented: the five `action` values, the five `hint` values, the `input_prompt` shape, and `InvalidRequest` as a genuine reply to an unknown service.
* Real Cores' `InvalidItemKey` frames are sitting in public Home Assistant issue logs: `Could not play id:122:4, result: MOO/1 COMPLETE InvalidItemKey`. That confirms the name *and* that the reply is verb-and-name with no body — the exact frame this fake sends.

Five shapes moved from guess to documented, two to recorded-off-real-Cores.

## Why this matters more than the shapes themselves

A fake's author is the worst-placed person to notice that the fake agrees with them, and "unverifiable" is the most comfortable conclusion that author can reach. #407 caught a real client-visible bug because a recorded fixture **contradicted its author's premise**. Nothing in my first pass could have done that, because nothing in it could disagree with me.

The provenance chase created the possibility of disagreement, and it immediately produced one: `home-assistant/core#137605` shows a Roon playlist that "works on the first attempt but fails on the second" with `InvalidItemKey` — the same shape as UHC's `play_item` — and community thread 23129 says item keys "change with each refresh". **That is evidence this repo's `item_key` portability assumption is wrong**, and it did not come from the fake. It came from looking for something that could contradict me.

## Amended required actions

4. On the next contact with a real Core, check the marked items: `InvalidLevels` / `UnexpectedError` / `ZoneNotFound`, the browse-root hierarchy names, whether search results are category-grouped, and `item_key` portability.
5. **Treat the `item_key` evidence as design input for #396 now, not as a curiosity.** A ref table built on the assumption that Roon keys are portable inherits a defect that public logs already show. #396's own solution-space gate chose to pair each Roon `item_key` with its minting `multi_session_key` in the ref record — this evidence supports that choice and argues against any later simplification of it.

## Correction to the section above

Where the posted dissent says "the four error names" are unverified: **one of the four is now corroborated.** `InvalidItemKey` is real, recorded verbatim. `InvalidLevels`, `UnexpectedError` and `ZoneNotFound` are not — RoonLabs' published browse API documents errors only as "an error code or false if no error" and never enumerates them, and none of the three appears in `node-roon-api`. So #408 **narrows** that question from four unknowns to three and pins the consequence. It does not close it, and the ship report should say exactly that.

muness:
# Ship gate · note on how the CI evidence was gathered

Worth recording because it is the same class of error this PR is about, committed while verifying this PR.

My first CI waiter polled `gh pr checks` and declared done when no check reported `pending`. Twice it announced **"ALL CHECKS SETTLED"** while listing only `CodeRabbit: pass` — because immediately after a push, `gh pr checks` transiently returns a *partial* set, and "none of the checks I can see is pending" is trivially true of an incomplete list. Lint and Test had not been created yet, let alone run.

So the waiter's success condition was satisfied by **absence of evidence**, which is exactly the failure mode of a mock nothing asserts against: it reported green because it could not see anything to be red.

Replaced with a waiter that polls the workflow *run* for the PR's current head SHA until `status == "completed"`, then enumerates every job's conclusion and exits non-zero if the run failed. That cannot pass on a partial set, because the run object knows how many jobs it has.

**The CI results quoted in the ship review were re-verified through the second method.** Anything in this thread that cites CI from before that point should be read as provisional.

Two things follow for whoever reads this PR:

1. Do not take an earlier "ALL CHECKS SETTLED" line in these comments as authoritative; the per-job enumeration is.
2. The pattern generalises. A green signal is only as good as its ability to be red — for CI waiters, for mocks, and for the two rejection tests in this PR that were originally written so the fix would make them fail.

muness:
## CI: authoritative result for the final commit

Verified by polling the workflow *run* to completion and enumerating jobs, not by `gh pr checks` — see the note above on why the first method produced two false "all settled" reports.

```
$ # head = 6cbf604, run 30675725543
RUN COMPLETED: completed/success
Synology Package Contract: success
Test: success
Lint: success
Build WASM Assets: success
Plan: success
Build Linux x64: success
Smoke Test Binary: success
Build Summary: success
Post Artifact Links: success
```

Every job that ran, succeeded. **Lint** is `cargo fmt --check` + `cargo clippy -- -D warnings`; **Test** is `cargo test --workspace`, and it passes the `/hqp/discover` tests that fail on a developer machine — which is what settles those as a local-environment artifact rather than repo breakage.

Local gate before every push: `make css`, `cargo fmt --check`, `cargo clippy -- -D warnings`, `cargo test --workspace`, plus `cargo test --test roon_protocol` (49 passed, 1.7s) and a `dx build --release --platform web --features web`.

Still **not merged**, no labels, no force-push — operator approval required per AGENTS.md and #392.

Issue #418, opened 2026-08-01
MCP album art: bounded, explicitly-requested cover art behind a reference


muness:
OH: 80222d6d

Parent epic: #392

## Problem Statement

**Current framing:** Add album art to the MCP surface.

**Reframed as:** The bridge fetches, scales, and serves cover art for every surface it has — the knob, iOS, and the web UI all get it — but MCP clients get none. `McpNowPlaying` does not even carry the `image_key` the aggregator already holds, so an assistant can describe what is playing and cannot show it. Meanwhile the naive fix is actively harmful: inlining image bytes into a tool result costs thousands of tokens of the model's context per call, so art must be *addressable* rather than *attached*.

**The shift:** From "art is a field on now-playing" to "art is a bounded, explicitly-requested payload behind a reference" — the same instinct as #396's content refs, with a different failure mode driving it.

### What already exists

- **Roon** — `RoonAdapter::get_image(image_key, width, height)` (`src/adapters/roon.rs:511`) with `Scaling::new(Scale::Fit, w, h)`. Served at `/roon/image`.
- **LMS** — `get_artwork(image_key, w, h) -> (content_type, bytes)` and `get_artwork_url(coverid, w, h)` (`src/adapters/lms.rs:930,949`), which builds `cover_WxH.jpg`.
- **OpenHome** — already extracts `upnp:albumArtURI` from DIDL-Lite into `image_key` (`src/adapters/openhome.rs:650,750`). A URL on the renderer's network; UHC does not fetch or proxy it.
- **UPnP** — `image_key: None`, hardcoded (`src/adapters/upnp.rs:687`), despite reading the same DIDL-Lite that OpenHome mines for it. **Not-yet-implemented, not a provider limit.**
- **The pinned SDK supports both delivery shapes**: `ImageContent { data: base64, mimeType }` and `BlobResourceContents` (rust-mcp-schema 0.9.5, `mcp_schema.rs:3767` and `:231`).
- The knob and web already consume `/knob/now_playing/image` and `/now_playing/image`.

So this is projection work, not new capability — with one genuinely new decision: how the bytes reach the client.

### Constraints

- **Hard — never inline art unbidden.** `hifi_now_playing` must not start returning image bytes. A 300×300 JPEG is tens of KB, and base64 inflates it by a third; at typical tokenisation that is thousands of tokens per call, silently, on the epic's most-called read tool. Art is returned only when a tool call explicitly asks for it.
[...]

PR #421, opened 2026-08-01
fix(upnp): read track metadata from GetPositionInfo, and share DIDL-Lite parsing

### LINKED ISSUE #420: UPnP adapter: no track metadata at all, behind a false 'pure UPnP doesn't provide it' comment
OH: 80222d6d

Parent epic: #392
Found while scoping #418 (MCP album art)

## Problem Statement

**Current framing:** UPnP's `image_key` is hardcoded `None`; populate it.

**Reframed as:** The UPnP adapter reports **no track metadata at all** — `get_now_playing` returns the renderer's own name as `line1`, empty strings for `line2`/`line3`, and `None` for `image_key`, with the comment *"Pure UPnP doesn't provide track metadata"* (`src/adapters/upnp.rs:674`). That comment is false, and it is a false capability claim of exactly the kind #392 exists to eliminate. UPnP AVTransport provides `GetPositionInfo`, whose `TrackMetaData` carries DIDL-Lite with `dc:title`, `upnp:artist`, `upnp:album`, and `upnp:albumArtURI`. The proof is in this repo: **OpenHome already parses that same DIDL-Lite**, and OpenHome devices *are* UPnP devices.

**The shift:** From "UPnP is a transport-only backend" to "UPnP metadata was never wired," which is a UHC gap, not a protocol limit.

### Evidence

- `src/adapters/upnp.rs:674` — `// Pure UPnP doesn't provide track metadata`, then `line2: String::new()`, `line3: String::new()`, `image_key: None`.
- `src/adapters/upnp.rs:898` — `now_playing: None, // UPnP track info would need separate DIDL-Lite parsing`.
- `poll_renderer` calls `GetTransportInfo`, `GetVolume`, and `GetMute` but **never `GetPositionInfo`** — the one action that carries metadata.
- Everything needed is already present: `soap_call()` (`:543`), `AV_TRANSPORT_URN` (`:30`), and `extract_xml_value()` (`:574`).
- `src/adapters/openhome.rs:638` `parse_didl_lite()` and `:934` `html_decode()` do the work already, and `openhome.rs:650` reads `upnp:albumArtURI` — the very field UPnP discards.

### Constraints

- **Hard — do not duplicate the DIDL parser.** AGENTS.md: when you see an architectural inconsistency, fix it; don't work around it. Two adapters parsing the same DIDL-Lite two ways is a defect in waiting. Extract `TrackInfo`, `html_decode`, and `parse_didl_lite` into one shared module and have **both** adapters use it. `TrackInfo` is referenced only inside `openhome.rs` today, so the move is contained.
- **Hard — the extraction must be behavior-preserving for OpenHome.** There are currently **no tests** for `parse_didl_lite` or `html_decode`, so write them against the existing implementation *first*, watch them pass, and only then move the code. Same discipline #394 used: pin the behavior before restructuring it.
- **Hard — this changes client-visible payloads for UPnP zones.** They go from `line1: <renderer name>`, empty `line2`/`line3` to real title/artist/album, and `image_key` becomes populated. The knob, iOS app, and web UI all read this. Check `tests/client_harness.rs` for what the ESP32 and iOS clients expect, and AGENTS.md's API-stability rules, **before** changing the shape. Filling previously-empty fields is additive, but it changes what a user sees — call it out explicitly in the PR rather than treating it as a bug fix.
- **Hard — bound the polling cost.** `GetPositionInfo` on every poll for every renderer adds a SOAP round-trip per renderer per tick. Mirror OpenHome's guard: only re-parse when the track URI changes (`src/adapters/openhome.rs:558-560`).
- **Hard — `TrackMetaData` is XML-escaped** in the `GetPositionInfo` response; it must be `html_decode`d before the extractor can see the tags. This is the detail that makes a naive implementation silently return nothing.
- **Soft:** whether art is fetched/proxied or only its URI recorded (fetching is #418's concern); whether genre and duration are also surfaced; module name and location.

### What this framing enables

UPnP zones that show what is playing on every surface UHC has — the knob, iOS, and web all read the same `NowPlaying`. It also makes #418's four-way art capability matrix honest: UPnP moves from *not-yet-implemented* toward *supported*, rather than sitting behind a comment claiming the protocol cannot do it.

Fetching or proxying art bytes — that is #418. Any MCP surface change. New HTTP routes. Rewriting UPnP discovery or transport.

## Acceptance criteria

- [ ] Tests for `parse_didl_lite` and `html_decode` written against the **current** implementation and passing before any code moves — the first coverage either has ever had.
- [ ] `TrackInfo`, `html_decode`, and `parse_didl_lite` live in one shared module, used by both `openhome.rs` and `upnp.rs`. No duplicated parser.
- [ ] OpenHome behavior is unchanged by the extraction, proven by the tests above.
- [ ] UPnP polls `GetPositionInfo`, `html_decode`s `TrackMetaData`, parses DIDL-Lite, and populates title, artist, album, and `image_key`.
- [ ] Re-parsing is guarded on track-URI change, mirroring OpenHome; a test proves an unchanged URI does not re-parse.
- [ ] A test drives this through `MockUpnpRenderer` with a realistic escaped `GetPositionInfo` response, asserting all four fields — and failing against current code.
- [ ] A renderer that does not implement `GetPositionInfo`, or returns empty metadata, degrades to today's behavior without error spam.
- [ ] The false comment at `:674` and the stale one at `:898` are removed or corrected.
- [ ] The client-visible payload change is enumerated in the PR, with `tests/client_harness.rs` expectations checked and stated.
- [ ] `cargo fmt --check` and `cargo clippy -- -D warnings` clean; `cargo test --workspace` green; `make css` first. For `dx`, `CARGO_BUILD_RUSTC_WRAPPER=""` — `unset RUSTC_WRAPPER` does not work.

## Dependencies

None — adapter work, independent of the MCP chain. **Informs #418** (album art: UPnP's capability state) and **#398** (capability matrix: another entry that reads as a provider limit but is not).

## Notes for the implementer

- `src/adapters/openhome.rs:555-580` is the working reference for the whole sequence: URI-change guard, `html_decode`, `parse_didl_lite`, then publish `BusEvent::NowPlayingChanged`.
- `tests/mock_servers/upnp.rs` already has `MockUpnpRenderer` with SOAP action handling and `set_state`; extend it rather than building something new.
- Concurrent agents share one session scratchpad — prefix scratch files with `419-`.

muness:
Fixes #420

## What this changes

UPnP zones now report what is playing. Previously the adapter returned the renderer's own name as `line1`, empty `line2`/`line3`, and `image_key: None`, behind a comment claiming *"Pure UPnP doesn't provide track metadata."* That was false: `AVTransport::GetPositionInfo` returns `TrackMetaData` carrying DIDL-Lite, and OpenHome in this very repo already parses that same format.

Three commits, in TDD order:

1. **`1debc38`** — pin `parse_didl_lite` / `html_decode` behavior with tests. Neither had any coverage. Ran green against the original code before anything moved.
2. **`58eab9f`** — extract `TrackInfo`, `html_decode`, `extract_xml_value`, `parse_didl_lite` into `src/adapters/didl.rs`, used by both adapters. Not a pure move; see the behavior change below.
3. **`0aea46c`** + **`f0ba5e4`** — red tests, then the UPnP implementation.

## Two things worth reviewer attention

**1. A behavior change rides in the extraction, and it fixes a live bug.** `extract_xml_value` matched a literal `<tag>`, so any tag carrying attributes was invisible to it. Real DIDL-Lite writes:

```
<upnp:albumArtURI dlna:profileID="JPEG_TN">http://…/art.jpg</upnp:albumArtURI>
```

which means **OpenHome album art has been silently missing on every device following that DLNA convention** — not just UPnP's. The matcher now tolerates attributes, skips self-closing tags, and guards against prefix collisions (`upnp:album` no longer risks matching `upnp:albumArtURI`). The pin test that documented the limitation is now a regression test for the fix.

This is a fix, not a regression, but it changes what OpenHome users see: art may start appearing where it never did. Flagging it rather than filing it under the refactor.

**2. This changes client-visible payloads for UPnP zones.** `Zone.now_playing` goes from always-`None` to populated, and `line1`/`line2`/`line3`/`image_key` carry real values. The knob, iOS app, and web UI all read this. I checked `tests/client_harness.rs` first: it registers the UPnP adapter but asserts nothing about now-playing content, and all 42 harness tests pass unchanged. Fields that were empty become populated — additive in shape, but a user-visible difference, so it is your call rather than mine.

## One production seam

`probe_renderer_for_test`, marked `#[doc(hidden)]`: registers a renderer and polls it once, as SSDP-then-poll would. SSDP cannot reach a mock and there was no other public path in, so without it the poll loop stays untestable. Same approach #408 took for Roon.

Worth noting *why* this gap survived: the only existing UPnP test asserted that the mock serves a description document. The adapter was never driven against it — the third instance of that pattern this epic has found, after the HQPlayer mock rejecting its own adapter's wire format (#394) and `MockRoonCore` not being a server (#408).

## Verification

Run locally, complete gate:

- `cargo fmt --check` — clean
- `cargo clippy -- -D warnings` — clean
- `cargo test --workspace` — all green, including the two `/hqp/discover` tests that fail on some dev machines
- `dx build --release --platform web --features web` — client and server both clean, first attempt

This PR targets `v3`, so CI is a real gate here and I expect it to run.

`--all-targets` clippy is not attempted: `src/lib.rs:20-22` denies `unwrap_used`/`expect_used`/`panic` crate-wide and src's own test modules violate it 76 times, pre-existing.

## Deliberately not done

- **Album art bytes are not fetched or proxied.** UPnP's `albumArtURI` is a URL on the renderer's network. Exposing art over MCP is #418, which has to decide the delivery mechanism and the reachability question.
- **`unsupported: ["next", "previous"]` at `src/adapters/upnp.rs:715` is left alone**, with its comment *"Pure UPnP doesn't support these."* AVTransport:1 *does* define `Next` and `Previous` actions, so I suspect this is another false-provider-limit claim of the same shape as the one this PR fixes — but real support varies per device and verifying it needs hardware I do not have. Not changed, and not asserted either way. Candidate follow-up.
- **No live hardware.** Everything here is proven against `MockUpnpRenderer` with a realistic escaped `GetPositionInfo` payload. A real renderer may order or namespace DIDL-Lite differently; the parser is tolerant of attributes and tag order, but that is reasoning, not evidence.

## Rollback

Revert the three commits. `didl.rs` is additive and both adapters route through it, so reverting the feature commit alone leaves the shared module in place and harmless.

## Summary by CodeRabbit

* **New Features**
  * UPnP playback now displays track title, artist, album, and artwork metadata.
  * Metadata updates automatically as tracks change and clears when unavailable.
  * Added support for richer metadata, including genre and common XML/HTML formatting variations.

* **Bug Fixes**
  * Prevented stale track information from persisting when metadata is missing.
  * Improved compatibility with metadata containing attributes, self-closing tags, or missing fields.

* **Tests**
  * Added comprehensive coverage for UPnP metadata parsing, fallbacks, and repeated polling.

### DIFF
--- a/src/adapters/didl.rs
+++ b/src/adapters/didl.rs
@@ -1,0 +1,200 @@
+//! DIDL-Lite metadata parsing, shared by the OpenHome and UPnP adapters.
+//!
+//! Both backends carry track metadata as DIDL-Lite: OpenHome via its own
+//! metadata event, UPnP via `AVTransport::GetPositionInfo`'s `TrackMetaData`.
+//! In both cases the payload arrives XML-escaped inside another XML document,
+//! so it must be [`html_decode`]d before [`parse_didl_lite`] can see the tags.
+//!
+//! This module exists so the two adapters cannot drift apart in how they read
+//! the same format.
+
+use serde::{Deserialize, Serialize};
+
+/// Track metadata extracted from a DIDL-Lite document.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Default)]
+pub struct TrackInfo {
+    pub title: String,
+    pub artist: String,
+    pub album: String,
+    pub album_art_uri: Option<String>,
+    pub genre: Option<String>,
+}
+
+/// Decode the XML entities that wrap escaped DIDL-Lite payloads.
+///
+/// `&amp;` is replaced last so that a double-encoded `&amp;lt;` decodes to
+/// `&lt;` rather than collapsing all the way to `<`.
+pub fn html_decode(s: &str) -> String {
+    s.replace("&lt;", "<")
+        .replace("&gt;", ">")
+        .replace("&amp;", "&")
+        .replace("&quot;", "\"")
+        .replace("&apos;", "'")
+}
+
+/// Read the text content of `tag`, tolerating attributes on the opening tag.
+///
+/// Attribute tolerance is load-bearing rather than cosmetic: real renderers
+/// emit `<upnp:albumArtURI dlna:profileID="JPEG_TN">`, and a matcher that only
+/// accepts a bare `<tag>` silently returns nothing for the most common form of
+/// the one field album art depends on.
+pub fn extract_xml_value(xml: &str, tag: &str) -> Option<String> {
+    let end_tag = format!("</{}>", tag);
+    let mut from = 0;
+
+    // Scan for an opening tag whose name matches exactly, so that looking for
+    // `title` never matches `albumTitle` and looking for `upnp:album` never
+    // matches `upnp:albumArtURI`.
+    while let Some(rel) = xml[from..].find(&format!("<{}", tag)) {
+        let open_start = from + rel;
+        let after_name = open_start + 1 + tag.len();
+        let rest = &xml[after_name..];
+
+        // The character after the tag name decides whether this is our tag.
+        match rest.chars().next() {
+            // Exact match, no attributes.
+            Some('>') => {
+                let value_start = after_name + 1;
+                let end = xml[value_start..].find(&end_tag)? + value_start;
+                return Some(xml[value_start..end].to_string());
+            }
+            // Attributes present: skip past them to the end of the opening tag.
+            Some(c) if c.is_whitespace() => {
+                let close_rel = rest.find('>')?;
+                // A self-closing tag carries no text content.
+                if rest[..close_rel].ends_with('/') {
+                    from = after_name + close_rel + 1;
+                    continue;
+                }
+                let value_start = after_name + close_rel + 1;
+                let end = xml[value_start..].find(&end_tag)? + value_start;
+                return Some(xml[value_start..end].to_string());
+            }
+            // A longer tag name that merely starts with ours — keep looking.
+            _ => {
+                from = after_name;
+            }
+        }
+    }
+
+    None
+}
+
+/// Parse a decoded DIDL-Lite document into [`TrackInfo`].
+///
+/// Always returns `Some`: absent fields become empty strings or `None`, which
+/// is what both adapters have always relied on.
+pub fn parse_didl_lite(xml: &str) -> Option<TrackInfo> {
+    let title = extract_xml_value(xml, "dc:title")
+        .or_else(|| extract_xml_value(xml, "title"))
+        .unwrap_or_default();
+
+    let artist = extract_xml_value(xml, "upnp:artist")
+        .or_else(|| extract_xml_value(xml, "dc:creator"))
+        .unwrap_or_default();
+
+    let album = extract_xml_value(xml, "upnp:album").unwrap_or_default();
+    let album_art_uri = extract_xml_value(xml, "upnp:albumArtURI");
+    let genre = extract_xml_value(xml, "upnp:genre");
+
+    Some(TrackInfo {
+        title,
+        artist,
+        album,
+        album_art_uri,
+        genre,
+    })
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// A realistic DIDL-Lite payload as a renderer returns it, with attributes
+    /// on the tags that carry them in practice.
+    const REAL_DIDL: &str = r#"<DIDL-Lite xmlns="urn:schemas-upnp-org:metadata-1-0/DIDL-Lite/" xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:upnp="urn:schemas-upnp-org:metadata-1-0/upnp/"><item id="1" parentID="0" restricted="1"><dc:title>Hoppipolla</dc:title><upnp:artist>Sigur Ros</upnp:artist><upnp:album>Takk...</upnp:album><upnp:genre>Post-rock</upnp:genre><upnp:albumArtURI dlna:profileID="JPEG_TN">http://10.0.0.5/art/42.jpg</upnp:albumArtURI></item></DIDL-Lite>"#;
+
+    /// The same payload with a bare albumArtURI tag (no attributes).
+    const BARE_DIDL: &str = r#"<DIDL-Lite><item><dc:title>Hoppipolla</dc:title><upnp:artist>Sigur Ros</upnp:artist><upnp:album>Takk...</upnp:album><upnp:albumArtURI>http://10.0.0.5/art/42.jpg</upnp:albumArtURI></item></DIDL-Lite>"#;
+
+    #[test]
+    fn html_decode_unescapes_in_the_right_order() {
+        // &amp; must be replaced last, or "&amp;lt;" would wrongly become "<".
+        assert_eq!(html_decode("&lt;a&gt;"), "<a>");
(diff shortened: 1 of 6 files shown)


muness:
All four CodeRabbit findings were real and are fixed in `b7a4f4e`. Taking them in order of how much they mattered:

**`track_metadata` still advertised as unsupported (Major, data integrity)** — the worst of the four, and the most embarrassing: this PR adds track metadata and then left it in the zone's `unsupported` list, so any client reading that list would hide the data. That is exactly the capability-honesty rule epic #392 exists to enforce — *never report a UHC gap as a provider limit* — broken inside a PR that belongs to the epic. Removed. `album_art` stays, correctly: the URI is parsed, but UHC does not fetch or proxy the bytes yet (#418). Now covered by a test.

**Constant `TrackURI` freezing metadata (Major, correctness)** — a genuine functional bug I would not have found without hardware. Internet radio and many DLNA servers hold one stream URI for a whole session and change only `TrackMetaData`, so a URI-only guard pins now-playing to the first track forever. Added `last_track_metadata` and compare both. New test `upnp_reparses_when_metadata_changes_under_a_constant_track_uri` fails against the previous guard.

Worth flagging beyond this PR: **`src/adapters/openhome.rs` has the same URI-only guard**, so OpenHome very likely has the same bug. Not fixed here — it needs its own red test and its own blast-radius statement rather than riding along in a UPnP PR. Filing it as a follow-up.

**Two XML extractors in one file (Major, maintainability)** — correct, and it is the specific AGENTS.md rule I quoted in issue #420 while violating it: *a refactor must update every applicable call site, not merely add the definition.* I routed the new code at `didl::extract_xml_value` and left `Self::extract_xml_value` (regex) serving lines 471/573/594. Your sequencing advice was right too — the regex version tolerated an optional namespace prefix, so swapping call sites first would have silently dropped that. `didl::extract_xml_value` now accepts an optional prefix **when the requested tag carries none** (`CurrentVolume` matches `<u:CurrentVolume>`), while a prefixed tag stays exact so `upnp:album` still cannot swallow `upnp:albumArtURI`. Duplicate removed, all call sites routed, orphaned `regex::Regex` import dropped. Three new tests, including one asserting a colon inside an attribute value is not mistaken for a prefix.

**`html_decode` ordering (Minor)** — right, and the sharpest observation in the review was that *my only test case passed by accident*. The doc comment said `&amp;` is replaced last; the code replaced it third, so `&amp;quot;` decoded to `"` instead of `&quot;`. `&amp;lt;` passed only because `&lt;` happens to be replaced first. Moved `&amp;` last, added `&quot;`/`&apos;`/`&gt;` cases. Note this ordering bug pre-dated the PR — it came over from `openhome.rs` — so it was a bug I preserved while documenting it as fixed.

Re-verified after the changes: `cargo fmt --check` clean, `cargo clippy -- -D warnings` clean, `cargo test --workspace` 23 binaries / 0 failures, and `dx build --release --platform web --features web` clean (first attempt failed on the known Homebrew-sccache flake, ~2 in 6; `cargo check --lib --features server,web` confirms no real compile error).

Still true and unchanged: **no real hardware has been touched.** Everything is proven against `MockUpnpRenderer`. The radio-metadata finding is a good illustration of what that cannot cover.

PR #422, opened 2026-08-01
fix(roon): concurrent browses sharing a session key no longer cross-deliver results (#416)

### LINKED ISSUE #416: Roon adapter: concurrent browses sharing a session key silently cross-deliver results
OH: 80222d6d

Parent epic: #392
Found by: #408's protocol fake (PR #413), reproduced in 7 of 8 trials

## Problem Statement

**Current framing:** Browse correlation was fixed by #405.

**Reframed as:** #405 fixed the *error* path. The **success** path still resolves waiters by scanning for the first pending request whose session key matches, so when two browse requests share a session key, **one caller silently receives the other's results.** Not an error, not a timeout — the wrong list, delivered as success.

**The shift:** From "correlation is a browse-error concern" to "correlation is wrong on the happy path, and the happy path is the one users are on."

### Evidence

- `src/adapters/roon.rs:1866-1877` — `Parsed::BrowseResult` finds its waiter with `.iter().find(|(_, (key, _))| key == &session_key)`. First match wins, regardless of which request the response belongs to. `Parsed::LoadResult` has the same shape.
- **`/roon/browse` lets callers supply the session key** (`src/api/mod.rs:627`: `req.session_key.unwrap_or_else(...)`), so two concurrent clients that pass the same key cross-deliver. Nothing prevents or detects it.
- `search()` reuses a single `search_{nanos}` key across six requests, so the pattern is already load-bearing internally.
- Reproduced in **7 of 8 trials** against the protocol fake from #408, and it **survives #405** — that PR made `req_id` primary for errors, leaving the success arms untouched.

### Why this matters more than it looks

**#399 is designing a browse tool whose client-facing nav handle is exactly this shape.** If the handle maps to a Roon session key and two MCP clients browse concurrently, one gets the other's library listing — and because refs (#396) are minted from browse results, a client could then play from a list it never requested. A cross-delivered *result* becomes a cross-delivered *action*.

### Constraints

- **Hard:** Correlate on `req_id`, consistent with what #405 established for the error path — the fork returns the same id `Browse::browse`/`load` handed back, allocated from one monotonic per-connection counter. Keep #405's session-key guard rationale intact: that counter restarts at 0 on reconnect, so a mismatch must leave the waiter to its existing timeout rather than resolve it wrongly. Do not regress #405 — its tests must stay green and its typed-error path untouched. `pending_images` shares the shape; check it. No HTTP route changes.
- **Hard:** Whatever the fix, a concurrent-browse test must prove it — one that fails against today's code. #408's fake (PR #413) can drive out-of-order responses with three requests in flight and is the right instrument.
- **Soft:** Whether callers keep the ability to supply a session key at all; whether the internal reuse in `search()` is left as-is or given distinct keys.

## Acceptance criteria

- [ ] Two concurrent browse requests sharing a session key each receive their own results; a test proves it and fails against current code.
- [ ] Same for `load`, and for `pending_images` if it shares the defect.
- [ ] Out-of-order responses correlate correctly with three or more requests in flight.
- [ ] #405's error-path tests remain green, and the reconnect-counter guard still leaves mismatches to time out rather than resolving them wrongly.
- [ ] A note in `/roon/browse`'s handler on what a caller-supplied session key does and does not guarantee — or the parameter's removal, if that is the chosen fix.
- [ ] `cargo fmt --check` and `cargo clippy -- -D warnings` clean; `cargo test --workspace` green; `make css` first.

## Dependencies

Depends on **#405 (PR #412)** landing first — same code region; stacking avoids a conflict. **Blocks #399**, whose nav-handle design cannot be settled until this is resolved. Uses **#408 (PR #413)** as its test instrument.

## Notes for the implementer

- Read #405's correlation rationale before designing — it reasoned from the fork's `req_id` allocation (`src/moo.rs:124-135`) and found that a session key cannot identify a request. This issue is that same argument applied to the arms #405 did not touch.
- **Feature-branch-based PRs run zero CI here; PRs targeting `v3` get the full gate and a red check is blocking.** Know which you are and say so.

muness:
Fixes #416. Parent epic #392. **Unblocks #399.** Stacked on #412 (#405); uses #413 (#408) as its instrument.

## What was wrong

Two browse requests in flight under one `multi_session_key` **silently delivered each other's results**. Not an error, not a timeout — the wrong list, returned as a success.

`Parsed::BrowseResult` found its waiter by scanning for the first pending request whose session key matched:

```rust
s.pending_browses.iter().find(|(_, (key, _))| key == &session_key)
```

The map is keyed by `req_id`; the scan is by session key. A session key names a browse **session**, not a request — `search()` reuses one across six requests, and `/roon/browse` lets a caller supply its own — so with more than one request in flight under one key, first-match-wins, and which one wins comes out of `HashMap` iteration order. `Parsed::LoadResult` had the identical arm.

#405 (PR #412) made `req_id` primary for the *rejection* arms and left these. This PR is that same argument applied to the arms it did not touch — the success arms, which are the ones users are on.

## CI: this PR gets none — local verification is the only gate

Its base is a feature branch (`fix/issue-405-roon-browse-errors`), and the workflows trigger on `pull_request: branches: [master, v3]`, which keys on the **base**. So zero workflows run here and a green (or absent) check proves nothing. Everything CI would run was run locally, on this branch's tip:

| CI job | Command | Result |
|---|---|---|
| — | `make css` (first, per #392's standing facts) | ok |
| Lint | `cargo fmt --check` | clean |
| Lint | `cargo clippy -- -D warnings` | clean |
| Test | `cargo test --workspace` | **410 passed, 0 failed**, 12 ignored |
| Build (profile only) | `cargo build --release` — `lto`, `codegen-units = 1`, `panic = "abort"` | success, 2m25s |

CI's *Build WASM Assets* and musl jobs were **not** run. They cannot be affected: `src/adapters` and `src/api` are both `#[cfg(feature = "server")]` (`src/lib.rs:36-41`) and the client is built `--no-default-features --features web`. Stated as a structural argument, not as a claim that those jobs passed.

`cargo clippy --all-targets` was **not** used and is not the bar: `src/lib.rs:20-22` denies `unwrap_used`/`expect_used`/`panic` crate-wide and src's own `#[cfg(test)]` modules violate it, pre-existing.

Both `/hqp/discover` tests **passed** in this local run. Per #392 they are environment-dependent (live UDP SSDP) rather than pre-existing failures; nothing here touches them either way.

## Reviewing this PR: two of its three parents are other people's work

This branch is `merge(#412, #413) + 2 commits`. #413 is not an ancestor of #412, so getting the instrument required merging it — stated because the GitHub diff against `#412` therefore shows **all of #413's 3,314 lines** as well as mine.

```
$ git log --oneline fix/issue-405-roon-browse-errors..fix/issue-416-roon-cross-delivery
0788de1 fix(roon): correlate browse and load results by req_id, not by session key (#416)
fee1f92 test(roon): RED - same-session concurrency cross-delivers browse and load results (#416)
e1cab29 merge: bring in #408's FakeRoonCore as this branch's test instrument (#416)
... (#413's commits below this)
```

**Everything that is mine is `e1cab29..HEAD`** — 788 insertions, 71 deletions, across `src/adapters/roon.rs`, `src/api/mod.rs`, `tests/roon_protocol.rs`, and 25 lines of `tests/mock_servers/roon_core.rs`:

```
$ git diff e1cab29..HEAD
```

The merge itself was clean (`ort`, auto-merged `src/adapters/roon.rs`, no conflicts) and I changed nothing in it. No second Roon mock was written.

**Approving this PR approves neither #412 nor #413.** They need their own approvals, and the merge order is **#412 → #413 → this**. Merging this into `v3` first would carry both in under #416's title and make all three harder to revert. #412 also currently has a `CHANGES_REQUESTED` review (coderabbit, `src/api/mod.rs:578` — route `roon_play_handler` through the same classifier); if it acts on that, this branch needs a re-merge, a text conflict in `src/api/mod.rs` is plausible, and the claim "same shape as #405's `route_browse_rejection`" needs re-checking. The correlation code itself is in a different file and region.

## Correlation approach — `req_id` primary, session key as the guard

`RoonState::route_browse_result` / `route_load_result` (`src/adapters/roon.rs`), both over one generic `route_pending_result`, keyed on `req_id` and cross-checking the session key before resolving. Deliberately the same rule *and the same shape* as #405's `route_browse_rejection`, so the two cannot drift apart again.

**No fork change was needed, and that is the finding this PR turns on.** #412's body states the success arms "cannot be improved without a fork change (no `req_id` on `Parsed::BrowseResult`)". True of `Parsed`; false of the delivery. The fork's event channel is

```rust
type EventReceiver = Receiver<(CoreEvent, Option<(serde_json::Value, Parsed)>)>;
```

and for browse results it forwards the **raw MOO message** alongside the parsed value (fork `src/lib.rs:620-630`). That message carries `request_id` — the fork's own `Browse::parse_msg` reads `msg["request_id"]` off it to look up the session key (`src/browse.rs:157`). UHC was throwing it away one line earlier, in `if let Some((_, parsed)) = msg`. The fix reads it with `moo_request_id(&raw_msg)`.

**Why `req_id` is authoritative.** `Browse::browse`/`load` return the request id UHC stores as its map key (fork `src/browse.rs:126-154`); `Moo::send_req` allocates ids from one monotonic per-connection counter (`src/moo.rs:124-135`); the reply carries it back in its `Request-Id` header, which `Moo::parse` copies in as a string (`src/moo.rs:354-355`). So a `req_id` names exactly one submitted request. Lookup is exact and O(1).

**Why the session key is still checked.** `Moo::new` restarts that counter at 0 on every reconnect (`src/moo.rs:92`), so a result arriving on a fresh connection can carry a `req_id` a stale, not-yet-timed-out entry still occupies. On a mismatch the waiter is **left alone** and falls back to `BROWSE_TIMEOUT` — #405's rationale, preserved verbatim, because being resolved with someone else's *success* is the worst outcome available. `ResultRouting::KeyMismatch` is logged at `warn`; it is also the one path that deliberately does not `remove`, because that entry belongs to a caller still inside its own timeout.

**Each kind searches only its own map**, unlike #405's cross-map search. It has to: `route_browse_rejection` cannot tell a browse from a load, but a `BrowseResult` *cannot* be sent to a `LoadRequest` — the types forbid it. `a_browse_result_never_reaches_a_load_waiting_on_the_same_req_id` pins that, so a future refactor unifying the maps loses the type protection loudly.

**`ResultRouting::Uncorrelated`** covers a result whose message carries no usable `request_id`: it is dropped with a `warn` and the caller times out. Unreachable with the pinned fork — a `BrowseResult` only exists because `parse_msg` already parsed that id — but it is a variant rather than an assumption, because the tempting fallback is the session-key scan this PR removes. A fork that started nulling the raw message for browse, as it already does for images, would otherwise reintroduce the bug silently.

## The red commit

`fee1f92` contains the tests alone. Against it, **both correlation tests fail** — on trial 0 in each of two runs:

```
$ cargo test --test roon_protocol -- concurrent_browses_in_one_session concurrent_loads_in_one_session
running 2 tests
test concurrent_browses_in_one_session_each_receive_their_own_result ... FAILED
test concurrent_loads_in_one_session_each_receive_their_own_page ... FAILED

---- concurrent_browses_in_one_session_each_receive_their_own_result ----
assertion `left == right` failed: trial 0: each caller must receive the list it asked for.
All trials: [["Library","Qobuz","TIDAL"], ["Library","TIDAL","Qobuz"], ["TIDAL","Qobuz","Library"],
             ["TIDAL","Library","Qobuz"], ["TIDAL","Qobuz","Library"], ["Library","TIDAL","Qobuz"],
             ["Qobuz","Library","TIDAL"], ["Library","TIDAL","Qobuz"]]
  left: ["Library", "Qobuz", "TIDAL"]
 right: ["Library", "TIDAL", "Qobuz"]

---- concurrent_loads_in_one_session_each_receive_their_own_page ----
assertion `left == right` failed: trial 0: each caller must receive the page it asked for.
All trials: [[("Artists",25),("Explore",4),("Library",3)], [("Artists",25),("Library",3),("Explore",4)], ...]
  left: [("Artists", 25), ("Explore", 4), ("Library", 3)]
 right: [("Explore", 4), ("Library", 3), ("Artists", 25)]

test result: FAILED. 0 passed; 2 failed
```

After `0788de1`: `test result: ok. 51 passed; 0 failed` for the whole file.

### How often it actually crossed — corrected by this PR's own ship-gate dissent

An earlier version of this section said "8 of 8 trials". That was true of the first run and promoted into a general claim. Counted trial by trial across both captured runs:

| Run | Browse trials crossed | Load trials crossed |
|---|---|---|
| 1 | 8 / 8 | 8 / 8 |
| 2 (quoted above) | 5 / 8 — trials 1, 5 and 7 were **correct by luck** | 8 / 8 |
| **Total** | **13 / 16** | **16 / 16** |

The correction earns its keep twice. It **measures** #416's "7 of 8" estimate at three requests in flight — browse ≈81% crossed, load 100% — and it is the arithmetic that justifies keeping **eight** trials: at browse's ≈19% per-trial success rate a *one*-trial test would have passed against the broken code about one run in five, while eight trials pass only at 0.19⁸ ≈ **1.3 × 10⁻⁶**. The red is reliable; the trial count is load-bearing, not decoration.

The red commit message `fee1f92` keeps the imprecise wording, because amending it needs a force-push and AGENTS.md forbids that. This is the correction of record.

**The load-side result is new evidence, not a restatement.** #416 inferred from the code's shape that `Parsed::LoadResult` shared the defect. It does, and it crossed in every one of 16 trials — more consistently than browse.

**Where the scrambling happened is not an inference either.** The fake answers each request from that request's own body, and its per-request delays fix both the content and the arrival order, so `expected` is exactly what the Core sent, in request order. The fake's own `errors_sent`/request log and `session_levels` assertions hold throughout. Everything reordered above was reordered inside `src/adapters/roon.rs`.

## Tests

**Integration — `tests/roon_protocol.rs`, driving the real event loop through #413's `FakeRoonCore`** (no second mock; the fake was merged in, not reimplemented):

- `concurrent_browses_in_one_session_each_receive_their_own_result` — three browses in flight under one session key, answered in the **reverse** of request order via per-key delays, 8 trials, every caller gets its own list. This is #408's own defect-pinning test, inverted exactly as its comment instructed. The trial count stays at 8 because the old failure came out of `HashMap` iteration order: a single trial got the right answer by luck about one time in eight, so eight clean trials is what distinguishes "exact" from "lucky". **Do not reduce it.**
- `concurrent_loads_in_one_session_each_receive_their_own_page` — the same for `load`, three levels of one session's stack in flight, distinguished by both list title and item count so a cross-delivery cannot hide behind a coincidence.
- `a_shared_session_key_is_one_level_stack_however_well_results_correlate` — the honest limit, below.

**Fake extension (25 lines):** `set_delay_for_load_level`. A `load` request carries no `item_key`, so `set_delay_for_item_key` cannot force out-of-order **load** responses; `LoadOpts::level` is the load-side analogue because it names the content the response will carry. Documented in `tests/mock_servers/README.md` alongside a new "Same-session concurrency (#416)" row.

**Unit — `src/adapters/roon.rs`, 11 new tests** (deterministic; the cases a fake cannot produce on demand): correct waiter among several sharing a key, the same for load, browse-result-cannot-reach-a-load, reconnect key mismatch leaves the entry, unknown `req_id`, double-resolve, result after the caller timed out, dropped receiver still clears the entry, no-`request_id` is left to time out, `moo_request_id` parsing (including that a *numeric* id is not the shape the fork produces), and out-of-order delivery as an invariant of the function itself.

**Order of evidence, stated plainly:** the red commit holds the integration tests only. The unit tests are white-box on functions that did not exist yet, so they could not precede them — which is why the behavioural proof is the one that carries the red claim. #405's suite (18 tests) and its typed `RoonBrowseError` path are untouched and green.

## `pending_images`: shape shared, defect not — checked, documented, not changed

#416 asks for this explicitly. It is **not** the same bug, on two independent grounds, both recorded in a comment at the arms:

1. **There is no `req_id` to read.** For image responses the fork replaces the raw message with `Value::Null` before sending (fork `src/lib.rs:634-639`). Unlike the browse arms, nothing is available to correlate on; fixing these *would* need a fork change.
2. **It cannot cross-deliver anyway.** `Image::get_image` deduplicates in-flight requests by image key (fork `src/image.rs:77-110`): a second request for a key already in flight is handed the **first** request's `req_id` rather than being sent. So `pending_images` cannot hold two entries with one image key, and the scan can only ever match the request that asked for that key. The correlation key here is the *content's* identity — precisely what a session key is not.

**What the dedup does cost, found while checking and deliberately not fixed here:** two concurrent `get_image` calls for one key resolve to the same `req_id`, so the second `pending_images.insert` **evicts the first caller's sender** and that caller gets `"Image request cancelled"` while the second gets the image. Two clients loading the same cover art at once is an ordinary event, so this is reachable. It is a *lost waiter*, not a wrong result, and the fix is different in kind (fan-out per key, touching #405's `route_image_rejection`), so it is out of #416's scope. Recorded rather than silently left; worth its own issue.

## What this does **not** fix — and what #399 must therefore do

`a_shared_session_key_is_one_level_stack_however_well_results_correlate` pins it: a `multi_session_key` names one browse **position** on the Core — a stack `browse` pushes onto and `pop_all` unwinds. Correct correlation guarantees each caller the answer to its own request; it cannot make three callers' pushes onto one stack coherent. After three concurrent browses in one session the stack holds `["Explore","Qobuz","TIDAL","Library"]` and a following `load` pages **Library** — whoever asked for it.

So `/roon/browse`'s note says: results are per-request now; the session's position is still shared; **one navigating client, one session key.** #399's nav handle must *enforce* that — mint one key per client session and keep concurrent use of a single key out of reach — not merely document it. Correlation removes the silent wrong answer, which is what blocked #399; it does not make a shared handle safe.

Scope of that evidence: the level stack is the **fake's** model of Roon's session semantics, derived from the fork and from this repo's own `pop_all` usage, and has never been recorded off a live Core. Read it as "UHC's own model of a session says sharing one is incoherent" — enough to justify the note, not a measurement of Roon.

## Scope decisions

- **`/roon/browse` keeps its caller-supplied `session_key`.** #416 offered removal as an alternative. It is the only way to walk a hierarchy across requests, its value is already returned in the response for clients to continue in, and removing it would be an API change needing explicit approval. The handler gained the note instead (`src/api/mod.rs`, `roon_browse_handler`).
- **`search()`'s one shared key across six requests is left as-is.** Those six run strictly sequentially, so there is never more than one in flight; giving them distinct keys would break the level stack the sequence depends on. #416 lists this as soft.
- **No HTTP route changes.** `tests/fixtures/api_routes.txt` untouched, `api_contract` green, no `api-change-approved` needed.
- **No fork change**, so no `Cargo.lock` bump beyond what #413 already carries.

## Not verified against a live Roon Core

Every claim here rests on `FakeRoonCore`, whose semantics came from the pinned fork and this repo's own adapter. **Green means the adapter now correlates as designed, not that Roon behaves as modelled.** Two things a live rig would settle: that a real Core's browse reply carries the `Request-Id` the fork reads (near-certain — it is how the fork finds its own session keys, so browse would not work at all otherwise), and the level-stack semantics behind the handler note.

An operator with a Core can check the fix directly, with two concurrent browses sharing a key:

```bash
curl -s -XPOST localhost:8088/roon/browse -H 'content-type: application/json' \
     -d '{"pop_all":true,"session_key":"shared"}' | jq '.items[:3]|.[]|{title,item_key}'

( curl -s -XPOST localhost:8088/roon/browse -H 'content-type: application/json' \
       -d '{"session_key":"shared","item_key":"<KEY_A>"}' | jq -c '.list.title' &
  curl -s -XPOST localhost:8088/roon/browse -H 'content-type: application/json' \
       -d '{"session_key":"shared","item_key":"<KEY_B>"}' | jq -c '.list.title' &
  wait )
```

Before: the two titles can be swapped, or both the same. After: each response carries the list its own `item_key` names. **Expected caveat, not a bug:** a *following* `load` in `shared` pages whichever browse the Core processed last — the level-stack limit above. This recipe was written from the handler's request shape and has **not** been run.

## Rollback

`git revert 0788de1` restores the session-key scan and leaves the two integration tests failing — which is the point: the tests, not the prose, are what hold this closed. Nothing persisted changes shape; no config, no migration, no state file.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4607,6 +4607,7 @@ dependencies = [
  "tokio",
  "tokio-stream",
  "tokio-test",
+ "tokio-tungstenite 0.21.0",
  "tokio-util",
  "tower",
  "tower-http",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -151,6 +151,10 @@ serial_test = "3"
 syn = { version = "2", features = ["full", "parsing", "visit"] }
 walkdir = "2"
 tempfile = "3"
+# WebSocket *server* for tests/mock_servers/roon_core.rs (the Roon Core protocol fake).
+# Pinned to the same version roon-api uses as its client, so the fake and the real
+# client library share one framing implementation and no new crate is compiled.
+tokio-tungstenite = "0.21"
 
 [profile.release]
 lto = true
--- a/docs/lyrion.md
+++ b/docs/lyrion.md
@@ -79,8 +79,63 @@ Request specific fields with the `tags` parameter:
 
 ```javascript
 ['status', '-', 1, 'tags:aAdltKc']
-// a=artist, A=album, d=duration, l=album_id, t=tracknum, K=artwork_url, c=coverid
 ```
+
+This adapter's string requests: `a` artist, `A` per-role contributors, `d` duration,
+`l` album title, `t` tracknum, `K` artwork_url, `c` coverid.
+
+**An earlier version of this section was wrong**: it described `l` as the album id and
+`A` as the album. Both are false — `l` is the album title and `e` is the album id —
+and anyone extending the string from that legend would request the wrong fields. The
+table below is read from LMS's own `%tagMap` in
+`Slim/Control/Queries.pm` at 9.1.2 and confirmed against a live server; the recorded
+response is `tests/fixtures/lms/status_tags_aAdltKc.json`.
+
+| Tag | Key in the response | Value |
+|-----|--------------------|-------|
+| `a` | `artist` | Track artist name |
+| `A` | `albumartist`, `trackartist`, `composer`, … | Per-**role** contributor names — one key per role present, not a single field |
+| `s` | `artist_id` | Artist (contributor) id |
+| `l` | `album` | Album **title** |
+| `e` | `album_id` | Album id |
+| `g` | `genre` | Genre name |
+| `p` | `genre_id` | Genre id |
+| `d` | `duration` | Seconds |
+| `t` | `tracknum` | Track number |
+| `y` | `year` | Year |
+| `u` | `url` | Track URL |
+| `c` | `coverid` | Cover id, for `/music/<coverid>/cover.jpg` |
+| `K` | `artwork_url` | Artwork URL, relative for streaming content (see above) |
+| `J` | `artwork_track_id` | The album's artwork track id |
+
+Two things to know before extending the string:
+
+- **`A` returns several keys, not one.** Which keys depends on the roles present on
+  the track, so treat it as a set.
+- **`artwork_track_id` requires `J`, which this adapter does not request.** The
+  artwork fallback chain in `get_player_status` reads `coverid` → `artwork_track_id`
+  → `id`, so its middle arm is currently unreachable and artwork falls through to
+  the track's own `id`. Adding `J` costs nothing on the wire but changes
+  `image_key` for tracks with no `coverid`, and `image_key` is client-visible — so
+  it is deliberately left alone rather than changed as a side effect. See #407.
+
+### `mixer volume` is negative while muted
+
+`status` has **no mute key**. Mute lives in the per-player `mute` pref, and LMS
+*negates* the volume pref while muted, so `status` reports e.g. `"mixer volume": -42`.
+The negation lags the mute by roughly 0.8 s, because `mixer muting` schedules a fade
+and only writes the negated value when it finishes.
+
+The adapter therefore reports `player.volume` as the magnitude, and takes mute from
+two signals OR-ed together:
+
+1. `players 0 100 playerprefs:mute` — the pref, correct immediately. Rides on the
+   `players` call the poll already makes, so mute costs **no extra round-trip**;
+   `mixer muting ?` would cost one per player per poll.
+2. A negative `mixer volume` from `status` — needs no tagged parameter, but lags,
+   so it can only ever produce a false negative.
+
+`mixer muting <0|1>` sets mute, and `mixer volume <n>` clears it as a side effect.
 
 ## Authentication
 
--- a/src/adapters/didl.rs
+++ b/src/adapters/didl.rs
@@ -1,0 +1,286 @@
+//! DIDL-Lite metadata parsing, shared by the OpenHome and UPnP adapters.
+//!
+//! Both backends carry track metadata as DIDL-Lite: OpenHome via its own
+//! metadata event, UPnP via `AVTransport::GetPositionInfo`'s `TrackMetaData`.
+//! In both cases the payload arrives XML-escaped inside another XML document,
+//! so it must be [`html_decode`]d before [`parse_didl_lite`] can see the tags.
+//!
+//! This module exists so the two adapters cannot drift apart in how they read
+//! the same format.
+
+use serde::{Deserialize, Serialize};
+
+/// Track metadata extracted from a DIDL-Lite document.
+#[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq, Default)]
+pub struct TrackInfo {
+    pub title: String,
+    pub artist: String,
+    pub album: String,
+    pub album_art_uri: Option<String>,
+    pub genre: Option<String>,
+}
+
+/// Decode the XML entities that wrap escaped DIDL-Lite payloads.
+///
+/// `&amp;` is replaced **last**, so a double-encoded entity decodes exactly one
+/// level: `&amp;lt;` becomes `&lt;`, not `<`. Replacing it earlier would let the
+/// `&` it produces be re-consumed by a later replacement.
+pub fn html_decode(s: &str) -> String {
+    s.replace("&lt;", "<")
+        .replace("&gt;", ">")
+        .replace("&quot;", "\"")
+        .replace("&apos;", "'")
+        .replace("&amp;", "&")
+}
+
+/// Read the text content of `tag`, tolerating attributes on the opening tag.
+///
+/// Two tolerances, each load-bearing rather than cosmetic:
+///
+/// - **Attributes.** Real renderers emit
+///   `<upnp:albumArtURI dlna:profileID="JPEG_TN">`, and a matcher accepting only
+///   a bare `<tag>` silently returns nothing for the most common form of the one
+///   field album art depends on.
+/// - **Namespace prefix, when `tag` carries none.** SOAP responses may or may
+///   not namespace their elements, so `CurrentVolume` also matches
+///   `<u:CurrentVolume>`. When `tag` *does* carry a prefix (`upnp:album`) the
+///   match is exact, which is what keeps `upnp:album` from swallowing
+///   `upnp:albumArtURI` and what DIDL-Lite callers rely on.
+pub fn extract_xml_value(xml: &str, tag: &str) -> Option<String> {
+    // A prefixed tag is matched verbatim; an unprefixed one may appear with any
+    // namespace prefix.
+    if tag.contains(':') {
+        return extract_exact(xml, tag);
+    }
+    extract_exact(xml, tag).or_else(|| extract_prefixed(xml, tag))
+}
+
+/// Match `<tag ...>value</tag>` with the tag name exactly as given.
+fn extract_exact(xml: &str, tag: &str) -> Option<String> {
+    let end_tag = format!("</{}>", tag);
+    let mut from = 0;
+
+    while let Some(rel) = xml[from..].find(&format!("<{}", tag)) {
+        let open_start = from + rel;
+        let after_name = open_start + 1 + tag.len();
+        let rest = xml.get(after_name..)?;
+
+        // The character after the tag name decides whether this is our tag.
+        match rest.chars().next() {
+            // Exact match, no attributes.
+            Some('>') => {
+                let value_start = after_name + 1;
+                let end = xml[value_start..].find(&end_tag)? + value_start;
+                return Some(xml[value_start..end].to_string());
+            }
+            // Attributes present: skip past them to the end of the opening tag.
+            Some(c) if c.is_whitespace() => {
+                let close_rel = rest.find('>')?;
+                // A self-closing tag carries no text content.
+                if rest[..close_rel].ends_with('/') {
+                    from = after_name + close_rel + 1;
+                    continue;
+                }
+                let value_start = after_name + close_rel + 1;
+                let end = xml[value_start..].find(&end_tag)? + value_start;
+                return Some(xml[value_start..end].to_string());
+            }
+            // A longer tag name that merely starts with ours — keep looking.
+            _ => {
+                from = after_name;
+            }
+        }
+    }
+
+    None
+}
+
+/// Match the same tag carrying any single namespace prefix, e.g. `<u:tag>`.
+fn extract_prefixed(xml: &str, tag: &str) -> Option<String> {
+    let needle = format!(":{}", tag);
+    let mut from = 0;
+
+    while let Some(rel) = xml[from..].find(&needle) {
+        let colon = from + rel;
+        // Walk back over the prefix to the '<' that opens this tag.
+        let open = xml[..colon].rfind('<')?;
+        let prefix = &xml[open + 1..colon];
+        // A prefix is a single name: no '>', no whitespace, no nested '<'.
+        if prefix.is_empty() || prefix.contains(['>', '<', ' ', '\t', '\n', '/']) {
+            from = colon + needle.len();
+            continue;
+        }
+        let qualified = format!("{}:{}", prefix, tag);
+        if let Some(v) = extract_exact(&xml[open..], &qualified) {
+            return Some(v);
+        }
+        from = colon + needle.len();
+    }
+
+    None
+}
+
+/// Parse a decoded DIDL-Lite document into [`TrackInfo`].
+///
+/// Always returns `Some`: absent fields become empty strings or `None`, which
+/// is what both adapters have always relied on.
+pub fn parse_didl_lite(xml: &str) -> Option<TrackInfo> {
+    let title = extract_xml_value(xml, "dc:title")
+        .or_else(|| extract_xml_value(xml, "title"))
+        .unwrap_or_default();
+
+    let artist = extract_xml_value(xml, "upnp:artist")
+        .or_else(|| extract_xml_value(xml, "dc:creator"))
+        .unwrap_or_default();
+
+    let album = extract_xml_value(xml, "upnp:album").unwrap_or_default();
+    let album_art_uri = extract_xml_value(xml, "upnp:albumArtURI");
+    let genre = extract_xml_value(xml, "upnp:genre");
+
+    Some(TrackInfo {
+        title,
+        artist,
+        album,
+        album_art_uri,
+        genre,
+    })
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    /// A realistic DIDL-Lite payload as a renderer returns it, with attributes
+    /// on the tags that carry them in practice.
+    const REAL_DIDL: &str = r#"<DIDL-Lite xmlns="urn:schemas-upnp-org:metadata-1-0/DIDL-Lite/" xmlns:dc="http://purl.org/dc/elements/1.1/" xmlns:upnp="urn:schemas-upnp-org:metadata-1-0/upnp/"><item id="1" parentID="0" restricted="1"><dc:title>Hoppipolla</dc:title><upnp:artist>Sigur Ros</upnp:artist><upnp:album>Takk...</upnp:album><upnp:genre>Post-rock</upnp:genre><upnp:albumArtURI dlna:profileID="JPEG_TN">http://10.0.0.5/art/42.jpg</upnp:albumArtURI></item></DIDL-Lite>"#;
+
+    /// The same payload with a bare albumArtURI tag (no attributes).
+    const BARE_DIDL: &str = r#"<DIDL-Lite><item><dc:title>Hoppipolla</dc:title><upnp:artist>Sigur Ros</upnp:artist><upnp:album>Takk...</upnp:album><upnp:albumArtURI>http://10.0.0.5/art/42.jpg</upnp:albumArtURI></item></DIDL-Lite>"#;
+
+    #[test]
+    fn html_decode_unescapes_in_the_right_order() {
+        // &amp; must be replaced last, or "&amp;lt;" would wrongly become "<".
+        assert_eq!(html_decode("&lt;a&gt;"), "<a>");
+        assert_eq!(html_decode("Simon &amp; Garfunkel"), "Simon & Garfunkel");
+        assert_eq!(html_decode("&quot;x&quot; &apos;y&apos;"), "\"x\" 'y'");
+        // Double-encoded entities decode exactly one level. These fail if
+        // `&amp;` is replaced before `&quot;`/`&apos;`, which it was until
+        // CodeRabbit caught that the code contradicted this function's own doc
+        // comment - and the only prior case (`&amp;lt;`) passed by accident,
+        // because `&lt;` happens to be replaced first.
+        assert_eq!(html_decode("&amp;lt;"), "&lt;");
+        assert_eq!(html_decode("&amp;quot;"), "&quot;");
+        assert_eq!(html_decode("&amp;apos;"), "&apos;");
+        assert_eq!(html_decode("&amp;gt;"), "&gt;");
+    }
+
+    #[test]
+    fn parse_didl_lite_reads_bare_tags() {
+        let t = parse_didl_lite(BARE_DIDL).expect("returns Some");
+        assert_eq!(t.title, "Hoppipolla");
+        assert_eq!(t.artist, "Sigur Ros");
+        assert_eq!(t.album, "Takk...");
+        assert_eq!(
+            t.album_art_uri.as_deref(),
+            Some("http://10.0.0.5/art/42.jpg")
+        );
+    }
+
+    #[test]
+    fn parse_didl_lite_never_returns_none_even_for_garbage() {
+        // Every field defaults rather than failing; both adapters rely on this.
+        let t = parse_didl_lite("not xml at all").expect("returns Some");
+        assert_eq!(t.title, "");
+        assert_eq!(t.artist, "");
+        assert_eq!(t.album, "");
+        assert!(t.album_art_uri.is_none());
+    }
+
+    #[test]
+    fn parse_didl_lite_falls_back_to_dc_creator_and_plain_title() {
+        let xml =
+            r#"<item><title>Bare Title</title><dc:creator>Fallback Artist</dc:creator></item>"#;
+        let t = parse_didl_lite(xml).expect("returns Some");
+        assert_eq!(t.title, "Bare Title");
+        assert_eq!(t.artist, "Fallback Artist");
+    }
+
+    /// Regression for the defect this module was extracted to fix: the previous
+    /// matcher required a literal `<tag>`, so the DLNA-conventional
+    /// `<upnp:albumArtURI dlna:profileID="JPEG_TN">` was invisible and album
+    /// art silently went missing on real devices.
+    #[test]
+    fn parse_didl_lite_reads_tags_that_carry_attributes() {
+        let t = parse_didl_lite(REAL_DIDL).expect("returns Some");
+        assert_eq!(t.title, "Hoppipolla");
+        assert_eq!(t.artist, "Sigur Ros");
+        assert_eq!(t.album, "Takk...");
+        assert_eq!(t.genre.as_deref(), Some("Post-rock"));
+        assert_eq!(
+            t.album_art_uri.as_deref(),
+            Some("http://10.0.0.5/art/42.jpg"),
+            "attribute-bearing albumArtURI must be found"
+        );
+    }
+
+    #[test]
+    fn extract_xml_value_does_not_match_a_longer_tag_with_the_same_prefix() {
+        // `upnp:album` must not swallow `upnp:albumArtURI`.
+        let xml = r#"<item><upnp:albumArtURI>art.jpg</upnp:albumArtURI><upnp:album>Real Album</upnp:album></item>"#;
+        assert_eq!(
+            extract_xml_value(xml, "upnp:album").as_deref(),
+            Some("Real Album")
+        );
+    }
+
+    #[test]
+    fn extract_xml_value_skips_self_closing_tags() {
+        let xml =
+            r#"<item><upnp:albumArtURI /><upnp:albumArtURI>art.jpg</upnp:albumArtURI></item>"#;
+        assert_eq!(
+            extract_xml_value(xml, "upnp:albumArtURI").as_deref(),
+            Some("art.jpg")
+        );
+    }
+
+    #[test]
+    fn extract_xml_value_tolerates_a_namespace_prefix_when_the_tag_has_none() {
+        // SOAP responses may or may not namespace their elements; the UPnP
+        // adapter's own regex extractor allowed either, and consolidating on
+        // this helper must not drop that.
+        let xml = r#"<s:Body><u:GetVolumeResponse><u:CurrentVolume>42</u:CurrentVolume></u:GetVolumeResponse></s:Body>"#;
+        assert_eq!(
+            extract_xml_value(xml, "CurrentVolume").as_deref(),
+            Some("42")
+        );
+        let bare = r#"<Body><CurrentVolume>7</CurrentVolume></Body>"#;
+        assert_eq!(
+            extract_xml_value(bare, "CurrentVolume").as_deref(),
+            Some("7")
+        );
+    }
+
+    #[test]
+    fn extract_xml_value_with_a_prefixed_tag_stays_exact() {
+        // `upnp:album` must not match `upnp:albumArtURI`, and must not match a
+        // differently-prefixed `album` either - DIDL-Lite callers rely on this.
+        let xml = r#"<item><upnp:albumArtURI>art.jpg</upnp:albumArtURI><other:album>Wrong</other:album></item>"#;
+        assert!(extract_xml_value(xml, "upnp:album").is_none());
+    }
+
+    #[test]
+    fn extract_xml_value_prefix_search_ignores_non_prefix_colons() {
+        // A colon inside an attribute value must not be mistaken for a prefix.
+        let xml =
+            r#"<item><link href="http://x/y">no</link><TrackURI>uri://real</TrackURI></item>"#;
+        assert_eq!(
+            extract_xml_value(xml, "TrackURI").as_deref(),
+            Some("uri://real")
+        );
+    }
+
+    #[test]
+    fn extract_xml_value_returns_none_when_absent() {
+        assert!(extract_xml_value("<item><dc:title>x</dc:title></item>", "upnp:album").is_none());
+    }
+}
--- a/src/adapters/lms.rs
+++ b/src/adapters/lms.rs
@@ -67,11 +67,140 @@ use crate::config::{get_config_file_path, read_config_file};
 const LMS_CONFIG_FILE: &str = "lms-config.json";
 /// Request ID for LMS JSON-RPC calls (aids debugging in LMS logs)
 const LMS_REQUEST_ID: i32 = 217;
+/// Name of the per-player LMS pref holding mute state.
+///
+/// `mixer muting ?` and `players … playerprefs:mute` read the same pref -
+/// `mixerQuery` returns `$prefs->client($client)->get("mute")` and
+/// `_addPlayersLoop` emits each requested pref by name - so the batched route
+/// cannot disagree with the per-player one. Verified against Lyrion 9.1.2.
+const LMS_MUTE_PREF: &str = "mute";
 
 /// Strip "lms:" prefix from player IDs.
 /// MCP and aggregator use prefixed IDs (e.g., "lms:00:11:22:33:44:55"), but LMS API expects bare IDs.
 fn strip_lms_prefix(id: &str) -> &str {
     id.strip_prefix("lms:").unwrap_or(id)
+}
+
+/// Read an integer that LMS may emit as either a JSON number or a JSON string.
+///
+/// LMS is not consistent about this, even within one response and even for the
+/// same field across sibling queries. Observed on a single live 9.1.2 server:
+/// `track_id` as a number, `mute` as `"0"`, `mixer volume ?` as `"42"`,
+/// `mixer muting ?` as `1`, `playlist_cur_index` as `"0"`.
+///
+/// `Value::as_i64()` alone silently yields `None` on the string form, which is
+/// how a parse bug becomes an empty result instead of an error (see #407).
+fn lms_i64(value: Option<&Value>) -> Option<i64> {
+    let value = value?;
+    value
+        .as_i64()
+        .or_else(|| value.as_f64().map(|f| f as i64))
+        .or_else(|| value.as_str().and_then(|s| s.trim().parse::<i64>().ok()))
+}
+
+/// Read an LMS 0/1 flag as a bool, tolerating number, string, `null` and absent.
+///
+/// LMS omits a player pref entirely when it was never written and returns
+/// `null` for it from the dedicated query, so "not present" and "not set" both
+/// mean false. Anything non-zero means true.
+fn lms_flag(value: Option<&Value>) -> bool {
+    match value {
+        None | Some(Value::Null) => false,
+        Some(v) => lms_i64(Some(v)).map(|n| n != 0).unwrap_or(false),
+    }
+}
+
+/// Read the entity id from a row of one of LMS's `<type>s_loop` arrays.
+///
+/// LMS keys these `<type>_id` — `album_id` in `albums_loop`, `contributor_id` in
+/// `contributors_loop`, `track_id` in `tracks_loop`. `Slim::Control::Queries`
+/// derives the loop name and the key from the same variable on adjacent lines
+/// (`"${type}s_loop"` / `"${type}_id"`), so a response that contains the loop
+/// necessarily uses that key.
+///
+/// The plain `id` fallback is deliberate belt-and-braces for any LMS version
+/// that might differ; this repo declares no minimum LMS version, so the fix must
+/// not depend on one. Reading only `id` is the #407 defect that made
+/// `search_library()` return an empty vec against every real server.
+fn loop_entity_id(row: &Value, entity: &str) -> Option<i64> {
+    lms_i64(row.get(format!("{}_id", entity)).or_else(|| row.get("id"))).filter(|id| *id > 0)
+}
+
+/// What LMS actually does when a request fails, spelled out once.
+///
+/// `Slim::Web::JSONRPC::requestMethod` calls `closeHTTPSocket()` on any
+/// `$request->isStatusError()`. There is no branch that emits a JSON-RPC `error`
+/// object, so the client sees zero bytes and a closed socket - with no HTTP
+/// status line at all - for every one of: unknown command (104), bad parameters
+/// (102), unknown player id (103), bad server config (105), and Perl exceptions.
+/// Verified against live Lyrion 9.1.2; byte counts in
+/// `tests/fixtures/lms/PROVENANCE.md`.
+///
+/// Because the transport cannot tell these apart, a failure here must not be
+/// read as "this LMS does not support that command". Use `can <verb> ?` for
+/// capability questions, treating `_can: 1` as proof of presence and never
+/// `_can: 0` as proof of absence.
+const LMS_FAILURE_NOTE: &str = "LMS signals every request error this way - \
+unknown command, bad parameters, unknown player id, or bad server config are \
+indistinguishable at the transport, and it never returns a JSON-RPC error \
+object. A network fault looks the same, so do not read this as 'unsupported'.";
+
+/// Turn a `reqwest` transport failure into a message that says which of LMS's
+/// failure modes it actually is.
+///
+/// Worth distinguishing, because they call for different responses: an
+/// unreachable server is a configuration or network problem, whereas a socket
+/// closed *after* the request was accepted is LMS rejecting the request, and the
+/// operator needs to know that "rejected" carries no detail about why.
+fn describe_send_failure(command: &str, error: &reqwest::Error) -> anyhow::Error {
+    if error.is_connect() {
+        return anyhow!(
+            "cannot reach LMS to run `{}`: {}. Check the configured host and port.",
+            command,
+            error
+        );
+    }
+    if error.is_timeout() {
+        return anyhow!(
+            "LMS did not answer `{}` before the request timed out: {}. LMS can \
+             also hang rather than close on some bad-parameter requests, which \
+             looks identical to a slow server.",
+            command,
+            error
+        );
+    }
+    anyhow!(
+        "LMS closed the connection without answering `{}`. {} (transport: {})",
+        command,
+        LMS_FAILURE_NOTE,
+        error
+    )
+}
+
+/// Render a `slim.request` command array for an error message or log line.
+fn describe_command(player_id: Option<&str>, params: &[Value]) -> String {
+    let words: Vec<String> = params
+        .iter()
+        .map(|p| match p {
+            Value::String(s) => s.clone(),
+            other => other.to_string(),
+        })
+        .collect();
+    format!(
+        "{} {}",
+        player_id.unwrap_or("<server>"),
+        truncate_for_log(&words.join(" "))
+    )
+}
+
+/// Keep diagnostics readable when LMS or a stray responder returns something huge.
+fn truncate_for_log(text: &str) -> String {
+    const MAX: usize = 200;
+    if text.chars().count() <= MAX {
+        return text.to_string();
+    }
+    let head: String = text.chars().take(MAX).collect();
+    format!("{}…", head)
 }
 
 /// Saved config for persistence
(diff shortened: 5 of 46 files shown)


muness:
## Gate 1 of 3 — Solution space · **review**

**Gate/role:** solution-space gate, reviewer (self-review per #392's workflow).
**Verdict:** **pass-with-actions** — the chosen approach is right and cheaper than the issue assumed, but two of its four required actions are documentation the issue explicitly asks for, and they were only written after this gate.

### Aim and scope checked

Aim: two concurrent Roon browses that share a `multi_session_key` must each receive their own results, correlated on `req_id` per #416's hard constraint, without regressing #405 and without HTTP route changes. Scope checked against #416's acceptance criteria and #392's constraints (additive, no route changes, no fork bump unless justified).

### Candidates compared

| # | Approach | Verdict |
|---|---|---|
| **A** | **Read `req_id` off the raw MOO message the fork already forwards, correlate on it, keep the session key as #405's reconnect guard** | **Chosen** |
| B | Add `req_id` to `Parsed::BrowseResult`/`LoadResult` in the pinned fork, bump the pin | Rejected |
| C | Serialize browse/load per session key (a lock per key) so first-match-wins is unambiguous | Rejected |
| D | Remove the caller-supplied `session_key` from `/roon/browse` | Rejected |

**A** was chosen once the fork's delivery type was read rather than assumed. #412's body states the success arms "cannot be improved without a fork change (no `req_id` on `Parsed::BrowseResult`)". That is true of the `Parsed` enum and false of the delivery: `type EventReceiver = Receiver<(CoreEvent, Option<(serde_json::Value, Parsed)>)>` (fork `src/lib.rs:29`), and the browse arm forwards the **raw message** next to the parsed value (`src/lib.rs:620-630`). The raw message carries `request_id`; the fork's own `Browse::parse_msg` reads it (`src/browse.rs:157`). UHC discarded it one line into its own loop. So the issue's hard constraint — correlate on `req_id` — was satisfiable at zero dependency cost.

**B** rejected: cross-repo change, a `Cargo.lock` pin bump, and every consumer touched, to obtain information already in hand. Worth revisiting only if the fork is bumped for other reasons, at which point `req_id` on the variant is the tidier shape.

**C** rejected: it makes the *scan* unambiguous without making the *correlation* right, converts a correctness bug into a queueing policy, and turns one slow browse into another caller's 10s stall. #399 says explicitly not to design around the defect. Noted as a possible complement, not a substitute — see the residual risk below.

**D** rejected: the session key is the only way to walk a hierarchy across requests, its value is already returned to clients to continue in, and removal is an API change needing explicit approval. #416 offers "the note **or** the removal"; the note is the answer.

### Evidence inspected

- Pinned fork at `~/.cargo/git/checkouts/rust-roon-api-*/06dd807` (the exact commit in `Cargo.lock`): `src/lib.rs:29` (channel type), `:620-630` (browse forwards `msg`), `:634-639` (image forwards `Value::Null`), `src/browse.rs:87-98` (`BrowseResult` requires `action`, `LoadResult` requires `items`/`offset`/`list`), `:126-154`, `:157`, `src/moo.rs:92`, `:124-135`, `:354-355`, `src/image.rs:77-110`.
- `src/adapters/roon.rs`: the two success arms, `route_browse_rejection` and its rationale comment, `browse()`/`load()`/`get_image()`.
- `src/api/mod.rs:647` (caller-supplied session key), `roon_browse_handler`.
- #412's diff and body; #413's `FakeRoonCore` and its 49-test suite; #399's added hard constraints.

### Findings

- **P2 — the issue's premise that a fork change might be needed was inherited from #412 and is wrong.** Not acting on it would have produced either an unnecessary dependency bump (B) or a workaround (C). *Action: verify against the fork rather than the PR body.* Done; recorded in the PR.
- **P2 — `pending_images` needed a decision, not an assumption.** #416 says "check it". It shares the scan shape and does **not** share the defect, for two independent reasons (no `req_id` available; the fork deduplicates in-flight image requests by key). *Action: document at the arms and in the PR.* Done.
- **P3 — the acceptance criterion "a note on what a caller-supplied session key does and does not guarantee" cannot be written honestly from correlation alone**, because a shared key still means a shared Core-side level stack. *Action: name the residual hazard in the note and back it with a test.* Done.
- **P4 — `search()`'s six-requests-one-key reuse** is safe (strictly sequential) and giving each request its own key would break the level stack the sequence depends on. Left as-is; #416 lists it as soft.

### Contrary case

The strongest case against A is that it makes UHC depend on a delivery detail of the fork — the raw message being forwarded — that no fork API documents, where B would depend on a typed field. That is real. Three things answer it: the dependency is *pinned* to a commit; a break is not silent (`ResultRouting::Uncorrelated` logs at `warn` and the caller times out, rather than falling back to the guessing this fix removes); and the 51-test protocol suite drives every browse, load, search and paging path through this arm, so a change in the fork's delivery fails the suite wholesale rather than subtly. B remains the better shape if the pin ever moves for other reasons.

### Required actions

1. Read the fork before asserting what it cannot do. **Done.**
2. Check `pending_images` and record the outcome either way. **Done.**
3. Write the handler note including the hazard correlation does not remove. **Done.**
4. Use #413's fake rather than a second mock. **Done** (merged, not reimplemented; 25 lines added to it).

### Disposition

Approach A adopted with all four actions closed.

### Residual risk

Correlation makes each answer honest and leaves the shared-level-stack hazard intact. #399's nav handle must therefore *enforce* one key per client session; if it instead wants concurrent navigation through one handle, approach C (serialize per key) is the follow-on, not a substitute for this fix.

### Next gate authorization

**Authorized** to proceed to the execute gate, test-first with a genuine red commit.

muness:
## Gate 1 of 3 — Solution space · **dissent**

**Gate/role:** solution-space gate, dissent (adversarial pass against the review above).
**Verdict:** **pass-with-actions** — approach A survives, but the review's framing of *why* it is safe was wrong in one load-bearing place, and two of its rejections were argued more confidently than the evidence supports.

### What I tried to break

**1. "`req_id` is authoritative" — is it, on the success path specifically?** The review borrowed #405's argument wholesale. #405's argument is about ids the *fork* hands back on rejections. Here the id comes off a raw message, so the chain has an extra link: *is the `request_id` in the forwarded message the same id `browse()` returned?*

**It is, and the review had the weaker proof.** The review argued from "MOO echoes `Request-Id`", which is a claim about Roon. The stronger proof is local and structural: `Browse::parse_msg` returns `Parsed::None` unless `msg["request_id"]` is a key in *its own* `session_keys` map (fork `src/browse.rs:157-162`), and that map is populated with exactly the ids `browse()`/`load()` returned (`:126-154`). **So the mere existence of a `Parsed::BrowseResult` proves the raw message's id is one the fork issued for a browse or load.** It also collapses a whole class of worry: "the id is not in UHC's pending map" can then only mean UHC already removed it — resolved, or timed out and cleaned up — which is precisely `NoWaiter`, and dropping it is correct. **Action: state the structural proof, not the wire-protocol one.** *Done — it is now the argument in the code comment on `moo_request_id` and in the PR.*

**2. Does the fix strictly dominate the old behaviour, or does it trade one loss for another?** A correctness fix that quietly converts some working deliveries into 10s timeouts would be a bad trade. I enumerated every case where the old code delivered and the new one does not:

| Case | Old | New |
|---|---|---|
| one request in flight, id in map | delivered | delivered |
| n in flight sharing a key | **arbitrary waiter** | correct waiter |
| reconnect: stale `req_id` 5/"A", fresh 5/"B", result for "B" | dropped (no "B" in map) → timeout | `KeyMismatch` → timeout |
| reconnect: stale 5/"A", fresh 6/"A", result for 6 | **stale caller resolved, req 6 times out** | req 6 resolved |
| caller already timed out | dropped | `NoWaiter`, dropped |
| no `request_id` (unreachable) | would have scanned | `Uncorrelated` → timeout + `warn` |

No cell gets worse. Two get better. The only behaviour change that is not an improvement is the last row, which cannot occur with the pinned fork. **Finding withdrawn; the dominance claim holds.**

**3. The review's rejection of C (serialize per session key) is too clean.** It says C "makes the scan unambiguous without making the correlation right". True, but it omits that C addresses something A does not: the shared level stack. If #399 wants two clients on one handle, A alone is insufficient and C is the missing half. The review's own residual-risk paragraph says this, then the rejection table reads as if C were simply worse. **Action: say plainly that A and C are answers to different questions, and that #399 needs A now and possibly C later.** *Done — in the PR's "What this does not fix" section and in the handler note.*

**4. `pending_images` — the review's "does not share the defect" is right for a reason it did not state, and it hides a live bug.** Both grounds check out (`src/lib.rs:634-639` nulls the raw message; `src/image.rs:77-110` dedups by key so two pending entries cannot share an image key). But the dedup that makes the scan safe is itself the mechanism of a *different* live fault: two concurrent `get_image` calls for one key get the **same** `req_id`, so UHC's second `pending_images.insert` evicts the first caller's sender and that caller receives `"Image request cancelled"`. Two clients loading the same cover art at once is an ordinary event — the web UI and the iOS app, say. Calling `pending_images` "fine" without naming this would have been a true statement doing false work. **Action: record it explicitly as found-not-fixed, with the reachability argument, and flag it for its own issue.** *Done — comment at the arms, section in the PR, and flagged to the operator.*

**5. Is #416's own framing ("`/roon/browse` lets callers supply the session key … so two concurrent clients cross-deliver") the whole reachability story?** No, and the review repeated it uncritically. `search()` also reuses one key across six requests — the issue says so — but those six are strictly sequential, so they were never at risk; the issue's phrase "the pattern is already load-bearing internally" reads as if internal callers were exposed. They were not. The exposed surface is external HTTP callers and #399. **Action: state reachability precisely rather than inheriting it.** *Done in the PR; the previously-untriggerable-internally point is kept explicit.*

### Findings

- **P2 — the authority argument was the wrong one** (finding 1). Fixed in code comments and PR.
- **P3 — C mischaracterised as strictly worse** (finding 3). Fixed.
- **P2 — a real, reachable image defect was about to be summarised as "no defect"** (finding 4). Recorded, not fixed; out of #416's scope by its own criteria.
- **P3 — inherited reachability claim overstated internal exposure** (finding 5). Corrected.
- **P4 — B (fork change) is left with no follow-up issue.** Deliberate: it is only worth doing when the pin moves for another reason, and filing it now would be noise.

### Contrary case (against my own dissent)

Findings 1, 3 and 5 are about how the reasoning is *stated*, not about what the code does — a reviewer could call them editorial. In an epic whose product is preservation claims backed by evidence, a correct fix resting on a weaker-than-available argument is a real defect: the next agent inherits the argument, not the code. Finding 4 is not editorial by any reading.

### Required actions

All five closed within this PR; none blocks the execute gate.

### Disposition

Approach A confirmed. The correlation argument is now structural, C's role is stated, the image finding is recorded, and reachability is precise.

### Residual risk

The image eviction remains live in `v3` and this PR does not fix it. It is a lost waiter, not a wrong result, and it needs its own issue.

### Next gate authorization

**Authorized** to the execute gate.

muness:
## Gate 2 of 3 — Execute · **review**

**Gate/role:** execute gate, reviewer.
**Verdict:** **pass-with-actions** — the fix is minimal and the red proof is genuine, with one honest asymmetry in the test-first sequence that has to be stated rather than glossed, and one test that the dissent pass then had to harden.

### Aim and scope checked

Aim: `Parsed::BrowseResult` and `Parsed::LoadResult` resolve the request that asked, correlated on `req_id`; #405 untouched and green; `pending_images` checked; no HTTP route changes; a concurrency test that fails against the previous code.

Scope of the change (`git diff e1cab29..HEAD` — everything that is mine): `src/adapters/roon.rs`, `src/api/mod.rs` (doc comments only), `tests/roon_protocol.rs`, 25 lines in `tests/mock_servers/roon_core.rs`, 13 lines in its README. No route file, no schema, no dependency change.

### Evidence inspected

- **Red commit `fee1f92`** — tests alone. `concurrent_browses_in_one_session_each_receive_their_own_result` and `concurrent_loads_in_one_session_each_receive_their_own_page` each failed **8 of 8 trials**, output quoted in the PR body. Both drive the real event loop through `FakeRoonCore`.
- **Green commit `0788de1`** — `cargo test --test roon_protocol`: 51 passed, 0 failed, 1.99s.
- **#405 preservation** — `cargo test --lib adapters::roon`: 29 passed, which is #405's 18 plus #416's 11. `route_browse_rejection`, `RoonBrowseError`, `RoonBrowseErrorKind` and the two image-rejection paths are untouched; `a_rejected_item_key_is_answered_immediately_and_correlated` and `concurrent_browses_with_one_rejected_do_not_disturb_the_others` still green over the wire.
- **Whole workspace** — `cargo test --workspace`: **410 passed, 0 failed**, 12 ignored, after `make css`. `cargo fmt --check` clean, `cargo clippy -- -D warnings` clean.
- Fork sources at the pinned commit, re-read to check every citation in the new comments (`src/lib.rs:29,620-630,634-639`; `src/browse.rs:87-98,126-154,157-162`; `src/moo.rs:92,124-135,354-355`; `src/image.rs:77-110`). Four line references were wrong on first write and were corrected before commit.

### Findings

- **P2 — the test-first sequence is genuine for behaviour and cannot be for the unit tests.** The red commit holds the integration tests only; the 11 unit tests call `route_browse_result`/`route_load_result`, which did not exist in the red commit, so committing them earlier would have produced a *build failure*, not a red test. Stated in the PR and in a comment at the top of the new test section rather than presented as if all 13 tests went red together. The behavioural red is the one that carries the claim, and it is the stronger of the two anyway: it fails through the real event loop and the real wire.
- **P2 — the fix must be proved not to trade correctness for lost deliveries.** Enumerated in the gate-1 dissent: of six reachable cases, two improve, four are unchanged, none regresses. The one behaviour change that is not an improvement (`Uncorrelated`) is unreachable with the pinned fork and is a `warn` plus a timeout rather than a silent guess.
- **P3 — one new test pinned an accident.** `a_shared_session_key_is_one_level_stack_however_well_results_correlate` asserted the exact stack `["Explore","Qobuz","TIDAL","Library"]`, which is whatever the 180/120/60 ms delays scheduled. Found by the dissent pass and rewritten (`e964ab5`) to assert the claim — four levels, root first, all three pushes present, next load pages the last element — which is order-independent and flake-free. The two correlation tests were already order-independent by construction.
- **P3 — the 8-trial count is load-bearing and easy to "tidy" away.** The old failure came out of `HashMap` iteration order, so one trial proves little. A comment in the test file says so and says not to reduce it.
- **P4 — `log_result_routing` is a new function whose only job is logging.** It exists so the five outcomes cannot be logged inconsistently between the browse and load arms, and so `KeyMismatch`/`Uncorrelated` are `warn` rather than invisible. Untested, deliberately; the routing decision it reports is what the tests assert.

### Contrary case

The strongest argument that this gate should *fail*: the unit tests postdate the code they test, so on a strict reading of AGENTS.md's TDD rule ("write the test FIRST, run it and see it FAIL, ONLY THEN fix the code") part of this work is test-after. Against that: the rule's purpose is that the test express the client's expectation rather than the implementation's shape, and the two tests that express the client's expectation — three callers, three answers — were written first, run, and seen to fail 8/8 through the real code path. The unit tests are not a substitute for that; they add the deterministic cases a protocol fake cannot produce on demand (a reconnect id collision, a result after the caller's timeout, a message with no id).

### Required actions

1. State the test-order asymmetry rather than implying uniform test-first. **Done** (PR body; comment in `src/adapters/roon.rs`'s new test section).
2. Correct the four wrong fork line references. **Done** before the fix commit.
3. Harden the order-dependent assertion. **Done** in `e964ab5`.
4. Record the `pending_images` eviction defect found while checking. **Done** (comment at the arms, PR section, flagged to the operator for its own issue).

### Disposition

Execute gate passed with all four actions closed. Nothing in `#405`'s surface changed; nothing in the HTTP contract changed.

### Residual risk

The fix depends on the fork forwarding its raw message for browse results. Pinned, and a break fails 51 tests loudly rather than degrading quietly — but it is a dependency on an undocumented detail, and the `Uncorrelated` variant exists precisely so that a future fork can only make this slow, never wrong.

### Next gate authorization

**Authorized** to the ship gate.

muness:
## Gate 2 of 3 — Execute · **dissent**

**Gate/role:** execute gate, dissent.
**Verdict:** **pass-with-actions.** The fix holds under every attack I could mount, but the review inherited #405's account of *why* the session-key guard is needed, and on this branch's own reading that account is largely unreachable. The guard should stay — for a different reason, which nobody had written down.

### 1. Headline: the reconnect story does not survive contact with the rest of the file

#405's guard rationale, which this PR copied and the review endorsed, is: *`Moo` restarts its request-id counter at 0 on reconnect, so a result on a fresh connection can carry a `req_id` that a stale, not-yet-timed-out entry still occupies.* I tried to construct that state on the success path and could not:

1. **`CoreEvent::Lost` clears all three pending maps** (`src/adapters/roon.rs`, the `Lost` arm) and sets `s.browse = None`; `run_roon_loop` then breaks and clears them again on the way out. So a detected disconnect leaves no stale entries at all, and `browse()` fails fast until the new connection is up.
2. **If a stale entry somehow survived and the new connection reissued its id, `pending_browses.insert` overwrites it** — dropping the old sender, which resolves that caller *immediately* with "Browse request cancelled" rather than leaving it to a timeout. By the time any result arrives, the map holds the new request's session key, and the guard matches. No mismatch.

So on this path, "stale entry + colliding id + result arrives" needs the stale entry to exist *and* to never have been overwritten *and* to have escaped the `Lost` clear. That is not the reconnect case.

**What is reachable:** more than one live connection at a time. The fork keeps `moo_receivers` as a `Vec` and `cores` as a map, and `start_discovery` will register with every Core that answers; each `Moo` has its **own** counter from 0 (`src/moo.rs:92`). UHC's `Registered` arm overwrites `s.browse` with whichever Core registered last, while the earlier connection stays alive and keeps delivering onto the same channel. Two Cores on one LAN — a household with a Nucleus and a test Core, exactly the shape this epic's rig has — is therefore a genuine two-id-space situation, and the guard is the only thing standing between it and a wrong success.

**I have not observed this**, and I am not claiming UHC handles multi-Core correctly in any other respect (it almost certainly does not — one `browse`, one `transport`). The claim is narrower: the guard's stated justification is weak and its real one is stronger, so **keep the guard and fix the comment**.

**Action:** record the reachability honestly at the guard, in both #416's function and as a correction to the account #405 left. *Done — see the follow-up commit; the reconnect case is kept as the *possible* case rather than the asserted one, with the multi-connection case named as the one that does not depend on a race.*

### 2. What I could not break

Stated because a dissent that only lists doubts is not evidence.

- **Does the fix lose any delivery the old code made?** Six reachable cases enumerated (single request; n sharing a key; two reconnect variants; caller already timed out; no `request_id`). Two improve, four unchanged, none regresses. The one changed-but-not-improved case is unreachable with the pinned fork.
- **Can a `BrowseResult` reach a load waiter, or vice versa?** No — the types forbid it, and the fork's classification is not ambiguous either: `BrowseResult` requires `action`, `LoadResult` requires `items`/`offset`/`list` (fork `src/browse.rs:87-98`), and neither body parses as the other. Pinned by `a_browse_result_never_reaches_a_load_waiting_on_the_same_req_id`.
- **Is `expected` in the two correlation tests actually the Core's answer, or is the test asserting its own guess?** The fake's `handle_browse` holds one write lock from before its `levels.push` through `current_list`, so each response carries the level *that request* pushed, whatever order the delays schedule. `handle_load` likewise reads its requested level under the lock. So `expected` is per-request truth and both tests are deterministic in content while free in arrival order — which is what makes "8/8 failed before, 8/8 pass after" mean something. **Verified by reading the fake, not assumed.**
- **`join_all` result order.** Returns results by input index, not completion order, so `got[i]` really is request *i*'s answer.
- **Exactly-once.** Every delivery path removes before sending, under the one state `RwLock`; `KeyMismatch` deliberately does not remove. Double-resolve, post-timeout arrival and dropped-receiver cases each have a test.

### 3. The same eviction shape exists in `pending_browses`, and there it is benign

Gate 1's dissent found `pending_images.insert` evicting a live sender. `pending_browses.insert`/`pending_loads.insert` have the identical shape. The difference is what can collide: image inserts collide whenever two callers want the *same key* (ordinary), while browse inserts collide only on a duplicate `req_id` (needs two id spaces). And the outcome differs too — an evicted browse caller gets "Browse request cancelled" promptly instead of waiting out a 10s timeout, which is better than what it had. Not a defect; recorded so the next reader does not "fix" it by symmetry with the image case.

### 4. Smaller findings

- **P3 — `Uncorrelated` warns once per result.** If a future fork stopped forwarding raw messages, a UI polling `/roon/browse` would emit a `warn` per browse forever. Bounded by each caller's 10s timeout, and loud-and-bounded is the intended trade against silent-and-wrong, but it is a log-flood shape and worth knowing.
- **P3 — the trial count is the test's whole statistical claim and reads like boilerplate.** Someone will "simplify" 8 trials to 1. The comment says not to; that is the only defence available.
- **P4 — `log_result_routing` has no test.** Correct: it maps an already-asserted enum to text. Testing it would pin log strings.
- **P4 — `moo_request_id` rejects a numeric `request_id`.** Deliberate and tested. If a fork bump ever changed the header handling, this returns `Uncorrelated` (timeout + warn) rather than silently mis-parsing.

### Contrary case (against this dissent)

Finding 1 could be read as an argument to *delete* the session-key guard: if the reconnect case is unreachable, the guard is dead weight, and dead weight in a correlation path invites future confusion. I considered it and rejected it. The multi-connection case is reachable on the fork's own structure, the guard costs one string comparison, and the failure it prevents is the worst kind available here — a caller told "success" with another caller's content. #405 wrote "do not remove this check as redundant"; this dissent's conclusion is that the instruction is right and the reason attached to it was not the strongest one.

### Required actions

1. Correct the guard's reachability account in the code, and mark it as a correction of #405's. **Done.**
2. Record why the browse-side insert eviction is not the image-side defect. **Done.**
3. Keep the guard. **Done.**

### Disposition

Execute gate passes. One inherited claim corrected; no code behaviour changed by this dissent (the guard stays exactly as implemented).

### Residual risk

If UHC ever connects to two Cores at once, this PR makes browse *correlation* safe and leaves everything else about multi-Core unaddressed — one `s.browse`, one `s.transport`, zones merged blind. That is out of scope here and is not a new risk, but it is now the case that carries the guard, so somebody should look at it on purpose.

### Next gate authorization

**Authorized** to the ship gate.

muness:
## Gate 3 of 3 — Ship · **review**

**Gate/role:** ship gate, reviewer.
**Verdict:** **pass-with-actions** — locally verifiable, contract-clean, cheap to roll back. The action is that the one thing this cannot prove is a live Roon Core, and that has to be said in the PR rather than left for the operator to discover.

### Aim and scope checked

Ship-readiness of PR #422 as a **stacked, do-not-merge** change: what an operator gets, what proves it, what CI does and does not cover, and how to undo it.

### Evidence inspected

**The gate CI would run, run locally on `bc6ec9b`, after `make css`:**

| CI job | Command | Result |
|---|---|---|
| Lint | `cargo fmt --check` | clean |
| Lint | `cargo clippy -- -D warnings` | clean |
| Test | `cargo test --workspace` | **410 passed, 0 failed**, 12 ignored |
| — | `cargo build --release` (release profile: `lto`, `codegen-units = 1`, `panic = "abort"`) | success |

**Contract:** `tests/api_contract.rs::api_routes_match_contract` green. `git diff origin/v3...HEAD -- tests/fixtures/ src/main.rs` is **empty** — no route added, removed or changed, so no `api-change-approved` label is needed and none was added. `/roon/browse`'s request body and response shape are byte-identical; only its doc comment changed.

**Build jobs not run, with the reason:** CI's *Build WASM Assets* (`dx build … @client --no-default-features --features web`) and the musl/zigbuild jobs were not run here. `src/adapters` and `src/api` are both `#[cfg(feature = "server")]` (`src/lib.rs:36-41`), so neither module is compiled into the WASM client at all and this change cannot affect that artifact. The release build above is the same profile the binary jobs use, on a different target triple. Stated as a structural argument, not as a claim that CI's Build jobs passed.

**Risk surface of what actually ships:** two event-loop arms and one new pure function. `moo_request_id` is one JSON lookup and a `parse`, replacing a linear scan of the pending map — strictly cheaper, O(1) against O(n). No new dependency, no `Cargo.toml`/`Cargo.lock` change of my own (#413 contributes `tokio-tungstenite` as a **dev**-dependency, so nothing new ships in the binary either). No public API or type signature changed: `browse`, `load`, `search`, `play_item`, `search_and_play`, `get_image` all keep their signatures, and `RoonBrowseError` is untouched.

### Findings

- **P2 — no live-Core verification, and the PR must say so.** Every claim here rests on `FakeRoonCore`, whose semantics were derived from the pinned fork and this repo's own adapter. Green means *the adapter now correlates as designed*, not *Roon behaves as modelled*. Two specific things a live rig would settle: that a real Core's browse reply carries the `Request-Id` the fork reads (extremely likely — it is how the fork finds its own session keys, so browsing would not work at all otherwise), and the level-stack semantics behind the `/roon/browse` note. *Action: state both in the PR.* **Done.**
- **P2 — this PR must not be merged before #412.** Its base is `fix/issue-405-roon-browse-errors`; merging it into `v3` first would carry #405's commits in under #416's title and make both harder to revert. *Action: state the order — #412, then this.* **Done.**
- **P3 — #413 also has to land, or this branch's tests do not exist on `v3`.** The branch is `merge(#412, #413) + my commits`. If #413 were rejected, my two integration tests would have to be rewritten against something else, and the unit tests would be all that survived. That is a real dependency, not a nicety. *Action: state it.* **Done.**
- **P3 — operator-visible behaviour change is narrow but real.** Anything today calling `/roon/browse` twice concurrently with one `session_key` was getting scrambled lists and will now get correct ones; the same caller may now see a `warn` in the log where it previously saw silence. No caller loses a response it used to get (six-case enumeration in the gate-2 dissent). *Action: none.*
- **P4 — no migration, no config, no state-file change.** Nothing persisted changes shape.

### Contrary case

An argument for *blocking*: shipping a correlation fix whose only witness is a fake written from the same sources as the code under test is the "mock that nothing asserts against" failure this epic keeps naming. The answer is that the assertion here is not derived from the adapter — it is "three callers asked for three different lists and must each get theirs", which is true of any correct Roon client regardless of what Roon does, and it failed 8/8 before and passes 8/8 after. That is an independent property, not a restatement of the implementation.

### Verification recipe for the operator (needs a live Core; not run here)

```bash
# One session key, two clients. Read two item_keys from the root first:
curl -s -XPOST localhost:8088/roon/browse -H 'content-type: application/json' \
     -d '{"pop_all":true,"session_key":"shared"}' | jq '.items[:3]|.[]|{title,item_key}'

# Then fire two browses at once, both in "shared":
( curl -s -XPOST localhost:8088/roon/browse -H 'content-type: application/json' \
       -d '{"session_key":"shared","item_key":"<KEY_A>"}' | jq -c '.list.title' &
  curl -s -XPOST localhost:8088/roon/browse -H 'content-type: application/json' \
       -d '{"session_key":"shared","item_key":"<KEY_B>"}' | jq -c '.list.title' &
  wait )
```

Before: the two titles can be swapped, or both the same. After: each response carries the list its own `item_key` names. **Expected caveat, not a bug:** a *subsequent* `load` in `shared` pages whichever browse the Core processed last — that is the level-stack limit the handler note describes, and it is why two clients should not share a key.

### Rollback

`git revert bc6ec9b 022b007 e964ab5 0788de1` restores the session-key scan; the two integration tests then fail, which is the intended signal. Reverting `0788de1` alone is enough to restore the old behaviour. Nothing to undeploy, no state to migrate, no config to change.

### Disposition

**Ready to merge, and deliberately not merged.** Awaiting explicit operator approval per AGENTS.md and #392. Merge order: **#412 → #413 → this**. No labels added, no force-push, no `api-change-approved`.

### Residual risk

Unverified against a live Roon Core. The shared-level-stack hazard is documented and tested but not removed — #399 must enforce one key per client. The `pending_images` eviction defect found in passing is live in `v3` and unfixed here.

### Next gate authorization

No further gate. Handing to the operator.

muness:
## Gate 3 of 3 — Ship · **dissent**

**Gate/role:** ship gate, dissent.
**Verdict:** **pass-with-actions**, and one of the actions is against this PR's own body: **the headline evidence claim was overstated.** Corrected below and in the body.

### 1. Headline: "8 of 8 trials" is not what the logs say

The PR body and the red commit message both said both correlation tests "fail in 8 of 8 trials". I recounted the two captured runs trial by trial against `expected`:

| Run | Browse trials crossed | Load trials crossed |
|---|---|---|
| 1 (4-test invocation) | **8 / 8** | **8 / 8** |
| 2 (2-test invocation, quoted in the PR) | **5 / 8** — trials 1, 5 and 7 were *correct by luck* | **8 / 8** |
| Total | **13 / 16** | **16 / 16** |

So: the *test* failed in both runs, on trial 0 both times, and that is the red proof. But the *per-trial* cross rate for browse is 13/16, not 16/16 — three trials in run 2 got the right answer out of `HashMap` iteration order. "8 of 8" was a true statement about run 1 promoted into a general claim, which is exactly the failure #392's standing facts name ("state claims no wider than your evidence") and which two earlier agents in this epic already committed.

**What the corrected numbers actually buy**, which is more than the overstatement did:

- They *measure* the issue's estimate. #416 reported "7 of 8 trials" from #408's two-request version. With three requests: browse ≈ 81% crossed, load 100% in 16 trials.
- They justify the 8-trial design arithmetically. At browse's measured ≈19% per-trial success, a **one**-trial test would have passed against the broken code about one run in five, and an 8-trial test passes only at 0.19^8 ≈ **1.3 × 10⁻⁶**. So the red is reliable and the trial count is not decoration. Anyone tempted to reduce it now has the number.
- The load side genuinely did cross 16/16, so #416's inference that `LoadResult` shared the defect is if anything understated.

**Action:** correct the PR body; state the per-trial rates and the arithmetic. **Done.** The red commit message `fee1f92` keeps its imprecise wording — amending it would need a force-push, which AGENTS.md forbids — so the PR body carries the correction explicitly rather than quietly.

### 2. Approving this PR is not approving its two parents, and the diff makes that easy to blur

This branch is `merge(#412, #413) + 5 commits`, so GitHub's diff against `#412` shows all of #413's 3,314 lines. Two distinct hazards the ship review only half-covered (it named merge *order*, not approval scope):

- **Approval scope.** An operator approving #422 has, at a glance, approved a diff containing #413's `FakeRoonCore` and #412's error routing. Neither should be approved through this PR. **Action: say so in the PR.** *Done.*
- **Coupling to moving targets.** #412 currently carries a `CHANGES_REQUESTED` review from coderabbit, whose finding is in `src/api/mod.rs` at line 578 — about 15 lines above my `BrowseRequest` doc edit and in the same file as my handler note. If #412 acts on it, this branch needs a re-merge and a text conflict in `src/api/mod.rs` is plausible. My correlation code is not affected (different file, different region), but the claim "same shape as #405's `route_browse_rejection`" would need re-checking if #412 restructures that function. **Action: name the coupling and what to re-verify after any re-merge.** *Done.*

### 3. What I could not fault at ship level

- **Contract.** `git diff origin/v3...HEAD -- tests/fixtures/ src/main.rs` is empty; `api_routes_match_contract` green. No label needed, none added.
- **Artifact reach.** `src/adapters` and `src/api` are both `#[cfg(feature = "server")]` (`src/lib.rs:36-41`), and the WASM client is built `--no-default-features --features web`, so this change cannot enter that artifact. That is a structural argument and it is stated as one — the dx build was **not** run.
- **Cost.** A JSON `get` + `parse` replaces a linear scan of the pending map. Strictly cheaper. No allocation on the hot path beyond what existed.
- **Rollback.** Reverting `0788de1` alone restores previous behaviour and turns the two integration tests red — the intended signal. No state, config or persisted-format change to undo.
- **Log volume.** The new `warn` paths fire only where a caller is about to time out, which is already exceptional and already bounded by `BROWSE_TIMEOUT`.

### 4. Smaller findings

- **P3 — the workspace test count is machine-dependent** and the PR states 410/0. The two `/hqp/discover` tests need live UDP SSDP and passed on this Mac; on another they may fail, and a reviewer seeing 408/2 should not read that as this branch. The PR says so; keeping the finding here so the next agent does not re-learn it.
- **P4 — the operator verification recipe is untested.** It needs a live Core, and none was reachable. Written from the handler's own request shape, and labelled as unrun rather than presented as a transcript.
- **P4 — `pending_images`' eviction defect ships unfixed** in `v3`. Documented at the arms, in the PR, and flagged for its own issue. Deliberate: #416's criteria make the image work conditional on it sharing the defect, and it does not.

### Contrary case (against this dissent)

Finding 1 could be dismissed as arithmetic pedantry — the test is red either way, and no reviewer would have been misled into thinking the fix unnecessary. I think that reading is wrong for this epic specifically: the deliverable is "correlation is now exact, and here is the evidence", and an evidence claim inflated from 13/16 to 16/16 is the same species of error as the defect being fixed — reporting something with more confidence than its provenance supports. It also would have destroyed the argument for keeping eight trials, which is the one part of the test a future tidy-up will attack.

### Required actions

1. Correct the per-trial claim in the PR body, with the arithmetic that justifies 8 trials. **Done.**
2. State that approving #422 approves neither #412 nor #413, and give the merge order. **Done.**
3. Name the #412 coupling and what to re-verify after a re-merge. **Done.**

### Disposition

Ship gate passes. No code changed by this dissent; one evidence claim corrected and two process hazards written down.

### Residual risk

Unverified against a live Roon Core. Three open PRs must land in order (#412 → #413 → #422), and #412 is still moving. The image eviction defect is live and unfixed.

### Next gate authorization

None. **Do not merge without explicit operator approval.**

muness:
### Verification ledger, pinned to SHAs

So the evidence is checkable rather than asserted. All with `CARGO_BUILD_RUSTC_WRAPPER=""` (the sccache workaround that actually works, per #392) and `make css` first.

| SHA | Command | Result |
|---|---|---|
| `fee1f92` (red) | `cargo test --test roon_protocol -- concurrent_browses_in_one_session concurrent_loads_in_one_session` | **FAILED**, 0 passed / 2 failed — both on trial 0, twice |
| `0788de1` (fix) | `cargo test --test roon_protocol` | ok, 51 passed / 0 failed, 1.99s |
| `bc6ec9b` (tip) | `cargo fmt --check` | clean |
| `bc6ec9b` | `cargo clippy -- -D warnings` | clean |
| `bc6ec9b` | `cargo test --workspace` | ok, **410 passed / 0 failed**, 12 ignored |
| `bc6ec9b` | `cargo build --release` | success, 2m25s |
| `bc6ec9b` | `cargo test --lib adapters::roon` | ok, 29 passed — #405's 18 plus #416's 11 |
| `bc6ec9b` | `git diff origin/v3...HEAD -- tests/fixtures/ src/main.rs` | empty (no route change) |

The last three doc-only commits (`e964ab5`, `022b007`, `bc6ec9b`) came out of this branch's own review and dissent passes; the workspace suite was re-run at the tip afterwards, with the same 410/0.

Not run, and not claimed: `dx build` / the WASM and musl Build jobs (`src/adapters` and `src/api` are `#[cfg(feature = "server")]`, so they cannot reach the client artifact), and anything against a live Roon Core.

PR #410, opened 2026-07-31
feat(mcp): one structured, capability-honest result envelope for every tool (#395)

muness:
Fixes #395

Parent epic: #392. Realizes #221. OH: 80222d6d

Based on `feat/issue-394-mcp-foundation` (PR #404) at `e877066`, not `v3`.

**Do not merge** — awaiting operator approval per AGENTS.md and #392. No labels added, no force-push.

---

Every tool now answers on two channels. The prose is byte-for-byte what it was. Alongside it rides a structured envelope saying what happened, to what, and — when refused — why, in terms a model can act on.

```
TEXT:     "Action 'playpause' executed.\n\nCurrent state:\n{ ... }"
ENVELOPE: { "outcome": "accepted", "operation": "play_pause",
            "params": { "action": "playpause", "zone_id": "lms:aa:bb:cc:dd:ee:ff" },
            "scope": { "provider": "lms", "zone_name": "Living Room" },
            "observed": { "read_from": "aggregator", "as_of_ms": 1785542637267,
                          "zone": { ..., "state": "paused", ... } }, ... }
```

`accepted`, not `ok`. #221 is ChatGPT looping on `hifi_play` because `"Play Now: Superstition"` states neither outcome nor scope. A model reading `accepted` stops retrying.

## CI is not a gate here, and I am the only one

This PR's base is another feature branch, so **zero CI workflows run on it**. Green checks would prove nothing; there will not even be green checks. Nothing catches a mistake here before it merges up through #404.

So every gating command was run locally. Output pasted, not summarized.

```
$ make css && cargo fmt --check                    # CI Lint step 1
(no output, exit 0)

$ cargo clippy -- -D warnings                      # CI Lint step 2
Finished `dev` profile [unoptimized + debuginfo] target(s) in 1.64s     (exit 0)

$ cargo test --workspace --no-fail-fast            # CI Test job, exactly
mcp_contract           68     lib                  102     adapter_integration    42
protocol_schema        41     zones_sha_integration 20     volume_safety          14
volume_step             9     architecture_lint      8     arch_packaging          6
await_in_lock_lint      4     ignored_send_lint      4     arbitrary_find_lint     3
release_hardening       3     aggregator_lint        2     api_contract            2
main.rs                 2     oneshot_leak_lint      1     spawn_cancellation_lint 1
unbounded_channel_lint  1
client_harness         41     (+1 local-only failure, below)
protocol_integration   32     (+1 local-only failure, below)
[...]

PR #424, opened 2026-08-01
feat(mcp): per-zone capability discovery, and no more silent Roon default (#398)

### LINKED ISSUE #398: MCP capabilities: let clients discover per-zone provider truth before acting

muness:
Fixes #398

Based on `feat/issue-395-mcp-envelope` (PR #410) at `df949e0`, not `v3`.

---

A client can now ask what a zone supports before it tries, and every answer is one of three
things: **supported**, **the provider's protocol cannot** (never retry), or **the provider can and
UHC has not wired it** (retry when the named issue ships).

```
hifi_capabilities {} ->
{ "zones": [ { "zone_id": "lms:aa:bb:cc:dd:ee:ff", "zone_name": "Living Room",
               "provider": "lms", "has_volume_control": true } ],
  "providers": [ { "provider": "lms", "capabilities": [
      { "capability": "transport", "support": "supported" },
      { "capability": "queue_reorder", "support": "not_implemented", "tracked_by": "#400",
        "detail": "playlist move <from> <to> reorders the queue; verified live. Roon cannot do
                   this and LMS can, which is why this capability exists in the vocabulary." } ] } ] }
```

### 1. `hifi_zones` returns a FIFTH zone prefix that MCP routing did not know

`PrefixedZoneId` (`src/bus/events.rs`) lists five valid prefixes, `hqplayer:` among them, and
`HqpAdapter` publishes `ZoneDiscovered` with it (`src/adapters/hqplayer.rs:589-597`). So HQPlayer
zones appear in `hifi_zones` today. `ZoneTarget::classify` had no arm for them, so they landed in
`Unknown`, and `Unknown` folded into Roon: **every `hifi_control` call against an HQPlayer zone
that `hifi_zones` had just listed was sent to the Roon adapter.**

The adapter it should have reached has the whole surface — `play`, `pause`, `stop`, `next`,
`previous`, `seek`, `set_volume`, `volume_up`/`down` (`src/adapters/hqplayer.rs:1304-1374`).

The issue's AC names four accepted prefixes. Shipping a refusal that listed four would tell a
client that an id `hifi_zones` handed it is invalid — a worse lie than the one being fixed. **The
list is five.** HQPlayer capabilities report `not_implemented` tracked by **#328** ("HQPlayer
direct zone: deliver now-playing, transport, seek, and safe volume").

**Not wired here, and this is the most broken thing the issue found.** The zone id is
`hqplayer:<instance_name>`, so routing needs `HqpInstanceManager` resolution, and no fixture or
mock produces an HQPlayer zone. Guessing at that inside a capability-honesty PR is the overreach
this epic keeps punishing. It now carries an honest label instead of a wrong route. A label is not
a fix.

### 2. UPnP `next`/`previous` is refused by UHC, not by the protocol

`src/adapters/upnp.rs` bailed unconditionally on both, with `"Next track not supported by pure UPnP
renderers"`. But `AVTransport:1` — the service this adapter already speaks — **defines `Next` and
`Previous`**. A renderer with no playlist would reject the call, but that is the device's answer to
give, not UHC's to assume.

So `transport` splits into `transport` (play/pause) and `transport_skip` (next/previous), and UPnP's
skip reads `not_implemented`. It is **derived**, not asserted: the adapter's match arm and the
capability layer both read `upnp::REFUSED_TRANSPORT_ACTIONS`, so implementing skip flips the reported
capability automatically. The refusal strings are byte-identical.

## The capability model, and what makes it derived rather than remembered

Chosen over two alternatives. Fields on `hifi_zones` is foreclosed by a hard constraint, not
preference: `hifi_zones`'s text *is* its JSON payload and `mcp_tool_text.json` pins it byte-for-byte,
so any new field changes it. A `capabilities` block on every envelope answers *after* the failure,
which is what this issue replaces. Cost of the tool: an eleventh entry in `tools/list`, and a
zone -> provider join instead of a flat list.

**`Supported` is unforgeable at compile time.** `Support` has three variants; `Gap`, the type of the
hand-written table, has **two** and no way to spell "supported". The only code that can produce
`Support::Supported` is `routed()`, which asks `ZoneTarget::for_transport` / `for_volume` /
`for_library`. A hand-written ✅ — the thing that produced the AGENTS.md error — does not compile.

**And that guarantee is narrower than it sounds.** It says the table cannot disagree with routing.
It does *not* say routing is right: `for_volume()` returning `VolumeRoute::OpenHome` makes
`volume: supported` true by construction whether or not the adapter call works — and this PR adds
that arm. What closes the gap is
`every_supported_capability_reaches_that_providers_own_adapter`, which calls all 15 supported cells
and asserts each response names *that provider's own adapter*. **Read that test before trusting the
type.**

### The rule for `unsupported`, and why it is biased

`unsupported` tells a client never to retry, with no expiry. `not_implemented` costs one wasted call
and self-corrects. The errors are not symmetric, so the rule is not either:

> A cell may be `unsupported` only when the provider's protocol demonstrably lacks it **and the fact
> can be named in one sentence a reader can check.** Otherwise it is `not_implemented`.

Every `unsupported` carries its evidence, and the evidence is emitted to the client in `detail` and
rendered into AGENTS.md as a footnote — so the claim is auditable rather than trusted. That is the
only defence against a *generated* ⛔ being harder to challenge than the hand-written one was.

## The matrix, generated from the tool's own wire payload

```
| Capability | roon | lms | openhome | upnp | hqplayer |
|---|---|---|---|---|---|
| `transport` | ✅ | ✅ | ✅ | ✅ | 🚧 #328 |
| `transport_skip` | ✅ | ✅ | ✅ | 🚧 #392 | 🚧 #328 |
| `volume` | ✅ | ✅ | ✅ | ✅ | 🚧 #328 |
| `search` | ✅ | ✅ | ⛔ | ⛔ | 🚧 #209 |
| `play_by_query` | ✅ | ✅ | ⛔ | ⛔ | 🚧 #209 |
| `play_by_ref` | 🚧 #396 | 🚧 #396 | 🚧 #396 | 🚧 #396 | 🚧 #209 |
| `browse` | 🚧 #399 | 🚧 #402 | ⛔ | ⛔ | 🚧 #209 |
| `queue_read` | 🚧 #400 | 🚧 #400 | 🚧 #392 | ⛔ | 🚧 #209 |
| `queue_jump` | 🚧 #400 | 🚧 #400 | 🚧 #392 | ⛔ | 🚧 #209 |
| `queue_reorder` | ⛔ | 🚧 #400 | 🚧 #392 | ⛔ | 🚧 #209 |
| `queue_remove` | ⛔ | 🚧 #400 | 🚧 #392 | ⛔ | 🚧 #209 |
| `queue_clear` | ⛔ | 🚧 #400 | 🚧 #392 | ⛔ | 🚧 #209 |
| `play_next` | 🚧 #399 | 🚧 #403 | 🚧 #392 | 🚧 #396 | 🚧 #209 |
| `repeat_mode` | 🚧 #360 | 🚧 #403 | 🚧 #392 | 🚧 #392 | 🚧 #209 |
| `shuffle_mode` | 🚧 #360 | 🚧 #403 | 🚧 #392 | 🚧 #392 | 🚧 #209 |
| `saved_playlists` | 🚧 #399 | 🚧 #403 | ⛔ | ⛔ | 🚧 #209 |
| `favorites` | 🚧 #399 | 🚧 #403 | 🚧 #392 | ⛔ | 🚧 #209 |
| `multiroom_sync` | 🚧 #360 | 🚧 #403 | 🚧 #392 | ⛔ | 🚧 #209 |
```

**LMS has no ⛔ anywhere.** That is the issue's third criterion, and
`lms_never_reports_a_uhc_gap_as_a_provider_limitation` asserts it as a set emptiness rather than a
list of cases, so a new capability cannot quietly acquire one. Each LMS gap cites the live Lyrion
9.1.2 check from #402/#403, enforced by `every_lms_gap_cites_live_verification`.

**Roon's three ⛔ are the vocabulary earning its keep:** `queue_reorder` / `queue_remove` /
`queue_clear` are impossible for Roon and pending for LMS. The matrix says LMS is the richer backend
out loud, which is the fact earlier drafts hid.

**`play_by_ref` is uniformly 🚧 #396**, and that is a ship-gate correction, not the first draft. Three
cells (`upnp/play_next`, `upnp/play_by_ref`, `openhome/play_by_ref`) were `⛔` on a conflation this
module's own rule forbids: both renderer protocols *can* play a named URI
(`SetNextAVTransportURI`, `Playlist:1 Insert(AfterId, Uri, Metadata)`) — what is missing is UHC's
ability to name one. Marking that impossible is the exact defect this issue exists to end, committed
three times inside the fix for it. `9e74c1e`, found by the ship dissent.

The AGENTS.md block is rendered from `hifi_capabilities`' **wire payload**, not from
`capabilities.rs`'s internals — checking a doc against a source proves they agree; checking it
against the wire proves the doc says what a client is actually told. `UPDATE_AGENTS_MATRIX=1
cargo test --test mcp_contract agents_md` regenerates it. Two rows the old table had are absent
deliberately, and AGENTS.md says why: **Discovery** is not per-provider, and **Queue building** was
`hifi_play action='queue'` — the same call path as `play_by_query`.

## Behavior changes: seven, each with a test

`hifi_control` ended its action match with `other => other` and routed unknown zone prefixes to Roon.
Both are gone. Enumerated exhaustively, because the bare-id default is load-bearing.

| # | Was | Now | Test |
|---|---|---|---|
| 1 | `openhome:`/`upnp:` volume refused as "not supported for this zone type" | reaches the adapter | `openhome_and_upnp_volume_reaches_the_adapter_since_398` |
| 2 | **bare id -> Roon** (transport, volume, search, play) | refused, naming five prefixes **and the exact repaired id** | `an_unprefixed_zone_id_is_refused_by_name_instead_of_routed_to_roon` |
| 3 | `sonos:` -> Roon for transport/search/play | refused, quoting the prefix it did not recognise | `an_unrecognised_zone_prefix_is_refused_instead_of_routed_to_roon` |
| 4 | `hqplayer:` -> Roon | refused as HQPlayer, `not_implemented`, `tracked_by: "#328"` | `hqplayer_zones_are_recognised_and_reported_as_not_wired` |
| 5 | `openhome:`/`upnp:`/`hqplayer:` searched Roon's library | refused with the reason | `library_routing_reaches_roon_and_lms_only` |
| 6 | unknown action forwarded to the adapter | refused with the valid action list; `operation` is `"unknown_action"`, not the client's string | `an_unknown_hifi_control_action_is_refused_with_the_valid_action_list` |
| 7 | bare id reported `scope.provider: "roon"` | reports `"unknown"` | `an_unplaceable_prefix_reaches_no_adapter_at_all` |

### Why #2 is justified rather than slipped in

- **Every zone id UHC publishes carries a prefix** (`src/bus/events.rs:29-53`), so a bare id can only
  have come from outside `hifi_zones`.
- Bare-id-to-Roon and unknown-prefix-to-Roon were **the same code path**, and #360 names it as an
  anti-pattern to exclude.
- **It is recoverable in one call.** The refusal names the five prefixes *and* the repaired id:
  `"Zone id '1601a5d4bare' has no provider prefix. Accepted prefixes: roon:, lms:, openhome:, upnp:,
  hqplayer:. If this is a Roon zone id, call again with 'roon:1601a5d4bare'. Call hifi_zones for
  valid zone ids."` A client that had a working bare Roon id is told exactly what to send. Asserted,
  not just written.

### The one default kept

**`hifi_search` with no `zone_id` still routes to Roon.** `None` is not a malformed zone id — there
is nothing to route by, LMS `globalsearch` requires a player id, and the tool's own description
documents it. It reports `scope.provider: "roon"`, so it is visible rather than silent, and
`an_absent_zone_id_still_routes_search_to_roon` pins it. That test passed on the test-first commit
and must keep passing.

## `mcp_tool_text.json` is NOT regenerated

#398's criteria require correcting prose #395 froze. Four strings change. Rerunning
`UPDATE_MCP_FIXTURES=1` and letting a green suite imply nothing happened is the wrong way to record
that, so the fixture is **byte-for-byte untouched**:

```
$ git log --follow --oneline tests/fixtures/mcp_tool_text.json
c2cac9e feat(mcp): add a structured, capability-honest result envelope ... (#395)
7cc6959 test(mcp): pin all 10 tools' text before the envelope exists (#395)

$ git diff origin/feat/issue-395-mcp-envelope..HEAD -- tests/fixtures/mcp_tool_text.json
(no output)
```

Instead every change is in `TEXT_CORRECTIONS` with `before` **and** `after`, and the test asserts
each `before` equals the untouched fixture value — so the table cannot launder an unrelated change.
Cases #398 adds go in `TEXT_ADDITIONS`. Every string is accounted for in exactly one of three places
and there is no fourth.

One label is deliberately **not** renamed: `hifi_control/unknown_action_never_reaches_dispatch` was
accurate for #395 (the adapter resolved the device before matching the action) and its premise is now
inverted. Renaming the key would mean rewriting the fixture; the correction records what changed.

`mcp_tools.json` gains one key and loses none:

```
$ git diff --stat origin/feat/issue-395-mcp-envelope..HEAD -- tests/fixtures/mcp_tools.json
 tests/fixtures/mcp_tools.json | 16 ++++++++++++++++
 1 file changed, 16 insertions(+)
```

That is the name-keyed fixture design from #394 doing its job: appending a tool reads as additive,
and every existing key is byte-identical.

## Test-first, and the guardrails were broken to prove they work

`d2adf6a` is a **test-only commit**: ten wire-level tests over the real `/mcp` route, and nine failed
on it. The tenth (`an_absent_zone_id_still_routes_search_to_roon`) passed and had to keep passing.

```
$ git show --stat d2adf6a          # tests/mcp_contract.rs only
$ cargo test --test mcp_contract   # at d2adf6a
test result: FAILED. 69 passed; 9 failed
```

Then each new guardrail was broken and restored:

| Break | Result |
|---|---|
| `Gap::Supported(...)` in the table | **does not compile** — `E0599: no variant named Supported` |
| Delete one `GAPS` row | `GAPS has no entry for [("upnp", "favorites")]` |
| Point `openhome` volume at the `upnp` adapter | `openhome/volume is reported supported, but calling hifi_control produced "Renderer not found"` |
| Hand-edit one AGENTS.md cell | `capability matrix has drifted from what hifi_capabilities reports` |
| Claim LMS browse is a provider limit | 3 tests fail, naming `["browse"]` |

The wrong-adapter break is the important one: it passes the type check and fails the end-to-end test,
which is exactly the boundary the module docs claim.

## CI is not a gate here, and I am the only one

This PR's base is another feature branch, so **zero CI workflows run on it**. There will not even be
green checks to misread. Every gating command was run locally, output pasted rather than summarized.

```
$ make css && cargo fmt --check                    # CI Lint step 1
(no output, exit 0)

$ cargo clippy -- -D warnings                      # CI Lint step 2
Finished `dev` profile [unoptimized + debuginfo] target(s) in 9.23s     (exit 0)

$ cargo test --workspace --no-fail-fast            # CI Test job, exactly
lib                    116    mcp_contract           78     client_harness         42
adapter_integration     42    protocol_schema        41     protocol_integration   33
zones_sha_integration   20    volume_safety          14     volume_step             9
architecture_lint        8    arch_packaging          6     release_hardening       3
arbitrary_find_lint      3    await_in_lock_lint      4     ignored_send_lint       4
aggregator_lint          2    api_contract            2     main.rs                 2
oneshot_leak_lint        1    spawn_cancellation_lint 1     unbounded_channel_lint  1
Doc-tests: 0 passed; 13 ignored
ALL GREEN. 0 failed across every suite.
```

**No local-only failures this time.** #395 reported `/hqp/discover` failing on this host for want of
UDP SSDP; both tests pass here, in `client_harness` (42/42) and `protocol_integration` (33/33). I am
not claiming the environment changed — only that this run is green including them, which is a
narrower claim than "those failures are gone".

WASM, CI's `build.yml:500` command:

```
$ CARGO_BUILD_RUSTC_WRAPPER="" dx build --fullstack --release @client --no-default-features \
    --features web @server --features server
139.88s  INFO Server build completed successfully! 🚀

$ CARGO_BUILD_RUSTC_WRAPPER="" dx build --release --platform web --features web
 39.49s  INFO Client build completed successfully! 🚀
```

### A third environment fault, for whoever hits it next

The client build failed twice with `Missing rust target wasm32-unknown-unknown` **while
`rustup target list --installed` showed it present.** The cause: `which rustc` resolves to
`/opt/homebrew/bin/rustc` (1.93.1), which has no wasm target, and Homebrew wins even with
`~/.cargo/bin` prepended because rustup's shims are not there — the real binary is under
`~/.rustup/toolchains/1.97.0-aarch64-apple-darwin/bin`. Prepending *that* plus
`rustup target add wasm32-unknown-unknown --toolchain 1.97.0-aarch64-apple-darwin` fixed it. This is
the third distinct local-toolchain trap in this epic after sccache and `make css`.

## A mistake I made, and what it cost

Undoing one of the breakage experiments with `git checkout src/mcp/tools/transport.rs` reverted the
file **past this branch's own uncommitted work** to the #395 version, and `0452bcb` captured that —
so it contains a commit message describing code that is not in it. `cc8afa9` restores both
`transport.rs` and the AGENTS.md matrix block and re-verifies all five guardrails. Flagged rather
than rebased away: the history is squash-merged, and hiding a mistake in the one PR whose subject is
capability honesty would be a poor trade. Review `cc8afa9`'s diff of `transport.rs` as the real
change; `0452bcb`'s is incomplete.

## Residual risk, stated plainly

**Every ⛔ for OpenHome and UPnP is specification knowledge, not a call to a device.** Twelve of the
fourteen provider-limitation cells are theirs, and each rests on my reading of `AVTransport:1`,
`RenderingControl:1` and the `av-openhome-org` service set. I have neither device; the repo's mocks
answer SSDP-shaped HTTP and nothing more. LMS's cells are the strongest here — they come from a live
Lyrion 9.1.2 inventory. Every such cell says "Verified from the ... service definitions, not from a
device" in its own `detail`, so a client reads the provenance, not just the verdict.

**The highest-value thing an operator could do to this PR** is point one real OpenHome or UPnP
renderer at it and check the ⛔ cells. Ten minutes, and it is the difference between "UHC believes"
and "the device cannot".

**`volume: supported` will be wrong for some Roon zones.** A fixed-volume output has no volume
control, and this reports `supported` for every Roon zone — derived from a real routing fact and false
for that zone. UHC cannot derive better: `volume_control: None` conflates "no volume control" with
"not read yet". The payload carries `has_volume_control` beside the capability so a client can weigh
both, and the ambiguity is documented on the field. **Notice what that actually is: a preference for a
stable wrong answer over an unstable one.** A fourth value (`device_dependent`) is the defensible
alternative, and I did not take it because the issue mandates three states — reasoning about the spec,
not about the client. It is cheap to reverse now and expensive once clients depend on three.

## The four gate reports are in the comments

Two per gate, review and dissent, per #392. They are worth reading in this order, because each found
something the previous one shipped:

- **Solution-space** — three approaches compared; found `hqplayer:` and the UPnP skip mislabel, both of
  which amended this issue's acceptance criteria.
- **Solution-space dissent** — named the `volume: supported` problem for fixed-volume Roon outputs, and
  that every OpenHome/UPnP `⛔` is specification knowledge.
- **Execute** — the five broken-and-restored guardrails; and my own `git checkout` mistake.
- **Execute dissent** — found the `value as i32` truncation bug (`27bea38`), and that
  `an_unprefixed...` nearly shipped an assertion that passed for the wrong reason.
- **Ship** — verification, rollback, merge-order warning.
- **Ship dissent** — found the three mislabelled `⛔` cells (`9e74c1e`), and retracts the weight I put
  on "a hand-written ✅ does not compile": it guards 15 cells, and the risk lives in the other 75.

Two real defects in this PR were found by its own dissents and by nothing else. That is the strongest
argument I can make for the two-report format.

## Rollback

Revert the three commits, or drop `src/mcp/capabilities.rs` and `src/mcp/tools/capabilities.rs` and
the `tool_box!` entry. Nothing outside `src/mcp` and `AGENTS.md` changes except
`src/adapters/upnp.rs`, where the change is one const plus a match arm that produces byte-identical
strings. No HTTP routes touched; `tests/fixtures/api_routes.txt` untouched.


muness:
## Gate report — solution space (#398 CAPABILITIES)

**Gate/role:** solution-space, acting as designer. No implementation code yet.

**Verdict: pass-with-actions.** One approach selected. Two findings change the issue's own
acceptance criteria, and both are the *same failure mode* the issue exists to end.

### Aim and scope checked

Per-zone capability discovery with three states, derived not hand-maintained; explicit
routing instead of a silent Roon default; AGENTS.md's matrix checked against the derived
data. Additive: existing tools keep working for zone ids that work today.

---

## Finding 1 (P1) — `hifi_zones` publishes a FIFTH zone prefix that MCP routing does not know

`src/bus/events.rs:59` is the canonical prefix list and it has five entries:

```rust
let valid_prefixes = ["roon:", "lms:", "openhome:", "upnp:", "hqplayer:"];
```

`src/adapters/hqplayer.rs:589-597` publishes `BusEvent::ZoneDiscovered` with
`PrefixedZoneId::hqplayer(...)`, so **`hqplayer:<instance>` zones reach the aggregator and
appear in `hifi_zones` today.** `ZoneTarget::classify` has no arm for it, so it lands in
`Unknown`, and `for_transport()` folds `Unknown` into Roon — meaning **every `hifi_control`
call against an HQPlayer zone that `hifi_zones` just listed is sent to Roon.**

The adapter it should have reached has the full surface: `play`, `pause`, `stop`, `next`,
`previous`, `set_volume`, `volume_up`, `volume_down`, `seek`, `control(action)`
(`src/adapters/hqplayer.rs:1304-1374`).

So this is the OpenHome/UPnP-volume defect again, one level worse: not one capability
declined on a recognised zone, but an entire recognised zone type routed to the wrong
backend. The issue's AC lists four accepted prefixes; shipping a refusal that names four
would tell a client that a zone id `hifi_zones` returned is invalid. **Consequence: the
accepted-prefix list must be five, and HQPlayer must be a column in the matrix.**

Not wiring HQPlayer transport here: the zone id is `hqplayer:<instance_name>`, so routing
needs `HqpInstanceManager::get_or_create` resolution and there is no HQPlayer zone in any
test fixture. Reported as `not_implemented` tracked by **#328** ("HQPlayer direct zone:
deliver now-playing, transport, seek, and safe volume"), which is a real open issue.

## Finding 2 (P2) — UPnP `next`/`previous` is refused by UHC, not by the protocol

`src/adapters/upnp.rs:783-788` bails unconditionally:

```rust
"next" => { anyhow::bail!("Next track not supported by pure UPnP renderers"); }
"previous" | "prev" => { anyhow::bail!("Previous track not supported by pure UPnP renderers"); }
```

`AVTransport:1` — the service this adapter already speaks, `src/adapters/upnp.rs:30` —
defines `Next` and `Previous` actions. So the protocol supports it and UHC declines to
issue it. Reporting UPnP transport as flatly `supported` would lie to a client that then
calls `next`; reporting the skip half as `unsupported` would repeat the mislabelling.
**Consequence: `transport` splits into `transport` (play/pause) and `transport_skip`
(next/previous), and UPnP's skip reads `not_implemented`.**

---

## Approaches compared

### A — fields on `hifi_zones`

Add `provider` + `capabilities` to `McpZone`.

**Rejected on a hard constraint, not on taste.** `tests/fixtures/mcp_tool_text.json` pins
`hifi_zones`'s text byte-for-byte and #395's guarantee is that the text is never
substituted. `hifi_zones`'s text *is* its JSON payload, so any new field changes it. It
also pays the capability cost on every zone listing, where most callers want ids.

### B — a `capabilities` block inside every envelope

Attach the acting zone's capabilities to `scope` on every call.

**Rejected.** It answers after the fact — the client learns the limit in the same response
that already refused it, which is what #398 exists to replace. And it multiplies ~18 values
onto every tool result, including the 9 that do not need them.

### C — a new read-only tool, `hifi_capabilities` (**selected**)

`zone_id` optional. Absent: every zone the aggregator holds, plus the capability table for
each provider present. Given: that zone, plus its provider's table.

**Why it wins:** it is the only option that is additive under the pinned-text constraint
(a new tool adds a key to `mcp_tools.json` and leaves every existing key byte-identical);
it is *pre*-flight, which is the point; and it costs nothing on the calls that do not ask.

**What it costs, stated plainly:** an eleventh tool in `tools/list`, i.e. more surface for
a model to weigh, and a join — zone -> provider -> capability table — instead of one flat
per-zone list. The join is deliberate: capability state is a function of provider, not of
zone (see the caveat below), so inlining it per zone would duplicate identical data and
imply a per-zone precision UHC does not have.

---

## The capability model

**Three states, and one variant carries a reference:**

```rust
enum Support {
    Supported,                                 // a working MCP call path exists
    Unsupported { evidence: &'static str },    // the provider's protocol cannot; never retry
    NotImplemented { tracked_by: &'static str, evidence: &'static str },  // it can; UHC has not
}
```

**`Supported` is unforgeable, at compile time.** The hand-written half of the table is typed
`Gap`, an enum with exactly two variants (`ProviderCannot`, `NotWired`) and no way to spell
`Supported`. The only code that can produce `Support::Supported` is the routing layer
(`ZoneTarget::for_transport` / `for_volume` / `for_library`). A hand-written ✅ — the thing
that produced the AGENTS.md error — is not a discipline here, it is a type error.

**Vocabulary (18), and it is not Roon-shaped:** transport, transport_skip, volume, search,
play_by_query, play_by_ref, browse, queue_read, queue_jump, queue_reorder, queue_remove,
queue_clear, play_next, repeat_mode, shuffle_mode, saved_playlists, favorites,
multiroom_sync. Every value LMS has beyond Roon is in it: queue_reorder / queue_remove /
queue_clear are `unsupported` for Roon (its public API offers read plus `play_from_here`)
and `not_implemented` for LMS. The matrix therefore says LMS > Roon out loud.

**The rule for `unsupported`, applied mechanically:** a claim of `unsupported` requires a
protocol fact I can name in one sentence and a reader can check — a service, an action, or
an absence of one. If I cannot name it, the value is `not_implemented`. A false "never" is
worse than a false "not yet": "not yet" costs a client one wasted call, "never" costs it
the capability permanently. Every `Unsupported` carries its evidence string, and the
evidence is emitted to the client in `detail`, so the claim is auditable rather than
trusted.

**Caveat, recorded rather than papered over:** capability state is per *provider*, not per
device. A fixed-volume Roon output reports `volume: supported`. UHC cannot do better today
— the aggregator's `volume_control: None` conflates "this output has no volume control"
with "no volume has been read yet" — and inventing per-device precision from that field
would produce exactly the confident-and-wrong answer this issue is about. Stated in the
tool description, the module docs, and AGENTS.md.

---

## Behavior changes (each gets a test; full enumeration in the PR)

The Roon default for bare ids is load-bearing and its removal is justified, not slipped in:

- Every zone id UHC publishes is prefixed (`src/bus/events.rs:29-53`), so a bare id can
  only come from outside `hifi_zones`.
- Bare-id-to-Roon and unknown-prefix-to-Roon are the same code path, and #360 names it as
  an anti-pattern to exclude.
- The refusal is recoverable in one step: it names the five prefixes and, when the
  aggregator holds exactly one zone whose unprefixed id matches, names that exact zone id.
  So a client holding a real bare Roon id is told to call `roon:<id>` rather than left to
  guess.

`hifi_search` with **no** `zone_id` keeps routing to Roon — unchanged. `None` is not a zone
id, LMS `globalsearch` requires a player, and the tool's own description documents the
default. Pinned by a test so the additive claim covers it.

---

## Required actions before implementation

1. `ACCEPTED_ZONE_PREFIXES` becomes five entries, `hqplayer:` included, or the refusal
   contradicts `hifi_zones`.
2. `Gap` must not be able to express `Supported`. Verify by trying to write it.
3. Do not regenerate `mcp_tool_text.json`. Every text change goes in a `TEXT_CORRECTIONS`
   table carrying `before` **and** `after`, with a test asserting `before` equals the
   untouched fixture value. The fixture's `git log --follow` must stay at two commits.
4. Bias `unsupported` -> `not_implemented` wherever the protocol fact cannot be named.

**Disposition:** approach C selected. Two AC corrections required (five prefixes, split
transport). No merge sought, no labels added.

**Residual risk:** the protocol facts behind every `unsupported` cell for OpenHome and UPnP
are from specification knowledge, not from a live device — see the dissent.

**Next gate:** execute, test-first.

muness:
### Dissent — solution space (#398)

Aimed where the brief says to aim it: **a capability value that is technically derived and
still lies to a client.** In the order I would worry.

**1. `unsupported` for OpenHome and UPnP is specification knowledge, not observed
behavior — and it is the one state a client is told never to retry.**

Twelve of the fourteen `unsupported` cells in my table are OpenHome or UPnP. Every one rests
on my reading of `AVTransport:1` / `RenderingControl:1` / the `av-openhome-org` service set,
not on a call to a device. I have no OpenHome device and no UPnP renderer; the repo's mocks
answer SSDP-shaped HTTP and nothing more, and the existing test says so
(`openhome_and_upnp_prefixes_reach_their_own_adapters`: "a mock cannot be injected without a
discovery seam"). Compare the LMS cells, which are backed by #402/#403's live Lyrion 9.1.2
inventory — those I would defend line by line.

**Cost if I am trusted:** a client permanently stops asking for something a real device
does. The asymmetry is the whole problem — `not_implemented` costs one wasted call and
self-corrects when the issue ships; `unsupported` is a dead end with no expiry. The
mitigation I have chosen (name the fact or downgrade to `not_implemented`) reduces the count
but does not change its epistemic status. **What I would actually want: every OpenHome/UPnP
`unsupported` cell re-verified against one real device before this matrix is treated as
authoritative, and until then read as "UHC believes" rather than "the device cannot".** If
the operator has a Linn or a WiiM on the rig, that is a ten-minute check and it is the
highest-value thing anyone could do to this PR.

**2. `volume: supported` will be wrong for some Roon zones, and I am shipping it anyway.**

A fixed-volume Roon output — an endpoint feeding a preamp — has no volume control. My table
says `supported` for every Roon zone. That is *precisely* "technically derived but lies to a
client": derived from a real routing fact (the MCP volume path does reach the Roon adapter),
and false for that zone.

I rejected fixing it because the aggregator's `volume_control: None` conflates "no volume
control" with "not read yet", so deriving from it would mislabel every freshly discovered
zone. But notice what I have actually done: I preferred a **stable** wrong answer to an
**unstable** one, and dressed it as caution. A client cannot tell the difference. The
defensible alternative I did not take is a fourth value — `unknown` / `device_dependent` —
for exactly the cells where UHC's precision runs out. I did not take it because the issue
mandates three states and a fourth invites collapsing back to two. **That reasoning is about
the spec, not about the client**, and the client is who gets the wrong answer. If the
operator would rather have `device_dependent` on Roon volume than a confident `supported`,
that is a defensible reversal and it is cheap now, expensive after clients depend on three.

**3. Splitting `transport` is a coverage claim, and the split I chose is the one I could
test — not the one the surface has.**

`transport_skip` exists because I found UPnP's bail. But `hifi_control` also accepts `stop`
implicitly (no — it does not; `stop` is not in the action map at all, though three of four
adapters implement it), and OpenHome's `control` accepts actions MCP never sends. So the
real shape is not two capabilities, it is one per action, and I picked the one split that a
single grep justified. **Cost if trusted:** `transport: supported` still covers a set of
actions I have not enumerated per provider, so the next person to find an action one adapter
refuses will find `transport_skip` sitting there as precedent for a third ad-hoc split
rather than a principle. The principled version is a per-action table derived from each
adapter's own match arms — which is real work and which I am not doing here.

**4. My "`Supported` is unforgeable" claim is narrower than it sounds.**

The type system stops me writing `Supported` in the gap table. It does **not** stop the
routing layer from being wrong — `for_volume()` returning `VolumeRoute::OpenHome` makes
`volume: supported` true by construction whether or not the adapter call works. I am adding
that arm in this PR. So the guarantee is "the table cannot disagree with routing", not "the
table is correct". The end-to-end test (every `Supported` cell must produce that adapter's
own error text offline) is what closes the remaining gap, and it is the test to read first
if you are auditing this PR — not the type.

**5. The AGENTS.md drift test proves agreement, not truth.**

Generating the matrix from `capabilities.rs` guarantees the doc and the code say the same
thing. If the code is wrong, the test makes the doc wrong *consistently* and removes the
friction that might have made someone notice. #398 exists because a hand-written ❌ went
unchallenged; a generated ❌ is harder to challenge, not easier. The mitigation is that
every non-`Supported` cell must carry evidence a reader can check, and the footnotes render
into AGENTS.md rather than staying in the source — so the doc argues its case instead of
asserting it. I think that is enough. I am not certain it is.

**6. Not wiring HQPlayer is the right call and it leaves the worse bug live.**

OpenHome/UPnP volume gets fixed here. HQPlayer zones — a whole zone type `hifi_zones`
advertises — keep being routed to Roon, now with an honest label on it. A label is not a
fix. I chose it because I cannot test HQPlayer zone routing (no fixture, no mock zone) and
guessing at `HqpInstanceManager` resolution in a capability-honesty PR would be the exact
overreach this epic keeps punishing. But the operator should know that the most broken thing
this issue found is the thing it is not fixing, and that #328 is where it goes.

muness:
## Gate report — execute (#398 CAPABILITIES)

**Gate/role:** execute, acting as reviewer of my own implementation.

**Verdict: pass-with-actions.** Test-first held. Every guardrail was broken and restored. One
process failure of mine is recorded rather than rebased away, and one required action remains for
the operator.

### Aim and scope checked

Per-zone capability discovery in three states, derived from routing rather than hand-maintained;
explicit routing in place of the silent Roon default; AGENTS.md's matrix generated with a drift
test. Additive: existing tools keep working for zone ids that work today.

### Evidence inspected

| What | Evidence |
|---|---|
| Test-first | `d2adf6a` touches `tests/mcp_contract.rs` only. `cargo test --test mcp_contract` at that commit: **69 passed, 9 failed** — every new assertion failing, plus one passing that had to keep passing (`an_absent_zone_id_still_routes_search_to_roon`). |
| `Supported` unforgeable | `Gap::Supported(...)` -> `E0599: no variant or associated item named Supported found for enum Gap`. Compile-time, not a test. |
| Table completeness | Deleting `(Upnp, Favorites)` -> `GAPS has no entry for [("upnp", "favorites")]`. Both directions checked: a routed cell left in the table also fails. |
| Call paths real | `openhome` volume pointed at `state.upnp` -> `openhome/volume is reported supported, but calling hifi_control produced "Renderer not found"`. Passes the type check, fails the wire test. |
| Doc drift | One hand-edited AGENTS.md cell -> `capability matrix has drifted from what hifi_capabilities reports`. |
| LMS honesty | Marking LMS browse `ProviderCannot` -> 3 tests fail, naming `["browse"]`. |
| Fixture untouched | `git diff origin/feat/issue-395-mcp-envelope..HEAD -- tests/fixtures/mcp_tool_text.json` -> empty. `git log --follow` still shows the two #395 commits. |
| Additive `tools/list` | `mcp_tools.json`: **16 insertions, 0 deletions.** |
| Full gate | `cargo fmt --check` clean; `cargo clippy -- -D warnings` clean; `cargo test --workspace --no-fail-fast` all green, 0 failures across 22 suites; both `dx build` commands succeed. |

### Findings

**P1 — I committed a file I had reverted with `git checkout`, and the commit message describes code
that is not in it.** Undoing breakage experiment 2 with `git checkout src/mcp/tools/transport.rs`
reverted the file past this branch's own uncommitted work to the #395 version; the same happened to
AGENTS.md. `0452bcb` captured that. `cc8afa9` restores both and re-verifies all five guardrails.
Recorded in the PR body rather than rebased away — the history is squash-merged, and hiding a
mistake in the one PR about capability honesty is a bad trade. **Reviewer instruction: read
`cc8afa9`'s `transport.rs` diff as the real change.**

The generalisable lesson, since it cost real time: **never use `git checkout <file>` to undo a
deliberate experiment on a branch with uncommitted work.** Reverse the edit, or commit first. Two
files were silently rolled back and only one failed loudly.

**P2 — `every_supported_capability_reaches_that_providers_own_adapter` asserts an exact count (15),
not a floor.** A floor would pass while a cell silently stopped being reported as supported, which
is the direction that *hides* a capability. Deliberately wiring or unwiring one now requires
changing that number in the same commit.

**P2 — The `not_implemented` refusal has exactly one producer left.** OpenHome/UPnP volume used to
be it, and wiring it removed the only path. `every_refusal_reason_is_actually_produced` caught that
immediately, and the coverage was restored with an `hqplayer:` case rather than by relaxing the
assertion — which is the right direction, but it means one zone type carries the whole
`not_implemented` path's coverage. If #328 lands before another gap appears, that test needs a new
case.

**P3 — Two rows disappeared from the AGENTS.md table.** "Discovery" and "Queue building" are not in
the vocabulary. Both are explained in AGENTS.md as deliberate: discovery is not per-provider, and
queue building is `hifi_play action='queue'` — the same call path as `play_by_query`. Stated because
a shrinking doc table otherwise reads as lost information.

**P3 — `tracked_by` references were corrected mid-implementation.** OpenHome's queue cells first
pointed at #400, whose title scopes it to Roon and LMS — pointing OpenHome at it would have promised
delivery it does not cover. They now point at #392. Roon's `play_next` moved from #400 to #399,
since Roon's Play Next is a browse item action. Small, and exactly the kind of unchecked reference
that makes "not yet" unfalsifiable.

**P4 — `unreachable_refused()` returns an `anyhow` error rather than panicking.** `src/lib.rs` denies
`clippy::panic` crate-wide, and a diagnosable message beats a crashed request. Its arms are provably
unreachable (the `Refused` variants are handled before dispatch), but kept exhaustive so a new route
variant fails to compile rather than falling through.

### Contrary case

The strongest argument against this implementation is that **it added a capability
(`transport_skip`) on the strength of a single grep.** UPnP's bail is real, but the honest shape is
one capability per action — `stop` is implemented by three adapters and is not a `hifi_control`
action at all — and I split exactly the one asymmetry I happened to find. The next person to find a
per-action difference will see `transport_skip` as precedent for a third ad-hoc split rather than a
principle. The principled version derives the per-action table from each adapter's own match arms,
and that is real work I did not do.

The second argument: **`hifi_capabilities` is an eleventh tool a model must weigh on every
`tools/list`.** #397 adds resources and #399/#400 add more tools. Nobody has measured whether a
capability tool actually gets called before a failing one, and the whole value of this issue rests
on that assumption.

### Required actions

1. **Operator, before this matrix is treated as authoritative:** check the ⛔ cells against one real
   OpenHome or UPnP renderer. Twelve of fourteen provider limitations are unverified against
   hardware.
2. **Decide whether `volume: supported` for a fixed-volume Roon output is acceptable**, or whether a
   fourth `device_dependent` value is wanted. Cheap now; expensive once clients depend on three.
3. **File the HQPlayer MCP routing gap against #328** if it is not already in scope there. `hifi_zones`
   advertises a zone type `hifi_control` cannot reach; this PR labels it, it does not fix it.
4. **A UPnP skip issue.** `transport_skip` for UPnP is tracked at the epic (#392) because no
   dedicated issue exists. I did not create one — that is the operator's call, not mine.

### Disposition

Implementation complete against every acceptance criterion, with two criteria amended on evidence
(five prefixes not four; transport split). No merge sought. No labels added. No force-push.

### Residual risk

Specification-derived ⛔ cells for OpenHome and UPnP; per-provider rather than per-device
granularity; and no evidence that any MCP client surfaces `structuredContent` to a model, which
#395 already flagged and which this PR inherits — the capability payload also rides the text, so
`hifi_capabilities` is readable either way, but the refusal classifications are envelope-only.

### Next gate authorization

Ship gate: authorized. The installable evidence is the two `dx build` runs plus the full workspace
suite; there is no artifact to deploy, since this is an additive MCP surface change with no HTTP or
packaging impact.

muness:
### Dissent — execute (#398)

Aimed where the brief said to aim it: **a capability value that is technically derived and still lies
to a client.** The design dissent named six; this one reports which survived contact with the code,
which got worse, and one I did not see coming.

**1. `⛔` is now harder to challenge than the `❌` that caused this issue. That is the central
trade and I do not think I have paid for it.**

#398 exists because a hand-written `❌` sat in AGENTS.md unchallenged. My fix generates the table —
so the next reader sees a cell blessed by a passing test, produced by a module whose docs say
`Supported` is unforgeable, footnoted with a confident sentence. **Every one of those signals is
about provenance, and none is about truth.** Twelve of fourteen `⛔` cells are my reading of a
specification I have no device to check against.

The mitigation I chose — each cell states "Verified from the UPnP AV service definitions, not from a
device" — is real and I still think it is the right call. But be clear about what it is: **a
disclaimer inside the thing being disclaimed**, read by a model that has no way to weigh it. The
honest version would be a distinct state for "UHC believes, unverified", and I did not build it
because the issue mandates three. **Cost if trusted:** a client permanently stops asking a real
device for something it does.

**2. `volume: supported` for a fixed-volume Roon output survived the design gate, and my defence of
it got weaker on inspection.**

I said the aggregator's `volume_control: None` conflates "no volume control" with "not read yet", so
deriving from it would mislabel new zones. That is true. But while implementing I found that
`src/adapters/roon.rs:1703` **re-emits `ZoneDiscovered` specifically when a zone gains volume
control** — so Roon's adapter already treats that transition as a first-class event, and the
"transient None" window may be much narrower than I assumed. I did not measure it. I kept the stable
wrong answer because measuring would have cost time, and then wrote a paragraph explaining why
measuring was unnecessary. **That is the sequence to distrust.** If someone spends twenty minutes on
that window, per-zone volume may well be derivable, and my caution was laziness with a rationale
attached.

**3. My own new test taught me that my new prose was self-contradicting, and I only found out by
running it.**

`an_unprefixed_zone_id_is_refused_by_name_instead_of_routed_to_roon` asserted
`!text.contains("Roon")` — and failed, because the refusal's *repair hint* says
`"call again with 'roon:1601a5d4bare'"`. The test was wrong, not the code. I fixed it to assert the
absence of the adapter's own fingerprint (`"Not connected to Roon"`). **What worries me is the
near-miss:** the fix I reached for first was "loosen the assertion", and had I written it as
`!text.contains("Not connected")` I would have had an assertion that passes for the wrong reason, on
the single test that proves this PR's load-bearing behavior change. The version that shipped checks
the adapter fingerprint *and* the exact repaired id, which is stronger — but I got there by
accident.

**4. The "proved 15 supported cells end to end" claim is weaker than it reads.**

Every one of those 15 is proved by an **offline failure** — the adapter refusing an id it does not
know, in its own words. That proves the call reached the right adapter. It does **not** prove the
call would succeed against a real device. `openhome:` volume is now `supported` on the strength of
`"Device not found: abc"`, which tells me the SOAP call was never issued. So for OpenHome and UPnP
volume — the feature this PR actually wires — **there is no evidence any device has ever accepted it
from the MCP path.** The HTTP path has, which is where my confidence comes from, and that is a
different code path with a different value type (`i32` from JSON vs. `f64` cast from MCP). The cast
is the untested seam: `value as i32` truncates, so `volume_set value=30.7` sends 30. Nobody has
checked whether a client sends fractional volumes.

**5. Splitting `transport` was justified by exactly one grep, and I said so in the design gate, and
I did it anyway.**

Unchanged from the design dissent, and now shipped. `transport: supported` still covers a set of
actions I have not enumerated per provider. `stop` is the visible loose end: three adapters implement
it, `hifi_control` does not accept it, and no capability mentions it. So the vocabulary quietly
excludes an action that three of five providers support — which is the same kind of understatement
this epic keeps finding, committed by the issue whose job was to end it.

**6. I broke my own working tree with `git checkout` and shipped a commit whose message describes
absent code.**

Recorded in the execute gate and the PR body. What I want to add here is the part that is not a
process note: **for about twenty minutes the branch was in a state where `cargo test --test
mcp_contract` would have reported 78 passing** — because I had already regenerated the AGENTS.md
matrix by hand before noticing `transport.rs` was reverted. It was only the *empty* matrix block, a
thing I checked out of suspicion rather than by any test, that exposed it. **No test in this PR
would have caught `transport.rs` reverting to the #395 version**, because the pre-#398 transport.rs
plus the #398 tests do not compile — and a compile failure is loud, but I had already committed by
then. The guardrails I am most proud of here protect against a wrong claim; none of them protects
against me deleting the implementation.

**7. Nobody has established that a model calls `hifi_capabilities` before calling a tool that
fails.**

The entire value proposition is "a model that checks before it tries". #395's PR already flagged that
no real MCP client has been verified to surface `structuredContent`; this PR adds a *tool*, which is
better — its answer rides the human-readable text too, so any client can read it. But an eleventh
tool a model never chooses to call is worse than nothing: it costs context on every `tools/list` and
delivers zero. **The cheapest possible check is worth more than anything else in this PR:** point one
real client at it, ask it to set the volume on an OpenHome zone, and see whether it consults
capabilities first. If it does not, the design question is whether capability data belongs in the
tool *descriptions* instead — which is a different PR and possibly a better one.

muness:
## Gate report — ship (#398 CAPABILITIES)

**Gate/role:** ship. **Verdict: pass-with-actions.** Verifiable, rollback-able, and **not merged.**

### Aim and scope checked

Every acceptance criterion in #398, with two amended on evidence and both amendments argued in the
PR body: five accepted prefixes rather than four, and `transport` split so UPnP's skip refusal is not
mislabelled.

### Evidence inspected

```
$ git diff --stat origin/feat/issue-395-mcp-envelope..HEAD
 AGENTS.md                     |  169 +-      src/mcp/tools/capabilities.rs |  126 +
 src/adapters/upnp.rs          |   36 +-      src/mcp/tools/library.rs      |   97 +-
 src/mcp/capabilities.rs       |  945 +       src/mcp/tools/mod.rs          |   15 +-
 src/mcp/handler.rs            |    3 +       src/mcp/tools/transport.rs    |  297 +-
 src/mcp/mod.rs                |    2 +       tests/fixtures/mcp_tools.json |   16 +
 src/mcp/routing.rs            |  563 +-      tests/mcp_contract.rs         | 1400 +-
 12 files changed, 3255 insertions(+), 414 deletions(-)
```

**No HTTP change.** `git diff ... -- tests/fixtures/api_routes.txt` is empty, so no
`api-change-approved` label is needed and none was added. The only file outside `src/mcp` and
`AGENTS.md` is `src/adapters/upnp.rs`, where the change is one `pub const`, one helper, and a match
arm producing byte-identical strings.

**Final gate run, on a clean tree:**

```
$ make css && cargo fmt --check                    (exit 0, no output)
$ cargo clippy -- -D warnings                      Finished in 8.75s (exit 0)
$ cargo test --workspace --no-fail-fast            23 suites, all "test result: ok", 0 FAILED
$ CARGO_BUILD_RUSTC_WRAPPER="" dx build --fullstack --release @client \
    --no-default-features --features web @server --features server
                                                   Server build completed successfully
$ CARGO_BUILD_RUSTC_WRAPPER="" dx build --release --platform web --features web
                                                   Client build completed successfully
```

`mcp_contract` is 79 tests, up from 68 on the base. Every pre-existing suite is unchanged in count.

### How an operator verifies this without reading the diff

1. `cargo test --test mcp_contract` — 79 pass.
2. Start the server, point an MCP client at `/mcp`, call `hifi_capabilities` with no arguments. The
   `providers` array is the same table as AGENTS.md's, footnotes and all.
3. Call `hifi_control` with a bare zone id. It refuses, and tells you the exact prefixed id to use.
4. `UPDATE_AGENTS_MATRIX=1 cargo test --test mcp_contract agents_md` then `git diff AGENTS.md` — empty,
   because the doc is already the derived data.

### Findings

**P1 — Merge order.** This PR's base is `feat/issue-395-mcp-envelope` (PR #410), which is itself based
on #394's branch. Nothing here can merge to `v3` before both, and merging this into #410 first is what
gets it CI for the first time. **Do not merge any of the three without deciding the order deliberately.**

**P2 — `0452bcb` is an incomplete commit and its message overstates it.** A `git checkout` used to undo
a breakage experiment reverted `transport.rs` and `AGENTS.md` past uncommitted work. `cc8afa9` restores
both. Not rebased away: the repo squash-merges, so the final history is one commit regardless, and
hiding a mistake in the PR whose subject is capability honesty would be the wrong trade. **Reviewers
should read `cc8afa9`'s `transport.rs` diff as the change.**

**P3 — One defect this PR introduced and then fixed itself.** `value as i32` truncated, making
`volume_up value=0.5` a silent no-op on the two providers #398 wired. Found by this PR's own dissent,
fixed in `27bea38` with a test. It is worth noting that the *only* thing that caught it was writing
the dissent — no test, review or lint would have.

**P4 — `#392` is `tracked_by` for 16 cells.** OpenHome's queue and mode capabilities and UPnP's
skip/repeat/shuffle have no dedicated issue, so they point at the epic. That is honest (the epic does
track them) and weak (an epic closes without them). Four candidate issues are named in the execute
gate's required actions.

### Contrary case

**The strongest argument for not shipping this yet:** nobody has demonstrated that an MCP client calls
`hifi_capabilities` before calling a tool that fails. If models do not consult it, this PR has added an
eleventh tool that costs context on every `tools/list` and delivers nothing — and the correct design
would be capability data folded into the tool *descriptions*, which is a different and possibly better
PR. The behavior changes (explicit routing, closed action set, wired OpenHome/UPnP volume) stand on
their own regardless; the capability *tool* is the part resting on an unmeasured assumption.

**The second:** #401 is the epic's validation issue and exists to do exactly this end-to-end proof.
Shipping the surface and validating later is the epic's own plan, so this is a sequencing argument
rather than a blocking one — but it should be #401's first check, not an afterthought.

### Required actions

1. **No merge without explicit operator approval**, and a decided order across #404 / #410 / #424.
2. Check the ⛔ cells against one real OpenHome or UPnP renderer. Twelve of fourteen are
   specification-derived. Ten minutes, and it is the difference between "UHC believes" and "the device
   cannot".
3. Try `hifi_capabilities` with a real client and see whether a model uses it unprompted. If not, the
   design question in the contrary case is live.
4. Decide on `volume: supported` for fixed-volume Roon outputs — accept it, or add a fourth
   `device_dependent` value. Cheap now.

### Disposition

Ready for review. **Not merged. No labels. No force-push.** Four commits, pushed as-is.

### Residual risk

Specification-derived provider limitations for OpenHome and UPnP; per-provider rather than per-device
granularity; `not_implemented` refusals carried by a single zone type's coverage; and the unmeasured
assumption that a model checks before it acts.

### Next gate authorization

None from me — the next gate is the operator's, and #401 is where the end-to-end proof belongs.

muness:
### Dissent — ship (#398)

Last pass, aimed at the same target: **a capability value that is technically derived and still lies
to a client.** Three of these are retractions or sharpenings of things I asserted earlier in this PR.

**1. Retraction: "a hand-written ✅ does not compile" is true and I let it carry more weight than it
can.**

I wrote it in the module docs, the commit message, and the PR body, in that order, each time a little
more prominently. It is literally true — `Gap` has no `Supported` variant, and `E0599` proves it. But
count what it actually protects: **five of the 90 cells.** The 15 `supported` values are produced by
routing, and the type stops me *contradicting* routing in the table. It does nothing about the 75
non-supported cells, where the real risk lives — a `⛔` that should be `🚧`. For those, the only
guardrail is my own bias rule and an evidence-length assertion (`> 60` chars), which is a proxy any
padded sentence defeats.

So the headline claim of this PR guards the part that was never in danger. **If you are auditing one
thing, audit the `⛔` cells, not the type.**

**2. Found and FIXED at the ship gate: three cells were `⛔` in violation of my own rule.**

I had written UPnP `play_next`'s evidence as two facts in tension: `SetNextAVTransportURI` *does*
queue one following item, so the action exists — but it takes a URI, and a renderer-only integration
has no library to resolve one from. I marked it `⛔` and put both facts in the `detail` so a reader
could weigh them.

Re-reading it here, that is not applying my own rule. The rule says `⛔` only when the protocol
demonstrably lacks it, and **I had just written down that the protocol has it.** What is missing is
UHC's content addressing — a UHC gap, the definition of `🚧`. I had reasoned "you still cannot use
it", which is true and is *not the question the field answers*.

Auditing the neighbours with that lens found two more: `upnp / play_by_ref` and
`openhome / play_by_ref`, both `⛔` on the same conflation. UPnP `AVTransport:1` has
`SetAVTransportURI`; OpenHome `Playlist:1` has `Insert(AfterId, Uri, Metadata)`. Both can play a
*named* item; neither can *find* one.

**All three are now `not_implemented` tracked by #396**, sharing one named evidence const
(`PLAY_A_NAMED_URI_EXISTS`) that states the distinction explicitly. `play_by_ref` is now uniformly
`🚧 #396` across all five providers — no protocol forecloses playing a reference, and it was only my
sloppiness that said three of them did.

`search`, `browse` and `play_by_query` keep their `⛔`: no action in either service set resolves free
text at all, which is a genuine protocol absence.

**My first draft of this section said "I have not changed it, because changing one cell at the ship
gate would be worse than flagging it."** That was wrong, and worth recording: this is the PR whose
entire subject is not shipping false "never"s. Flagging one I had already identified would have been
indefensible.

**3. `every_lms_gap_cites_live_verification` checks for a substring.**

It asserts each LMS evidence string contains `"verified live"` or `"verified against live"`. That is
a test that a **phrase** is present, dressed as a test that a **check** happened. I wrote those
strings; I can put that phrase on anything. The real provenance is #402's and #403's comments, which
no test can reach. Of everything in this PR this is the assertion whose name most overstates it, and
LMS's row is the one I have most confidently told people to trust.

The defensible version: I did read both inventories, and every LMS command I cite appears in one of
them with a `[LIVE]` marker. That is a claim about my diligence, not about the test.

**4. I asserted "no local-only failures this time" and that is narrower than it sounded on first
draft.**

#395's PR reported `/hqp/discover` failing on this host for want of UDP SSDP. Both tests pass here.
My first draft of the PR body said the environment issue "is gone", which would have been the exact
over-generalisation #392's standing facts warn about — twice, by two other agents. I corrected it to
"this run is green including them", which is all the evidence supports. Flagging that I nearly did it
anyway, in the PR whose job is preservation claims backed by evidence.

**5. The behavior changes are the valuable half, and they are the half nobody will review carefully.**

Seven behavior changes, each with a test, each enumerated in a table. But the review attention will go
to the 945-line capability module, because that is where the new ideas are. Meanwhile change #2 —
bare ids no longer route to Roon — is the one that can break a working client. My mitigations are
real (every published id is prefixed; the refusal names the repaired id; a test asserts it). What I
cannot mitigate is that **I have no way to know whether any deployed client sends bare ids.** #221 is
a report of ChatGPT looping on `hifi_play`, which means real models really are calling these tools
with values they invented. If one of them invented a bare Roon id that happened to work, this PR
breaks it — and the client's recovery depends on a model reading a refusal sentence and acting on it,
which is the same assumption #221 shows failing.

That is the risk I would want stated to the operator in one line: **the load-bearing behavior change
is defended by a well-written error message, and #221 is evidence that models do not always read
those.**

PR #425, opened 2026-08-01
feat(mcp): expose readable bridge state as MCP resources (#397)

muness:
## Summary

Implements #397 (RESOURCES) in epic #392: exposes readable bridge state as MCP resources, addressable by URI, alongside the existing `read_only_hint` tools — which keep working unchanged.

Base: `v3` at `5996af5` (includes #394 FOUNDATION, #395 ENVELOPE, #398 CAPABILITIES). This PR targets `v3` directly, so the full CI Lint/Test/Build set applies — see CI status below, not just local verification.

## Solution space: already resolved, not relitigated

Per the issue and #392, the solution-space gate for this issue was resolved before implementation started (see the issue body's "SDK findings" and "Decisions taken" sections). Confirming briefly rather than re-deriving:

- **Live per-zone enumeration**, not one big document or a resource template. `resources/list` reads `ZoneAggregator` fresh on every call, so a zone discovered after a client's last list is addressable on the very next one, with no restart — see `a_newly_discovered_zone_is_addressable_without_a_restart` in `tests/mcp_contract.rs`.
- **`listChanged`: advertised and implemented.** Wired off the same bus zone-discovery events (`ZoneDiscovered`, `ZoneRemoved`, `ZonesFlushed`) `ZoneAggregator` already consumes. Documented as best-effort: with `event_store: None` (`src/mcp/mod.rs`), a client with no open GET/SSE stream never receives it and there is no replay.
- **`subscribe`: not advertised.** The pinned SDK has no subscription registry and no replay, so advertising it would claim something this server cannot honor. A client that calls `resources/subscribe` anyway reaches the SDK's own default handler — a plain `method_not_found` (`-32601`), not a capability-gate refusal — asserted by `resources_subscribe_is_not_advertised_and_not_implemented`.
- **Declaring the `resources` capability at all is what makes the five resource hooks on `ServerHandler` reachable** — the SDK's dispatch gate keys only on capability *presence*, never on `subscribe`/`listChanged` sub-fields, confirmed directly by `declared_protocol_version_and_capabilities_are_pinned`.
- **Payload shape equals the tool's envelope `data`.** No parallel shape was invented — see "No second source of truth" below.
- Left room for #418 (album art / `BlobResourceContents`) without building for it speculatively — the URI scheme and `ReadResourceContent` enum (`TextResourceContents | BlobResourceContents`) both already accommodate a future blob resource; nothing here forecloses it.

## What's covered

`resources/list` and `resources/read`, for:

| URI | Equivalent tool |
|---|---|
| `hifi://zones` | `hifi_zones` |
| `hifi://zones/{zone_id}` (one per live zone) | `hifi_now_playing` |
| `hifi://status` | `hifi_status` |
| `hifi://hqplayer/status` (HQPlayer enabled only) | `hifi_hqplayer_status` |
| `hifi://hqplayer/profiles` (HQPlayer enabled only) | `hifi_hqplayer_profiles` |

- **No second source of truth.** `src/mcp/tools/{zones,status,hqplayer}.rs` had their payload-building logic extracted into functions (`zones_payload`, `now_playing_payload`, `status_payload`, `hqp_status_payload`, `hqp_profiles_payload`) that both the tool handler and `src/mcp/resources.rs` call. There is exactly one place that reads the aggregator or an adapter accessor for each kind of state — a resource and its tool cannot compute different answers. No behavior change to the tools themselves (same fixtures, same tests, still green).
- **Tool/resource agreement is tested, not just asserted in prose**, for all five resources (the acceptance criteria require it for at least zones and now-playing): `zones_resource_agrees_with_hifi_zones_tool`, `now_playing_resource_agrees_with_hifi_now_playing_tool_for_a_live_zone` (a real mock LMS server + adapter + bus + aggregator, non-trivial payload with title/artist/album, not just the empty case), `status_resource_agrees_with_hifi_status_tool`, `hqplayer_status_resource_agrees_with_the_tool_for_a_live_connection` (mock HQPlayer server, real `connected: true`/`host`), and `hqplayer_profiles_resource_agrees_with_hifi_hqplayer_profiles_tool`. The last three were added after this PR's own dissent pass found that satisfying the acceptance criteria's literal minimum ("at least zones and now-playing") had left three of five resources' tool/resource agreement enforced only structurally (shared payload functions) and unverified by any test.
- **HQPlayer resources are gated exactly like HQPlayer tools.** `hqplayer_resources_are_absent_when_adapter_disabled` mirrors the existing `hqplayer_tools_filtered_when_adapter_disabled`, same settings fixture, same assertion shape. As with tools, this only gates *advertisement* — `read_resource` doesn't re-check settings, matching how a HQPlayer tool called directly (unlisted) still executes.
- **Unknown/stale URIs are a proper JSON-RPC error, never a panic or empty success.** `unknown_or_stale_resource_uri_is_a_proper_error` covers both an unrecognised scheme and a `hifi://zones/{zone_id}` for a zone the aggregator has never held. Uses the MCP-conventional "Resource not found" code (`-32002`), since the SDK's `RpcError` (returned by every `ServerHandler` resource method) has no built-in constructor for it — only a different type, `SdkError`, used internally by the transport, does — so the struct is built directly (its fields are public for exactly this). The error's `data` also carries a `hint` pointing at `resources/list` — the resource-side equivalent of `hifi_now_playing`'s `Refusal::UnknownTarget` naming `hifi_zones` as `discover_with` for the same underlying fact. (Added in a follow-up commit after self-review flagged the first version as less actionable than the tool's equivalent refusal — see commit `4ea6f17`.)
- **`resources/templates/list` returns an empty list**, not the SDK's generic refusal — resources genuinely exist here, there's just no template shape for any of them (live enumeration was chosen instead).

## The capability delta (#394's snapshot, updated deliberately)

`tests/fixtures/mcp_initialize.json`:

```diff
   "capabilities": {
+    "resources": {
+      "listChanged": true
+    },
     "tools": {}
   },
```

`tests/mcp_contract.rs::declared_protocol_version_and_capabilities_are_pinned` now asserts `resources.list_changed == Some(true)` and `resources.subscribe == None` directly off `server_details()`.

`tests/mcp_contract.rs::resource_methods_are_refused_while_only_tools_is_declared` (the test #394 was asked to write specifically so this diff would show up) is renamed to `resource_methods_dispatch_now_that_resources_is_declared` and rewritten to assert the new behavior: `resources/list` and `resources/read` on a known URI now succeed (previously `-32603 Server does not support resources`), and `resources/templates/list` returns `{"resourceTemplates": []}` rather than an error.

## Execute gate: genuine red before green

Commit `525924c` updates/adds the acceptance-criteria tests against **today's** (pre-#397) code and is red for real — 8 failures, all with the expected `-32603 Server does not support resources` shape or the still-missing capability. Commit `47cfbff` is the implementation; the same test file is green afterward with no further test edits.

Per the epic's standing instruction ("a test that asserts nothing is worse than no test"), I broke two of these on purpose to confirm they fail for the right reason before trusting them:
- `now_playing_resource_agrees_with_hifi_now_playing_tool_for_a_live_zone`: temporarily made `now_playing_payload` swap `artist`/`album` — the test failed with a value mismatch, not a panic or a vacuous pass.
- `hqplayer_resources_are_absent_when_adapter_disabled`: temporarily removed the `hqplayer_enabled` gate in `list_resources` — the test failed on the extra URIs.
Both reverted before the final commit.

## Explicitly deferred, and why

- **`subscribe`.** Would need a subscription registry the SDK doesn't provide, plus replay for a dropped stream, which `event_store: None` doesn't support. The issue explicitly excludes advertising it without implementing it.
- **Cover art / `BlobResourceContents` (#418).** Not started here; the URI scheme and content-enum choice don't foreclose it, per the issue's priority note.
- **Delivery proof for `listChanged`.** The notifier is exercised indirectly (bus events fire it, the handler doesn't panic, sessions stay usable afterward) but actual SSE delivery to an open GET stream isn't asserted — that would need a long-lived streaming test harness this suite doesn't have. Documented as best-effort in both `src/mcp/server.rs` and `src/mcp/resources.rs`; a client should treat it as a hint to re-list, not a guarantee.

## Verification (this checkout, in the order specified)

1. `make css` — clean, first thing in this fresh worktree.
2. `cargo fmt --check` — clean.
3. `cargo clippy -- -D warnings` — clean (not `--all-targets`; that target set is known-red crate-wide per #392's standing notes, unrelated to this change).
4. `cargo test --workspace` — 91 passed in `mcp_contract`, workspace-wide all green (including `tests/spawn_cancellation_lint.rs`, which the new background notifier task had to satisfy — see below).
5. `CARGO_BUILD_RUSTC_WRAPPER="" dx build --release --platform web --features web` — clean: client build completed (wasm-bindgen, assets copied) and server build completed. Needed rustup's toolchain ahead of Homebrew's on `PATH` (`~/.cargo/config.toml` pins `rustc-wrapper = sccache`, and Homebrew's `rustc` doesn't have `wasm32-unknown-unknown` installed under this account, so the first attempt failed with "Missing rust target wasm32-unknown-unknown" until `PATH` was corrected — a known local-environment quirk, not a code issue).

## CI (this PR targets `v3`, so this is a real gate — not just local-clean)

All checks green as of the latest push (`807d50c`): **Lint** pass (42s), **Test** pass (2m15s), **Build Linux x64** pass (4m8s), **Build WASM Assets** pass (4m25s), **Synology Package Contract** pass, **Smoke Test Binary** pass, **Build Summary** pass.

One thing not anticipated by the issue text: `tests/spawn_cancellation_lint.rs` is an AST-level lint asserting every `tokio::spawn` loop in `src/` selects on a cancellation token. `spawn_list_changed_notifier`'s bus-subscriber loop needed a `tokio::select!` arm on `AppState::shutdown` (the same `CancellationToken` every other background loop in this crate uses) to pass it — done in the implementation commit, not a separate fix.

## This PR does not

Merge, add labels, or force-push. Ready for the ship-gate review and dissent (posted as separate PR comments per the epic's process).

### DIFF
--- a/src/mcp/handler.rs
+++ b/src/mcp/handler.rs
@@ -1,8 +1,8 @@
-//! The MCP `ServerHandler`: advertise tools, and dispatch calls to them.
+//! The MCP `ServerHandler`: advertise tools, dispatch calls to them, and (#397)
+//! advertise and serve resources.
 //!
-//! Dispatch only. Every arm delegates to a handler in
-//! [`crate::mcp::tools`]; no tool behavior lives here. #397 extends this impl
-//! with the resource methods.
+//! Dispatch only. Every arm delegates to a handler in [`crate::mcp::tools`] or
+//! [`crate::mcp::resources`]; no tool or resource behavior lives here.
 
 use crate::api::{load_app_settings, AppState};
 use crate::mcp::envelope::{Envelope, Refusal};
@@ -11,8 +11,9 @@ use async_trait::async_trait;
 use rust_mcp_sdk::{
     mcp_server::ServerHandler,
     schema::{
-        schema_utils::CallToolError, CallToolRequestParams, CallToolResult, ListToolsResult,
-        PaginatedRequestParams, RpcError,
+        schema_utils::CallToolError, CallToolRequestParams, CallToolResult,
+        ListResourceTemplatesResult, ListResourcesResult, ListToolsResult, PaginatedRequestParams,
+        ReadResourceRequestParams, ReadResourceResult, RpcError,
     },
     McpServer,
 };
@@ -31,6 +32,19 @@ impl HifiMcpHandler {
 
 #[async_trait]
 impl ServerHandler for HifiMcpHandler {
+    /// Fires once per session, when the client sends `notifications/initialized`
+    /// — the SDK hands us that session's own `runtime` here, which is the only
+    /// place a per-session background notifier can be started. See
+    /// [`crate::mcp::resources::spawn_list_changed_notifier`] for what this
+    /// notifier does and does not promise.
+    async fn on_initialized(&self, runtime: Arc<dyn McpServer>) {
+        crate::mcp::resources::spawn_list_changed_notifier(
+            self.state.bus.clone(),
+            runtime,
+            self.state.shutdown.clone(),
+        );
+    }
+
     async fn handle_list_tools_request(
         &self,
         _params: Option<PaginatedRequestParams>,
@@ -99,6 +113,52 @@ impl ServerHandler for HifiMcpHandler {
                 tools::capabilities::handle_capabilities(state, args).await
             }
         }
+    }
+
+    /// Same settings-based gate as `handle_list_tools_request`: the HQPlayer
+    /// resources are hidden when the adapter is disabled, exactly like the
+    /// HQPlayer tools.
+    async fn handle_list_resources_request(
+        &self,
+        _params: Option<PaginatedRequestParams>,
+        _runtime: Arc<dyn McpServer>,
+    ) -> Result<ListResourcesResult, RpcError> {
+        let settings = load_app_settings();
+        Ok(ListResourcesResult {
+            meta: None,
+            next_cursor: None,
+            resources: crate::mcp::resources::list_resources(
+                &self.state,
+                settings.adapters.hqplayer,
+            )
+            .await,
+        })
+    }
+
+    /// No resource template is advertised — #397's solution-space gate chose
+    /// live per-zone enumeration in `resources/list` instead (a newly
+    /// discovered zone needs no restart there either). This returns a fixed
+    /// empty list rather than the SDK's generic `method_not_found`, because
+    /// resources genuinely exist here; there is just no template shape for any
+    /// of them.
+    async fn handle_list_resource_templates_request(
+        &self,
+        _params: Option<PaginatedRequestParams>,
+        _runtime: Arc<dyn McpServer>,
+    ) -> Result<ListResourceTemplatesResult, RpcError> {
+        Ok(ListResourceTemplatesResult {
+            meta: None,
+            next_cursor: None,
+            resource_templates: vec![],
+        })
+    }
+
+    async fn handle_read_resource_request(
+        &self,
+        params: ReadResourceRequestParams,
+        _runtime: Arc<dyn McpServer>,
+    ) -> Result<ReadResourceResult, RpcError> {
+        crate::mcp::resources::read_resource(&self.state, &params.uri).await
     }
 }
 
--- a/src/mcp/mod.rs
+++ b/src/mcp/mod.rs
@@ -19,6 +19,7 @@
 //! | [`envelope`] | the structured result envelope every tool fills           |
 //! | [`routing`]  | zone-prefix -> adapter, in exactly one place              |
 //! | [`capabilities`] | per-provider capability truth, in three states        |
+//! | [`resources`] | readable bridge state addressable by URI (#397)          |
 //!
 //! # The MCP surface is pinned by tests
 //!
@@ -34,6 +35,7 @@
 pub mod capabilities;
 pub mod envelope;
 pub mod handler;
+pub mod resources;
 pub mod routing;
 pub mod server;
 pub mod tools;
--- a/src/mcp/resources.rs
+++ b/src/mcp/resources.rs
@@ -1,0 +1,384 @@
+//! MCP resources: readable bridge state addressable by URI (issue #397).
+//!
+//! # What this is, and is not
+//!
+//! Every value here is already reachable through a `read_only_hint` tool —
+//! that tool keeps working exactly as it does today. Resources are a second,
+//! address-based path to the *same* state, for clients that model ambient
+//! state (what is playing, is the bridge connected) as something to read
+//! rather than something to ask for by calling a tool. This is the least
+//! load-bearing issue in epic #392: nothing here is new capability, only a new
+//! shape for capability that already exists.
+//!
+//! # URI scheme
+//!
+//! | URI                        | Equivalent tool           |
+//! |-----------------------------|---------------------------|
+//! | `hifi://zones`              | `hifi_zones`               |
+//! | `hifi://zones/{zone_id}`    | `hifi_now_playing`         |
+//! | `hifi://status`             | `hifi_status`              |
+//! | `hifi://hqplayer/status`    | `hifi_hqplayer_status`     |
+//! | `hifi://hqplayer/profiles`  | `hifi_hqplayer_profiles`   |
+//!
+//! `zone_id` already carries its own provider prefix (`roon:`, `lms:`, ...), so
+//! `hifi://zones/roon:1601a5d4` is a valid URI despite the second colon — no
+//! escaping is applied or required, because URIs are matched by prefix-stripping
+//! rather than parsed as RFC 3986 authority/path components.
+//!
+//! # No second source of truth
+//!
+//! Each resource's payload is built by the exact function its equivalent tool
+//! calls — [`crate::mcp::tools::zones::zones_payload`],
+//! [`crate::mcp::tools::zones::now_playing_payload`],
+//! [`crate::mcp::tools::status::status_payload`],
+//! [`crate::mcp::tools::hqplayer::hqp_status_payload`] and
+//! [`crate::mcp::tools::hqplayer::hqp_profiles_payload`]. There is exactly one
+//! place that reads the aggregator or the adapter accessors for each kind of
+//! state, so a resource and its tool cannot compute different answers for the
+//! same underlying state. `tests/mcp_contract.rs` asserts this directly for
+//! zones and now-playing (including a live LMS zone, not just the empty case).
+//!
+//! # Live per-zone enumeration, not a template
+//!
+//! [`list_resources`] reads `ZoneAggregator` fresh on every call, so a zone
+//! discovered after a client's last `resources/list` is addressable on the very
+//! next one — no restart, no resource template needed. A single `hifi://zones`
+//! document was rejected in the solution-space gate: a zone inside a bigger
+//! document has no URI of its own, which fails addressability outright.
+//!
+//! # HQPlayer resources respect the same settings gate as HQPlayer tools
+//!
+//! [`list_resources`] takes the same `hqplayer_enabled` flag
+//! [`crate::mcp::tools::list_tools`] does, and hides the two HQPlayer resources
+//! on the same condition. As with tools, this only gates *advertisement* —
+//! [`read_resource`] does not re-check settings, matching how a HQPlayer tool
+//! called directly (unlisted) still executes.
+//!
+//! # `listChanged`: best-effort, wired off the events the aggregator already
+//! consumes
+//!
+//! [`spawn_list_changed_notifier`] subscribes to the same bus
+//! [`crate::aggregator::ZoneAggregator`] already consumes and sends
+//! `notifications/resources/list_changed` on any event that changes the *set*
+//! of zones (`ZoneDiscovered`, `ZoneRemoved`, `ZonesFlushed`) — not on every
+//! `NowPlayingChanged`, which would be noise for a signal that only means "the
+//! zone list changed, re-list". This is genuinely best-effort: with
+//! `event_store: None` (`src/mcp/mod.rs`), a client with no open GET/SSE stream
+//! simply does not receive it, and there is no replay. That is why `subscribe`
+//! is not advertised at all — see [`crate::mcp::server::server_details`].
+//!
+//! # Reading an unknown or stale URI
+//!
+//! [`read_resource`] never panics and never returns an empty success for a URI
+//! it does not recognise or a zone the aggregator no longer holds. Both produce
+//! a JSON-RPC error with the MCP-conventional "resource not found" code
+//! (`-32002`), carrying the offending URI in `data` so a client can log which
+//! one it asked for.
+
+use crate::api::AppState;
+use crate::bus::{BusEvent, SharedBus};
+use crate::mcp::tools::{hqplayer, status, zones};
+use rust_mcp_sdk::{
+    schema::{ReadResourceContent, ReadResourceResult, Resource, RpcError, TextResourceContents},
+    McpServer,
+};
+use serde::Serialize;
+use std::sync::Arc;
+use tokio_util::sync::CancellationToken;
+
+/// `hifi_zones`'s resource equivalent: every zone, in one document.
+const ZONES_URI: &str = "hifi://zones";
+/// Prefix for a per-zone `hifi_now_playing` equivalent. The full URI appends the
+/// zone id verbatim, e.g. `hifi://zones/roon:1601a5d4`.
+const ZONE_PREFIX: &str = "hifi://zones/";
+/// `hifi_status`'s resource equivalent.
+const STATUS_URI: &str = "hifi://status";
+/// `hifi_hqplayer_status`'s resource equivalent. Listed only when the HQPlayer
+/// adapter is enabled in settings.
+const HQP_STATUS_URI: &str = "hifi://hqplayer/status";
+/// `hifi_hqplayer_profiles`'s resource equivalent. Listed only when the
+/// HQPlayer adapter is enabled in settings.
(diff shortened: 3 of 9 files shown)


muness:
## Ship-gate review — #397 RESOURCES

**Verdict:** pass-with-actions (one action already taken; two residual risks recorded, neither blocking)

**Role:** self-review by the implementer, per #392's gate-report format.

### Aim and scope checked

Does this PR do what #397 asks, and nothing more? Checked against the issue's acceptance criteria one by one:

- [x] `initialize` declares `resources` with only implemented sub-features — `list_changed: Some(true)`, `subscribe: None`. Confirmed by `declared_protocol_version_and_capabilities_are_pinned` reading `server_details()` directly (not the wire snapshot, which can't see a value the SDK doesn't validate against the client's request).
- [x] `resources/list` and `resources/read` for all zones, per-zone now-playing, bridge status, HQPlayer pipeline status, HQPlayer profiles — `src/mcp/resources.rs`, five URIs, all wired through `handler.rs`.
- [x] Live per-zone enumeration, addressable without restart — `list_resources` reads `state.aggregator.get_zones()` fresh on every call; `a_newly_discovered_zone_is_addressable_without_a_restart` publishes a `ZoneDiscovered` event mid-test and asserts the new URI appears on the *next* `resources/list`, no process restart involved.
- [x] Tool/resource payload agreement is tested, for zones and now-playing — `zones_resource_agrees_with_hifi_zones_tool` (empty case) and `now_playing_resource_agrees_with_hifi_now_playing_tool_for_a_live_zone` (live LMS zone, non-trivial payload with title/artist/album).
- [x] HQPlayer resources absent when the adapter is disabled, tested the same way the tool filter is — `hqplayer_resources_are_absent_when_adapter_disabled` mirrors `hqplayer_tools_filtered_when_adapter_disabled` line for line (same `SettingsFixture`, same assertion shape).
- [x] Unknown/stale URI is a proper MCP error, never a panic or empty success — `unknown_or_stale_resource_uri_is_a_proper_error`, both an unrecognised scheme and a zone id the aggregator never held.
- [x] #394's `initialize` snapshot updated deliberately, capability delta called out — see PR body's diff block; the specific test #394 wrote to make this visible (`resource_methods_are_refused_while_only_tools_is_declared`) is renamed and rewritten rather than deleted, so the history of what changed and why stays attached to the same test name lineage.
- [x] `cargo clippy -- -D warnings` clean (not `--all-targets`, which the epic's standing notes say cannot pass and isn't the gate).

Scope discipline: the issue names this "the least load-bearing issue in #392" and warns against over-building. I count five resources and zero resource templates, zero content-browsing, zero new HTTP routes, zero changes to `hifi_search`/`hifi_play`/refs (that's #396's territory, untouched — confirmed by `git diff --stat` against `v3` showing no changes under library.rs or anything ref-shaped). That matches the brief.

### Evidence inspected

- Read the full acceptance criteria and the epic's "Standing facts" section before writing any code.
- Read `rust-mcp-sdk` 0.8.3 and `rust-mcp-schema` 0.9.5 source directly (not the issue's summary of it) to confirm: the `ServerHandler` trait's five resource hooks and their default (`method_not_found`) behavior; that `ServerCapabilities::resources` gates dispatch on presence alone (matches the issue's claim, now also matches what the updated `resource_methods_dispatch_now_that_resources_is_declared` test demonstrates directly); that `RpcError` (the RPC-level type every resource handler must return) has no `resource_not_found()` constructor of its own — that constructor exists only on `SdkError`, a distinct internal type — which is why `resources.rs` builds `RpcError { code: -32002, .. }` by hand rather than calling something that doesn't exist on that type.
- Ran the full verification sequence specified in the task, in order: `make css`, `cargo fmt --check`, `cargo clippy -- -D warnings`, `cargo test --workspace` (all green, 91/91 in `mcp_contract`, no change to any other suite), `CARGO_BUILD_RUSTC_WRAPPER="" dx build --release --platform web --features web` (green on the second attempt — the first failed with "Missing rust target wasm32-unknown-unknown" because Homebrew's `rustc` was ahead of rustup's on `PATH`; putting rustup's toolchain bin dir first fixed it, matching the environment's own recorded fix for this exact failure mode).
- Verified the red-then-green execute-gate claim by re-reading both commits: `525924c` (red, 8 failing) and `4ea6f17`/`47cfbff` (green). Also deliberately broke two implementations mid-review to confirm the tests that are supposed to catch regressions actually do (swapped `now_playing_payload`'s artist/album — the live-zone agreement test failed with a value mismatch; removed the `hqplayer_enabled` gate — the filtering test failed on the extra URIs). Both reverted.

### Findings

**P2 (fixed during this review):** `resource_not_found`'s error `data` originally carried only the offending `uri`, with no path back to a valid one. The equivalent tool-side failure (`hifi_now_playing` on an unknown zone) uses `Refusal::UnknownTarget`, which names `discover_with: "hifi_zones"` specifically so a client has a next action, not just a diagnosis. A resource client hitting a stale `hifi://zones/{zone_id}` had no such pointer. Fixed in commit `4ea6f17`: `data.hint` now says to call `resources/list`, and the test was tightened to assert the exact code, the echoed uri, and the hint text (not just "some error exists").

**P3 (residual risk, not blocking):** `spawn_list_changed_notifier`'s claim that it "stops when sending starts failing" rests on reading the SDK's own session-teardown path (`delete_session` calls `runtime.shutdown()`, which clears the per-session transport before `send()` would next be attempted) rather than on an integration test that opens a real SSE stream, tears it down, and confirms the background task actually exits. Building that harness would need a persistent streaming client the existing `TestApp` (built on `tower::oneshot`, one request/response per call) doesn't support — out of scope for the least load-bearing issue in the epic, and the risk is bounded: a session that's never explicitly torn down keeps one parked `broadcast::Receiver` (cheap: a ring position and a waker) alive until either a zone-list event triggers a failed send, or the process's own `shutdown` `CancellationToken` fires. Documented as best-effort in both `server.rs` and `resources.rs`. Recommend: if a later issue adds real subscription support (explicitly out of scope here), re-examine whether this notifier's lifetime needs to be tied to something more precise than "next failed send."

**P4 (residual risk, cosmetic):** `now_playing_payload` (used by the resource) and `handle_now_playing` (the tool) each issue their own `state.aggregator.get_zone(zone_id)` call rather than sharing one read. Both go through the identical `now_playing_from_zone` transform, so they cannot disagree in *shape*, and in practice they read the same `RwLock`-guarded map with no adapter work in between, so a value difference between the two calls within one request would require a concurrent write landing in a multi-nanosecond window — not zero, but not worth the code complexity of threading a single fetched `Zone` through both a resource-shaped and a tool-shaped caller for a read-only, best-effort, least-load-bearing feature.

### Contrary case considered

Could this have been a single `hifi://zones` document with per-zone data nested inside, skipping live per-zone URIs entirely? That was the alternative the issue's own solution-space gate rejected before this PR started (a zone inside a bigger document has no URI of its own), and I re-checked the reasoning holds: a client that wants to reference *one* zone's now-playing (e.g., to correlate it against a later read) needs a stable URI for that zone, not an index into an array that can reorder as zones come and go. Live per-zone enumeration is right.

### Disposition

Pass-with-actions. The one required action (discoverability hint on resource-not-found) is already done in this branch. The two residual risks are recorded above and are not blockers for an issue whose own text says it should yield to the content work whenever they compete for attention.

### Next gate authorization

Ready for dissent, then operator review. Not merged; no labels added.

muness:
## Ship-gate dissent — #397 RESOURCES

**Verdict:** pass-with-actions (two actions taken during this pass; residual risks below are accepted, not fixed)

**Role:** devil's advocate against the review above. Goal: find something real, not rubber-stamp.

### Status at time of writing

CI on this PR (targeting `v3`, so this is a real gate) is fully green as of `807d50c`: Lint, Test, Build Linux x64, Build WASM Assets, Synology Package Contract, Smoke Test Binary, and Build Summary all pass. A green CI run proves the code compiles and the existing test suite doesn't regress — it says nothing about whether the *right* tests exist, which is what this pass is actually for.

### What I went looking for

The review's own residual-risk section (P3, P4) already flagged the listChanged-delivery-is-unverified and the double-aggregator-read points and argued both are acceptable. Rather than re-litigate those, I looked for things the review didn't mention at all, on the theory that a self-review is likeliest to miss its own blind spots, not under-weight things it already saw.

### Re-checking P2 (the discoverability-hint fix) rather than taking it on faith

The review reports P2 as already fixed, which is exactly the kind of self-graded claim to distrust most. Checked directly rather than accepting the summary:

- Read the actual diff in `4ea6f17`: `resource_not_found`'s `data` gained a `hint` field, `"call resources/list for the current set of valid URIs"`.
- Read the tightened test: it now asserts the exact code (`-32002`, not just "some code"), the echoed `uri` at `/data/uri`, and that `/data/hint` contains `"resources/list"` — not merely that an error exists. Confirmed this would fail if the hint were removed by running it against a version with the field stripped back out (reverted locally, ran `cargo test --test mcp_contract unknown_or_stale_resource_uri_is_a_proper_error`, watched it fail on the missing-hint assertion, then restored).
- One gap the review didn't mention: the hint is generic ("call resources/list") rather than specific to the failure. For the `hifi://zones/{zone_id}` case specifically, `hifi_now_playing`'s tool-side equivalent goes further and names the exact recovery tool (`hifi_zones`) *and* explains why (`"No zone with id ... is known to the aggregator"`). The resource-side hint doesn't distinguish "you misspelled the scheme" from "this zone existed once and is now gone" — both produce the identical generic hint. Considered making `resource_not_found` take a reason parameter to differentiate. Decided against: the two `_ =>` branches converge on one error constructor specifically so a stale zone and a typo'd scheme can't accidentally get different codes if someone edits one branch and not the other, and the *action* is identical either way (`resources/list`) — there's no second string to send. Recorded as a real but accepted trade-off, not a defect.

**Verdict on P2:** the fix is real, tested to a standard that would catch a regression, and the one further gap found (generic vs. specific hint) doesn't change client-observable behavior enough to justify the complexity of a differentiated message. No further action.

### Finding: the acceptance criteria's literal minimum left three resources' "no second source of truth" claim untested

The issue's constraint is unambiguous: *"a resource and its equivalent tool must never be able to disagree."* That's stated as hard, not soft. The acceptance criteria then narrow the **testing** requirement to "at least zones and now-playing" — which the original implementation satisfied literally, while the *code* (payload-sharing functions) enforced the hard constraint for all five resources structurally.

That gap is real: `hifi_status`/`hifi://status`, `hifi_hqplayer_status`/`hifi://hqplayer/status`, and `hifi_hqplayer_profiles`/`hifi://hqplayer/profiles` all shared a payload function with their tool, but nothing would have caught a future refactor that broke that sharing for any of the three — e.g., someone "simplifying" `hqp_status_payload` by inlining it back into `handle_status` and hand-rolling a second version for the resource reader, plausible exactly because it looks like harmless deduplication cleanup. The zones/now-playing tests would stay green throughout.

**This is fixed in commit `807d50c`**, adding the three missing agreement tests, one of them (`hqplayer_status_resource_agrees_with_the_tool_for_a_live_connection`) against a real mock-server connection rather than the disconnected default, so it is not vacuously true.

### Trying to refute the fix itself

Is the new HQPlayer profiles test (`hqplayer_profiles_resource_agrees_with_hifi_hqplayer_profiles_tool`) actually load-bearing, or did I just add a test that always passes `[] == []`? Checked: yes, it's the weak one of the three — `get_cached_profiles()` returns nothing without a prior `list_profiles` round-trip the mock doesn't drive in this test. I left it in anyway because (a) it still exercises the real dispatch path end to end, catching a wiring mistake (wrong URI routed to the wrong handler, wrong JSON key names) even though it can't catch a payload-content divergence, and (b) driving `MockHqpServer` through an actual profile listing to make it non-trivial would need extending the mock's request handling for a family (`list_profiles`) this issue has no reason to touch — a second-order scope creep to avoid a first-order one. Recorded here rather than silently accepted so a reader can judge that trade differently.

### Re-examining the review's own residual risks, briefly

- **P3 (listChanged delivery unverified by a real stream test):** stands. I looked at whether a cheap fake could substitute for a real SSE round-trip and didn't find one — `TestApp` is `tower::oneshot`-based by design (per #394), and building a persistent-stream variant just for this would be exactly the over-building the issue warns against for its "least load-bearing" status. Accepted as documented best-effort, consistent with the issue's own text ("its best-effort nature... is acceptable and must be documented").
- **P4 (double aggregator read for now-playing):** stands, and I'll add one thing the review didn't: this is *only* benign because both reads happen within one request handler with no `.await` point that yields to another writer between them apart from the lock acquisition itself. If a future change inserted async work between the resource reader's `now_playing_payload` call and some other now-playing-affecting operation in the same request, the "cannot disagree in practice" argument would need re-checking. Not a change this PR makes, so not a blocker for this PR — but worth a sentence in case a future PR touching `resources.rs` assumes more atomicity than the code actually has.

### Contrary case: should this PR have gone further and covered `resources/subscribe` behind a feature check?

Considered arguing the SDK gate technically permits advertising `subscribe: Some(false)` explicitly (rather than omitting the field) as a more emphatic "no" than an absent key. Rejected: `Some(false)` and `None` are wire-identical after `skip_serializing_if = "Option::is_none"` in the schema (`ServerCapabilitiesResources`), so there is no behavioral or observable difference to argue for — this would be a distinction without a difference, and inventing an explicit `Some(false)` field ties this code to whatever `Default` the SDK ships that we might not need. No action.

### Disposition

Pass. One real, testable gap found and closed in this same PR; the two carried-over residual risks were re-examined and stand as accepted, documented trade-offs for the least load-bearing issue in the epic.

### Next gate authorization

Ready for operator review. Not merged; no labels added.

PR #427, opened 2026-08-01
feat(mcp): opaque content refs, and the one tool that consumes them (#396)

muness:
## Summary

Implements #396, the keystone issue of epic #392: `hifi_search` mints an opaque, short-lived `ref` alongside each result that has a durable-enough handle, and a new `hifi_play_ref` tool is the single documented way to resolve one back to a playable action. `hifi_play`'s existing query path (search → first match) is unchanged.

**Targets `v3`, so CI is a real gate.** This PR is not stacked on another feature branch.

**Update, read this first:** an earlier version of this PR passed local verification and CI, but **failed live end-to-end testing** on a real deployment — `hifi_play_ref` hung and timed out against a real Roon Core every time, immediately after a working `hifi_search`. Root-caused and fixed against the live rig, and **the fix has since been confirmed live end-to-end twice, with two independent items on the same zone**:

1. `hifi_search("kind of blue")` → only the real hit ("Kind of Blue" / Miles Davis) carried a `ref`, category rows ("Artists," "Albums," ...) correctly carried none → `hifi_play_ref` on that ref returned `outcome: accepted`, `"Play Now: Play Album"` → a follow-up read confirmed the zone's state actually changed to `playing`, album "Kind of Blue," artist "Miles Davis," track "So What."
2. `hifi_search("blue train")` → only "Blue Train" / John Coltrane carried a `ref`, all five category rows ("Artists" 28 Results, "Albums" 29, "Composers" 27, "Tracks" 36, "Works" 17) correctly carried none → `hifi_play_ref` on that ref returned `accepted` → `hifi_now_playing` confirmed the zone genuinely playing "Blue Train" / John Coltrane -- a different item from test 1, ruling out leftover state from the first test.

I independently re-verified the search half of both (not just taking the report on faith): both `hifi_search` calls, run myself against the same live rig, show exactly the ref/no-ref shape described. See "What live testing found" below for the full arc, including how the underlying bug was found and fixed — this is the central fact about this PR's history, not a footnote. Playback was paused afterward both times since this was testing, not a real listening session.

## Solution-space: already settled, not relitigated here

Per the epic's process, #396's solution-space gate already ran (see the issue body's "Decisions taken by the solution-space gate" section and its gate-report comments). This PR implements that design rather than re-deriving it:

- **Server-side ref table** keyed by a 128-bit random token, not a self-describing/signed token. A Roon target pairs the `item_key` with its minting `multi_session_key`.
- **`hifi_play_ref` is its own tool**, not an optional `ref` param on `hifi_play` — so a model can never send both `query` and `ref` and force the server to silently pick one.
- **No client-visible TTL field.** Lifetime lives in the tool descriptions; the recovery instruction lives in the refusal.

One place this PR *does* make a call the gate flagged as open: the gate's dissent (D4) argued against a per-target-kind TTL (durable LMS targets living longer than ephemeral Roon ones) on the grounds that it makes the documented contract secretly wrong in one direction. I found that argument correct and implemented **one uniform TTL** (15 min) for every ref regardless of provider or target durability — see `src/mcp/refs.rs`'s module docs for the reasoning. Not confirmed by the epic's own process; flagged as a residual risk below.

## What's covered

**Roon** (`src/adapters/roon.rs`):
- `RoonAdapter::search_with_session` surfaces the `multi_session_key` that `search()` mints and previously dropped. `search()` itself becomes a thin wrapper that drops the session key, so `/roon/search` is byte-for-byte unchanged.
- `RoonAdapter::play_ref` resolves a ref **inside the session that minted it**, exactly the same way `play_item` resolves a key — see "What live testing found" for why it does *not* reset the session with `pop_all` first, which an earlier version of this PR did and which broke against a real Core.
- `resolve_item_key` is the shared body `play_item` and `play_ref` both call, extracted as a pure refactor.
- `hifi_search`'s Roon branch never mints a ref for a search-result *grouping* row ("Albums (32 Results)", "Tracks", ...) — also found live; see below.

**LMS** (`src/adapters/lms.rs`):
- `LmsPlayTarget` extracted from `search_and_play`'s three-way branch (`Library`/`Url`/`GlobalSearchItem`). `LmsAdapter::play_target` is the one method `hifi_play_ref` uses to execute a *specific* target — never `search_and_play`, which re-searches and takes `results[0]`. `LmsRpc::execute` stays private.
- `GlobalSearchItem` — verified against live Lyrion 9.1.2 to be a positional breadcrumb (`<session>.<index>.<index>`) that a fabricated id silently re-resolves by position on cache miss — is **never mintable**. Only `Library` and `Url` targets are (`LmsPlayTarget::mintable`), with `Library` preferred over `Url` when both are present.
- `LmsAdapter::assert_library_id_exists` validates a `Library` target against the live database *before* the mutating `playlistcontrol` call, by re-searching for the title captured at mint time (not a dedicated existence query — see below for why that changed mid-review).

**Ref table** (`src/mcp/refs.rs`, new module): 128-bit random tokens (base64 URL-safe), bounded capacity (512) with oldest-inserted eviction, one uniform TTL. `AppState.mcp_refs` is constructed internally by `AppState::new` (like `sse_connections`), so no existing call site changed.

**MCP surface**: `McpSearchResult` gains an optional `ref` field (additive, `title`/`subtitle` untouched). New `hifi_play_ref` tool (`ref`, `zone_id`, `action?`), appended to `tool_box!` per the epic's append-only convention. Tool descriptions teach search → hold ref → play by ref and state refs are short-lived (`src/mcp/tools/library.rs`).

## What live testing found (read this)

This PR passed local verification and CI once, and I believed it was ready to ship. The operator then installed the actual build on production hardware, connected to a real Roon Core ("nuc14," Roon 2.70), and ran the feature: `hifi_search` worked; **`hifi_play_ref` failed every time** with `Browse request timed out`. `hifi_play` (the pre-#396 query path) worked fine on the identical zone and query, isolating the defect to `play_ref`'s resolution path.

I reproduced this directly against the live rig via the existing `/roon/browse` HTTP endpoint (the same underlying `RoonAdapter::browse()` call `play_ref` uses) and found:

1. **`play_ref` combined `pop_all: true` with a present `item_key` in one browse request. A real Core never answers that combination.** The fake never caught this because its search/browse model doesn't reproduce it.
2. The "obvious" fix — send them as two separate requests — **also doesn't work**: `pop_all` invalidates every item_key the session had previously minted, including the one about to be entered.
3. The reset was solving a problem that doesn't exist: resolving one ref, navigating deep into its action list, then resolving a second, sibling ref from the same search succeeds immediately **with no `pop_all` at all** — verified live, twice.

**Fix:** `resolve_item_key` never sends `pop_all`. It is now byte-for-byte `play_item`'s existing logic, just re-entering the caller's session. `tests/mock_servers/roon_core.rs` (`FakeRoonCore`) is strengthened with a session-epoch model so this class of bug is now caught by the test suite in milliseconds (`tests/mcp_refs.rs::play_ref_never_combines_pop_all_with_an_item_key`), not only by a live rig.

**Re-scrutinizing the same live data one more time** (the ship-gate dissent process, applied for real) found a second, independent bug: a real Core's search results mix the actual hit with *grouping* rows ("Albums (32 Results)", "Tracks", "Artists", ...) sharing the identical shape (`hint: "list"`, a real `item_key`) as genuine content. `hifi_search` was minting a ref for these indistinguishably from "Kind of Blue" itself — resolving one would land in an existing "guess the first item" fallback and could silently play unrelated content. Fixed by wiring in `is_category()`, a check that already existed in the adapter for exactly this set of names but was never consulted by the ref-minting code, plus an additional source-independent signal (subtitle shaped like `"<N> Results"`) as defense-in-depth for TIDAL/Qobuz sources I could not test live. **That second signal mitigates, but does not close,** the TIDAL/Qobuz gap — see the dissent comment.

**Full end-to-end confirmation, closing the loop**: the operator built and installed the fixed package on the NAS and ran the actual flow twice, with two different items (see the top of this PR body for both) — search, mint a ref, resolve it with `hifi_play_ref`, confirm via a follow-up read that the zone is genuinely playing the right thing. I independently re-ran both searches myself against the live rig and confirmed the same ref/no-ref shape.

State this precisely, since it's tempting to overclaim from two clean runs: **this is two successful invocations, on one zone, one provider (Roon), one action (`play`)** — not exhaustive coverage of every action (`queue`/`radio` on Roon, `play`/`queue`/`next` on LMS), every provider (LMS was not live-tested at all in this round), or every zone. It is strong, repeated evidence the core resolution path and the category-row fix both work against a real Core; it is not a claim that every code path in this PR has been live-exercised.

## Acceptance criteria

- [x] `hifi_search` returns opaque `ref` alongside `title`/`subtitle`, both providers, additive.
- [x] Exactly one consumer, `hifi_play_ref`, covering Roon (`play_ref`) and LMS (`play_target`, all three `LmsPlayAction`s where supported).
- [x] Roon's minting `multi_session_key` surfaced from `search_with_session`.
- [x] Opacity test — no item_key/entity id/session key/title appears verbatim; a hand-mangled ref is refused.
- [x] Unguessable/uncorrelated refs.
- [x] Expired/unknown ref → distinct retryable outcome, never a silent fallback to first-match.
- [x] Cross-provider ref/zone refusal — and the ref is not burned by the rejected attempt.
- [x] Ref table bounded in size and time; eviction observable as expiry, not corruption.
- [x] #394's no-orphan-field test passes with `ref` registered as consumed.
- [x] Tool descriptions teach the flow and say refs are short-lived.
- [x] (Found necessary by live testing, not in the original list) Roon ref resolution actually works against a real Core, not only the fake — **confirmed live end-to-end twice**, one zone, `action=play`, two different items (see above).
- [x] (Found necessary by live testing) a search-result grouping row never gets a ref — **confirmed live, twice**: both end-to-end runs' search results showed a ref on the real hit and none on any category row.

## Explicitly NOT covered / deferred, and why

- **Live coverage of every action/provider/zone combination**: the live confirmation above is two invocations (Roon, `action=play`, one zone). `queue`/`radio` (Roon) and `play`/`queue`/`next` (LMS) are covered by the fake and by deliberate-break verification, not by a live rig; LMS was not live-tested at all in this round.
- **TIDAL/Qobuz grouping-row verification** — mitigated by a source-independent heuristic, not confirmed live. See the dissent comment (D2).
- **LMS `globalsearch` ref-decay / Roon cross-session portability empirical questions** from the issue: no reachable live LMS rig for the first; the second is answered by design (`play_ref` works whether keys are global or session-scoped) and now additionally live-verified for the Roon side via the `pop_all` investigation.
- **`assert_library_id_exists`'s title-based re-search** is unverified against a live LMS server (its failure direction is closed either way — see the review comment for the two-iteration history of this specific check).
- **`Url` target validation before mutate**: LMS exposes no cheap existence check for an arbitrary streaming/plugin URL. Stated as a known gap in `LmsAdapter::play_target`'s doc comment.
- Doc sync (AGENTS.md, README, `plugin/commands/setup.md`) deferred to #401, matching the epic's established pattern.

## Verification

In order, per the epic's standing recipe, re-run clean after every fix in this PR's history (most recently after the `pop_all`, category-row, and TIDAL/Qobuz-mitigation fixes):
1. `make css`
2. `cargo fmt --check` — clean
3. `cargo clippy -- -D warnings` — clean (lib only, per the epic's standing note)
4. `cargo test --workspace` — all green: 172 lib tests, 51 `roon_protocol.rs`, 38 `mcp_refs.rs`, 94 `mcp_contract.rs`, and everything else pre-existing.
5. `dx build --release --platform web --features web` with `CARGO_BUILD_RUSTC_WRAPPER=""` — clean.

CI on `v3`: **green on the final commit** (`54ea84c`) — Plan, Lint, Test, Check API Changes, Synology Package Contract, Build WASM Assets, Build Linux x64, Build QNAP (x64), Smoke Test Binary, Build Summary, CodeRabbit all pass. Independently re-verified locally too (`make css`/`fmt`/`clippy` clean, `cargo test --test mcp_refs` 38/38 including all three live-rig-discovered regression tests).

**Every load-bearing claim in this PR was verified by deliberately breaking the corresponding implementation and confirming the test fails, then reverting** — including, this round, the three live-rig-discovered bugs above (`play_ref_never_combines_pop_all_with_an_item_key`, `a_category_grouping_row_never_gets_a_ref`, `a_grouping_row_with_an_unlisted_title_is_still_caught_by_its_subtitle_shape`), each confirmed to fail fast (well under a second) rather than only via a slow live-rig round trip.

See the review and dissent comments for the full gate report, including what the dissent found this round that the review didn't (a real bug, and a real unclosed gap) and what I checked and could not break.

## Summary by CodeRabbit

- **New Features**
  - Search results can include short-lived references for exact-result playback.
  - Added `hifi_play_ref` to play, queue, start radio, or play next a specific result.
  - Supports Roon and LMS playback targets while preserving the correct playback session.
  - LMS references are available for durable library and URL results.

- **Bug Fixes**
  - Invalid, expired, cross-provider, unsupported, or stale references now return clear refusals without changing the queue.
  - LMS playback targets are validated before queue changes.

### DIFF
--- a/src/adapters/lms.rs
+++ b/src/adapters/lms.rs
@@ -818,6 +818,78 @@ impl LmsPlayAction {
     }
 }
 
+/// One resolved way to play/queue/insert an LMS result, independent of how it
+/// was found.
+///
+/// Extracted from `search_and_play`'s three-way branch (issue #396) so
+/// `hifi_play_ref` can execute a *specific* target — not `results[0]` — via
+/// [`LmsAdapter::play_target`], without `LmsRpc::execute` becoming public.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum LmsPlayTarget {
+    /// `playlistcontrol cmd:<...> <kind>_id:<id>`. Durable: an LMS database
+    /// id, surviving both UHC and LMS restarts. Invalidated only by a rescan
+    /// that reassigns ids — which is why [`LmsAdapter::play_target`] verifies
+    /// it still exists before mutating the queue.
+    Library { kind: LmsSearchResultType, id: i64 },
+    /// `playlist <load|add|insert> <url>`. Mostly durable: a plugin-resolvable
+    /// URL that survives restarts and dies only if the plugin is removed or a
+    /// subscription lapses.
+    Url { url: String },
+    /// `globalsearch playlist <play|add|insert> item_id:<id>`.
+    ///
+    /// **Not an identifier.** Verified against live Lyrion 9.1.2: this is a
+    /// positional breadcrumb of the form `<session>.<index>.<index>` that
+    /// changes across identical calls, and a *fabricated* session id resolves
+    /// successfully — LMS silently re-walks the position on a cache miss — so
+    /// a stale or wrong value here plays the wrong content with **no error**.
+    /// #396's hard rule: never mint a ref through this variant. It exists
+    /// only so `play_target` can still serve `search_and_play`'s existing,
+    /// unref'd first-result path unchanged.
+    GlobalSearchItem { item_id: String },
+}
+
+impl LmsPlayTarget {
+    /// The target `search_and_play` resolves its (only ever first) hit to,
+    /// preserving the exact precedence that method always used — `item_id`,
+    /// then `url`, then a positive library id — so extracting `play_target`
+    /// changes no observable behavior on that path.
+    pub fn from_search_hit(result: &LmsSearchResult) -> Option<Self> {
+        if let Some(item_id) = &result.item_id {
+            Some(Self::GlobalSearchItem {
+                item_id: item_id.clone(),
+            })
+        } else if let Some(url) = &result.url {
+            Some(Self::Url { url: url.clone() })
+        } else if result.id > 0 {
+            Some(Self::Library {
+                kind: result.result_type,
+                id: result.id,
+            })
+        } else {
+            None
+        }
+    }
+
+    /// The target #396's ref minter should record for this result, or `None`
+    /// if nothing about it is safe to mint a ref for.
+    ///
+    /// Different precedence than [`Self::from_search_hit`], and deliberately
+    /// so: a ref is a promise that outlives this call, so it prefers whichever
+    /// handle is durable — `Library` (a persistent database id) over `Url`
+    /// (mostly durable) over minting nothing at all. `GlobalSearchItem` is
+    /// never mintable, full stop; see that variant's own docs for why.
+    pub fn mintable(result: &LmsSearchResult) -> Option<Self> {
+        if result.id > 0 {
+            Some(Self::Library {
+                kind: result.result_type,
+                id: result.id,
+            })
+        } else {
+            result.url.clone().map(|url| Self::Url { url })
+        }
+    }
+}
+
 /// Internal state
 struct LmsState {
     host: Option<String>,
@@ -1585,6 +1657,14 @@ impl LmsAdapter {
     /// Searches using globalsearch (includes streaming services), finds the first
     /// playable result, and executes the specified action (play, queue, or insert).
     /// Uses URL-based playback for streaming items, playlistcontrol for library items.
+    ///
+    /// #396: the actual command construction now lives in [`Self::play_target`],
+    /// reached through [`LmsPlayTarget::from_search_hit`] so this method's
+    /// observable behavior — including which of the three addressing paths a
+    /// result with more than one handle takes — is unchanged from before the
+    /// extraction. `hifi_play_ref`'s resolution path calls `play_target`
+    /// directly, with a target this method never sees: a *specific* result's,
+    /// not `results[0]`'s.
     pub async fn search_and_play(
         &self,
         query: &str,
@@ -1602,66 +1682,14 @@ impl LmsAdapter {
 
         // Take the first result
         let result = &results[0];
-
-        // Determine playback method based on item type:
-        // 1. item_id present (globalsearch streaming) -> use "globalsearch playlist play item_id:XXX"
-        // 2. url present (direct URL) -> use "playlist load/add URL"
-        // 3. otherwise (library item) -> use "playlistcontrol" with entity ID
-        if let Some(ref item_id) = result.item_id {
-            // Globalsearch streaming item - use globalsearch playlist command
-            let method = match action {
-                LmsPlayAction::Play => "play",
-                LmsPlayAction::Queue => "add",
-                LmsPlayAction::Insert => "insert",
-            };
-            let item_id_param = format!("item_id:{}", item_id);
-            self.rpc
-                .execute(
-                    Some(player_id),
-                    vec![
-                        json!("globalsearch"),
-                        json!("playlist"),
-                        json!(method),
-                        json!(item_id_param),
-                    ],
-                )
-                .await?;
-        } else if let Some(ref url) = result.url {
-            // Direct URL item - use playlist command with URL
-            let method = match action {
-                LmsPlayAction::Play => "load",
-                LmsPlayAction::Queue => "add",
-                LmsPlayAction::Insert => "insert",
-            };
-            self.rpc
-                .execute(
-                    Some(player_id),
-                    vec![json!("playlist"), json!(method), json!(url)],
-                )
-                .await?;
-        } else if result.id > 0 {
-            // Library item - use playlistcontrol with entity ID
-            let id_param = match result.result_type {
-                LmsSearchResultType::Album => format!("album_id:{}", result.id),
-                LmsSearchResultType::Artist => format!("artist_id:{}", result.id),
-                LmsSearchResultType::Track => format!("track_id:{}", result.id),
-            };
-
-            let cmd_param = format!("cmd:{}", action.to_lms_cmd());
-
-            self.rpc
-                .execute(
-                    Some(player_id),
-                    vec![json!("playlistcontrol"), json!(cmd_param), json!(id_param)],
-                )
-                .await?;
-        } else {
-            // No valid playback handle - this shouldn't happen if parse_single_item works correctly
-            return Err(anyhow!(
+        let target = LmsPlayTarget::from_search_hit(result).ok_or_else(|| {
+            anyhow!(
                 "No playable result found for '{}' (missing item_id, url, and valid id)",
                 query
-            ));
-        }
+            )
+        })?;
+
+        self.play_target(&target, player_id, action).await?;
 
         // Build response message
         let action_verb = match action {
@@ -1697,6 +1725,143 @@ impl LmsAdapter {
         );
 
         Ok(format!("{} {}", action_verb, what))
+    }
+
+    /// Execute one resolved [`LmsPlayTarget`]: `play`, `queue`, or `insert`
+    /// (play-next) where the target supports it.
+    ///
+    /// #396: this is the one path `hifi_play_ref` uses to resolve a ref. It
+    /// never re-searches and never takes `results[0]` — the defect the issue
+    /// exists to remove from ref resolution. `LmsRpc::execute`
+    /// (`src/adapters/lms.rs`) stays private; this is the purpose-specific
+    /// `pub` method that replaces handing MCP arbitrary CLI construction.
+    ///
+    /// # Validate before mutate
+    ///
+    /// A stale [`LmsPlayTarget::Library`] id is checked against the live
+    /// database *before* the mutating command runs, because
+    /// `playlistcontrol cmd:load` with an id that no longer exists clears the
+    /// queue and *then* fails — a wipe reported as a failure is still a wipe.
+    ///
+    /// **`Url` targets are not pre-validated.** LMS exposes no cheap existence
+    /// check for an arbitrary streaming or plugin URL, and `playlist load
+    /// <bad-url>` returns a normal `{}` while emptying the queue regardless of
+    /// whether it is checked first. This is a known, stated gap, not a
+    /// silently-accepted one — see the issue for why ref minting still
+    /// prefers `Url` over [`LmsPlayTarget::GlobalSearchItem`] in spite of it:
+    /// an unresolvable URL is at least a real failure mode, where a positional
+    /// breadcrumb is a *silent* one.
+    pub async fn play_target(
+        &self,
+        target: &LmsPlayTarget,
+        player_id: &str,
+        action: LmsPlayAction,
+    ) -> Result<()> {
+        let player_id = strip_lms_prefix(player_id);
+
+        match target {
+            LmsPlayTarget::GlobalSearchItem { item_id } => {
+                // Globalsearch streaming item - use globalsearch playlist command
+                let method = match action {
+                    LmsPlayAction::Play => "play",
+                    LmsPlayAction::Queue => "add",
+                    LmsPlayAction::Insert => "insert",
+                };
+                let item_id_param = format!("item_id:{}", item_id);
+                self.rpc
+                    .execute(
+                        Some(player_id),
+                        vec![
+                            json!("globalsearch"),
+                            json!("playlist"),
+                            json!(method),
+                            json!(item_id_param),
+                        ],
+                    )
+                    .await?;
+            }
+            LmsPlayTarget::Url { url } => {
+                // Direct URL item - use playlist command with URL
+                let method = match action {
+                    LmsPlayAction::Play => "load",
+                    LmsPlayAction::Queue => "add",
+                    LmsPlayAction::Insert => "insert",
+                };
+                self.rpc
+                    .execute(
+                        Some(player_id),
+                        vec![json!("playlist"), json!(method), json!(url)],
+                    )
+                    .await?;
+            }
+            LmsPlayTarget::Library { kind, id } => {
+                self.assert_library_id_exists(player_id, *kind, *id).await?;
+
+                let id_param = match kind {
+                    LmsSearchResultType::Album => format!("album_id:{}", id),
+                    LmsSearchResultType::Artist => format!("artist_id:{}", id),
+                    LmsSearchResultType::Track => format!("track_id:{}", id),
+                };
+                let cmd_param = format!("cmd:{}", action.to_lms_cmd());
+
+                self.rpc
+                    .execute(
+                        Some(player_id),
+                        vec![json!("playlistcontrol"), json!(cmd_param), json!(id_param)],
+                    )
+                    .await?;
+            }
+        }
+
+        Ok(())
+    }
+
+    /// Confirm a [`LmsPlayTarget::Library`] id still exists before a mutating
+    /// `playlistcontrol` call touches the queue.
+    ///
+    /// Uses the dedicated taggedlist query for `kind` (`albums`/`artists`/
+    /// `titles`), filtered by the same `<kind>_id` tag that query's own result
+    /// rows are keyed by (`album_id`, `artist_id`, `track_id`) — the shape
+    /// LMS's published CLI reference documents for these three queries.
+    /// **Unverified against a live server**: this repo has already found one
+    /// LMS CLI naming surprise 9.1.2 disagreed with the published docs on
+    /// (`contributors_loop[].contributor_id`, not `artist_id` — see
+    /// `search_library`'s doc comment), so treat this filter's exact tag name
+    /// with the same caution until it is checked against a real instance.
+    async fn assert_library_id_exists(
+        &self,
+        player_id: &str,
+        kind: LmsSearchResultType,
+        id: i64,
+    ) -> Result<()> {
+        let (query, id_tag) = match kind {
+            LmsSearchResultType::Album => ("albums", "album_id"),
+            LmsSearchResultType::Artist => ("artists", "artist_id"),
+            LmsSearchResultType::Track => ("titles", "track_id"),
+        };
+        let result = self
+            .rpc
+            .execute(
+                Some(player_id),
+                vec![
+                    json!(query),
+                    json!(0),
+                    json!(1),
+                    json!(format!("{}:{}", id_tag, id)),
+                ],
+            )
+            .await?;
+        let count = result.get("count").and_then(|v| v.as_i64()).unwrap_or(0);
+        if count < 1 {
+            anyhow::bail!(
+                "{} id {} no longer exists in the LMS library (rescanned since this ref \
+                 was minted?); refusing rather than issuing playlistcontrol, which would \
+                 clear the queue and then fail",
+                kind,
+                id
+            );
+        }
+        Ok(())
     }
 }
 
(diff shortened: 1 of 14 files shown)


muness:
## BLOCKING — live-rig test reveals `hifi_play_ref` fails on a real Roon Core

Operator installed the QNAP build from this PR (post-merge with `v3`/#397) on their production NAS, connected to a real Core (`nuc14`), and I ran a live end-to-end probe against it (`192.168.1.2:8088`).

**`hifi_search`** works correctly — real results, opaque `ref` fields present, envelope shape correct.

**`hifi_play_ref`** fails every time, reproduced twice independently with a fresh search → immediate play_ref in the same MCP session (zero delay, ruling out staleness):
```json
{"outcome":"error","refusal":{"reason":"backend_error","detail":"Play error: Browse request timed out"}}
```

This is the one consumer path this issue exists to deliver, failing on the one thing #405s live-verified evidence said should work cleanly (a fast, correlated response) — instead it silently hangs to the full `BROWSE_TIMEOUT`. All prior verification here (unit tests, `FakeRoonCore` integration tests, CI) passed, but none of it exercised a real Core doing the actual resolve — only the fake, which models the *shape* of session-scoping but evidently not some behavior a real Core requires (most likely: a session needs an initializing browse call before an `item_key` means anything to Core, and resolution may be presenting the key to the minting session without re-establishing that session's hierarchy first).

**Do not merge as-is.** Flagged to the implementing agent to reproduce against the live rig and fix before the ship gate can pass. This is exactly the class of finding #392 is designed to surface: a design that looked sound against evidence and a fake, unproven against the one thing that matters.

muness:
## Gate report — ship gate, review role

**Gate/role:** Ship gate for #396, re-run after live end-to-end testing on a real deployment found the keystone path broken. This supersedes the earlier `pass-with-actions` conclusion reached from local verification and CI alone.

**Verdict:** `pass-with-actions`, but only after a P1 defect was found live, root-caused, and fixed during this gate. Headline: **local tests and green CI were not enough to prove this issue's one real deliverable actually worked, and it did not, until this pass.**

---

### What happened, in order

1. Implementation landed; self-review found and fixed two real bugs (a fail-open LMS validation bug, an inconsistent `operation` field); CI went fully green on `v3` (Lint, Test, Build, Synology/QNAP contracts, CodeRabbit).
2. The operator installed the actual QNAP build from this PR on production hardware, connected to a real Roon Core ("nuc14", Roon 2.70), and ran the feature this issue exists to deliver: `hifi_search` then `hifi_play_ref` on a real zone. `hifi_search` worked. **`hifi_play_ref` failed every time**, with `Browse request timed out`. A second, independent check confirmed `hifi_play` (the pre-#396 query path) worked perfectly on the identical zone and query, isolating the defect to `play_ref`'s resolution path specifically -- not Roon, not the Core, not the network.
3. Reproduced directly against the live rig via the existing `/roon/browse` HTTP endpoint (the same `RoonAdapter::browse()` call `resolve_item_key` uses internally, reached through a different Rust call site) -- not the fake, the real Core. Root-caused to one thing: `RoonAdapter::play_ref` combined `pop_all: true` with a present `item_key` in one browse request, and a real Core never answers that combination at all.
4. Live testing went further and found the "fix" that looks obvious (send `pop_all` and `item_key` as two separate requests) does not work either: `pop_all` invalidates every item_key the session had previously minted, including the one about to be entered next. It also found the reset was solving a problem that does not exist: resolving one ref, navigating deep into its action list, and then resolving a second, sibling ref minted by the same search succeeds immediately with **no `pop_all` at all**, verified live at two different navigation depths.
5. Fixed: `resolve_item_key` never sends `pop_all`. It is now byte-for-byte `play_item`'s existing, proven logic, with only the session key swapped for the caller's minting session. `tests/mock_servers/roon_core.rs`'s fake was strengthened (session-epoch key invalidation on `pop_all`, and a fast, structural rejection of the illegal combination) so this exact class of regression is now caught by the test suite in milliseconds, not only by a live rig.
6. Re-scrutinizing the *same* live search response one more time (rather than moving on) found a second, independent bug: real search results mix the actual hit with grouping rows ("Albums (32 Results)", "Tracks", "Artists", ...) sharing the identical `hint`/`item_key` shape as genuine content, which `hifi_search` was minting refs for indistinguishably. Fixed by wiring in the adapter's own pre-existing `is_category()` check (previously unused by the MCP layer) plus an additional, source-independent subtitle-shape signal as defense-in-depth for TIDAL/Qobuz sources, which were not live-tested. See the dissent comment (D1, D2) for the full account, including what remains unverified.
7. CI on the final commit went fully green (Lint, Test, Build, Synology/QNAP contracts, CodeRabbit), independently re-verified locally.
8. **The one gap left after item 7 — an actual end-to-end `hifi_play_ref` call against a rebuilt package, with real playback — has since been closed.** The operator rebuilt and installed the fixed package, then ran the full flow twice with two different items on the same zone: `hifi_search("kind of blue")` → ref on the real hit only, none on any category row → `hifi_play_ref` → `accepted`, `"Play Now: Play Album"` → confirmed playing "Kind of Blue"/Miles Davis/"So What". Then, independently, `hifi_search("blue train")` → ref on "Blue Train" only, none on any of five category rows → `hifi_play_ref` → `accepted` → `hifi_now_playing` confirmed playing "Blue Train"/John Coltrane — a different item, ruling out leftover state from the first run. I independently re-ran both searches myself against the live rig (not the fake) and confirmed the identical ref/no-ref shape reported.

### Evidence inspected

- The live reproduction sequence (search → mint ref → resolve, and isolated `/roon/browse` probes with and without the illegal combination) against `192.168.1.2:8088` / Core `nuc14`, run directly, not taken on faith from the operator's report alone.
- A full grep of `src/adapters/roon.rs` and `src/api/mod.rs` for every remaining `pop_all` use: two others exist (`search_with_session`, `search_and_play`'s root navigation), both send it alone, never combined with `item_key` -- confirming the bug was isolated to the one function this PR added, not a pattern already latent elsewhere.
- `tests/mcp_refs.rs::play_ref_never_combines_pop_all_with_an_item_key`, verified by deliberately reintroducing the fixed `pop_all: true` and confirming it fails in well under a second (not a 10-second hang) -- via both the timing assertion and `FakeRoonCore::illegal_pop_all_with_item_key_attempts`.
- `tests/mock_servers/roon_core.rs`'s new `Session::minted_at_epoch`/`epoch` model, checked for the failure mode it could have introduced (over-restricting cross-session key reuse under `ItemKeyScope::Global`, which `play_item` legitimately depends on): a first version did break `play_item_resolves_an_item_key_to_a_play_action` and two others; corrected to only restrict keys *this session itself* minted and later invalidated. All 51 `roon_protocol.rs` tests pass with the corrected model.
- Full local re-verification after the fix: `make css`, `cargo fmt --check`, `cargo clippy -- -D warnings`, `cargo test --workspace` (172 lib + 51 roon_protocol + 38 mcp_refs (after the pop_all, category-row, and TIDAL/Qobuz-mitigation fixes), all green).

### Findings

#### P1 — found and fixed during this gate

- The `pop_all`/`item_key` combination above (item 3-5). This is the headline finding.
- The search-result category/grouping-row ref leak (item 6). Fixed for the one source (Library) tested live; mitigated but **not confirmed** for TIDAL/Qobuz — see the dissent's D2, which I am not resolving in this review's favor by omission.

#### P2 (carried forward from the prior pass, still real, not re-litigated here)

- Uniform-TTL design choice (`src/mcp/refs.rs`) is reasoned but unilateral, still unconfirmed by the epic's process.
- `assert_library_id_exists`'s title-based re-validation is unverified against a live server (though its failure direction is closed, and was itself found and fixed once already this gate).

#### P3 — closed since the first draft of this report

This review originally flagged that the fix's own verification had a real limit: browse mechanics were verified live via `/roon/browse`, but not a full `hifi_play_ref` call end-to-end with real playback, since that needed a rebuilt package and starts real audio in someone's home. **That gap is now closed** (item 8 above): two independent end-to-end runs, different items, same zone, both confirmed accepted and genuinely playing via a follow-up read (`hifi_now_playing` for the second, ruling out the first test's residual zone state producing a false positive). Precisely stated, this is two invocations covering one zone, one provider (Roon), one action (`play`) — not every action/provider/zone combination; see Residual risk.

---

### Contrary case

The strongest case against `pass` even now: this is the second time in this same PR that "verified, tested, green" turned out wrong once real hardware was involved (first the LMS filter-tag guess, now the Roon `pop_all` behavior) -- and both times the bug was one this repo's own fake did not model, because the fake was built from the fork's source and published docs, not a real Core, and said so explicitly in its own PROVENANCE table. That hedge was correct. The honest update is that the hedge was load-bearing, not decorative, and a keystone issue's ship gate should have insisted on live verification of the resolution path *before* the first "pass," not after an operator happened to install it. That is a process gap worth the epic owner's attention beyond this one PR.

### Disposition

Ship. The one condition the earlier draft of this report set — confirmed end-to-end live playback — is now met, twice, with independent verification of the search half by me directly against the rig rather than taken on the operator's report alone.

### Residual risk

- **Medium**: TIDAL/Qobuz search-result grouping rows are unverified against a live source of either kind; the mitigation (subtitle-shape check) is plausible, not confirmed.
- **Medium**: the two live confirmations cover Roon only, `action=play` only, one zone. `queue`/`radio` (Roon) and all of LMS remain verified by the fake and deliberate-break testing, not by a live rig.
- **Low-medium**: the two P2 items carried forward, unchanged in severity from the prior gate pass.
- **Low**: the fake's new epoch model is an inference from a handful of live probes (root, one search, two siblings, one deep navigation), not an exhaustive characterization of every Roon Core browse-session rule -- stated as such in its own doc comments.

### Next gate authorization

Do not merge (operator's own call, to be made after reviewing this and the dissent). No further gate action required from this role; the one open condition from the prior draft is closed.

Dissent report follows in a separate comment.

muness:
## Gate report — ship gate, dissent role

**Gate/role:** Same gate as the preceding review comment, re-run after live testing found the keystone Roon path broken and the review's re-write above described the fix. Attacking that fix, and re-scrutinizing the two paths I was specifically asked to look at again: the operation-field fix, the LMS validate-before-mutate fix, and the Roon session-reuse resolution path.

**Verdict on the review:** The review is right that the `pop_all` fix is correct and live-verified at the browse-mechanics level. Re-scrutinizing the same live data one more time, as asked, found a **third** real bug in the same area, already fixed during this pass (below) — and left one real, honestly-unclosed gap the review doesn't mention.

---

### D1 — Found by looking harder at the Roon path, as asked: category rows were getting refs too, and could silently misplay

The review's evidence section describes the `pop_all` fix and stops. Going back through the actual live search response one more time (not just the one item I'd already extracted a ref from) turned up something the fake never exercised: **a real Core's search results mix the top hit with grouping rows** — searching "kind of blue" against nuc14 returned "Kind of Blue" *and* "Artists" (7 Results), "Albums" (32 Results), "Composers", "Tracks", "Works", every one sharing the identical `hint: "list"` shape and a real, navigable `item_key`. `handle_search`'s ref-minting logic checked only `item.item_key.is_some()` — so it minted a ref for "Albums (32 Results)" exactly as readily as for "Kind of Blue." Resolving that ref would land in `resolve_item_key`'s "guess the first item" fallback and could silently play some unrelated album out of the 32 — the precise failure mode (misresolve to real-but-wrong content) this whole issue exists to prevent, reached through a path this PR's own Roon test suite never exercised because `FakeRoonCore`'s search model is a flat, substring-matched list that never produces grouping rows at all.

This was fixable immediately and cheaply because the adapter already had the answer sitting unused: `is_category()` (`src/adapters/roon.rs`, checks title against `CATEGORY_NAMES`) already existed, already used elsewhere (`try_navigate_to_playable`) to skip exactly these rows — the MCP ref-minting loop simply never consulted it. Made `pub(crate)`, wired into `handle_search`, and proven with a test (`a_category_grouping_row_never_gets_a_ref`) built from a custom library that includes both a real hit and a category row, verified to fail without the check and pass with it.

**What this means for the review's confidence, not just this one bug**: the review's live verification covered exactly the one code path the operator happened to test (a single top-hit album). The category-row bug lived one field over, in data the *same* live search response already contained, and was missed until asked to look again specifically. That is worth naming as a pattern: this PR has now had a live-rig-shaped discovery surface a real bug three separate times (the `pop_all` hang, the LMS fail-open validation, and this), each one found by looking harder at data already in hand rather than by new probing. I would not bet the fourth pass is clean either.

### D2 — Mitigated, not closed: TIDAL and Qobuz sources are still unverified for the same bug

`is_category`'s list (`Albums`, `Tracks`, `Artists`, `Works`, `Composers`, `Genres`, `Tags`) was derived from `src/adapters/roon.rs`'s pre-existing code, not from anything I observed live for TIDAL or Qobuz. I only had a Library-source search to test against nuc14 (the operator's zone was set up for that source; I did not ask for a TIDAL/Qobuz test to avoid burning more of their time on this). **If a streaming service's grouping rows use different labels** (e.g. "Playlists," a category Library search never returns and `CATEGORY_NAMES` does not list), the exact same bug ships for `source=tidal`/`source=qobuz` searches, unfixed, because nothing catches it -- the fake still cannot produce this shape for any source.

Added a second, source-independent signal during this same pass
(`is_ungrounded_grouping`, `src/adapters/roon.rs`): every grouping row observed
live had a subtitle shaped exactly `"<N> Results"`, a pattern plausibly generated
by the same Roon-side grouping logic regardless of source, distinct from any
real hit's subtitle (an artist or album name). This is additive and can only
*remove* refs the title check missed, never mint a worse one, so it was safe to
include despite being equally unverified for TIDAL/Qobuz --
`tests/mcp_refs.rs::a_grouping_row_with_an_unlisted_title_is_still_caught_by_its_subtitle_shape`
proves it catches a title `CATEGORY_NAMES` does not list. **This mitigates the
gap; it does not close it.** If TIDAL/Qobuz's grouping rows use neither a listed
title nor that subtitle shape, the bug still ships for those two sources
unnoticed. I have positive evidence the failure mode exists for one source, a
plausible-but-unverified mitigation for the other two, and zero direct evidence
either way that the mitigation actually holds there.

### D3 — Re-scrutinized the operation-field fix for completeness: found it consistent, but only after tracing a case the review's evidence didn't mention

Went through every `handle_play_ref` return path again against `hifi_control`'s and `hifi_play`'s own conventions, specifically looking for a spot where an unvalidated `action` string could still leak into `operation`. The zone-refused path (`refuse_library_zone`) and the unknown-ref path both use the `"play_ref"` constant, matching `hifi_play`'s own zone-refused convention (`Envelope::write("hifi_play", "play")`, never `args.action`) -- this was itself a fix made during this same pass, after finding the *first* version of the operation-field fix had exactly the bug it was supposed to remove (`operation` echoing the raw, unvalidated `action` string, e.g. `"next"` on a Roon ref, when refused). I verified this by probing the live wire shape directly (a temporary assertion against the actual JSON, not an assumption) before writing the fix, and it is now covered by tests (`operation_of(&played) == "play_ref"` in both the cross-provider and unknown-ref tests, `== "invalid_action"` in the refused-action test). I did not find a remaining gap in this path on this pass, but I note the same fix went through two iterations to get right, which is more evidence for D1's pattern than a clean bill of health.

### D4 — Re-scrutinized LMS validate-before-mutate: checked for the Roon-shaped bug there too, found none, but the check itself is worth stating plainly rather than assuming

Given D1's shape (a real Core's response includes non-content rows the fake doesn't model, and the MCP layer didn't filter them), I checked whether LMS has an analogous leak: could `search_library`'s results include a "category" row indistinguishable from real content? No -- `parse_single_item`/`search_library`'s own parsing only ever emits `Album`/`Artist`/`Track` typed results with a durable id, and LMS's `is_audio` gate (already existing, pre-#396) refuses to emit anything that isn't independently confirmed playable. There is no LMS equivalent of a Roon "Albums (32 Results)" row reaching `LmsSearchResult` at all, so `LmsPlayTarget::mintable` was never exposed to this class of bug. This is a real check, not an assumption -- but it is also the first time this specific question was asked and answered on this pass; had LMS's parsing had the same gap, it would very plausibly not have been caught until a live LMS server, exactly like the other three bugs in this PR.

One lower-severity note found while re-checking `assert_library_id_exists`: it is a check-then-mutate across two separate RPC calls, so a rescan landing in the gap between them is a real (if vanishingly unlikely) TOCTOU window the design doesn't claim to close -- worth naming for completeness, not worth fixing, since the whole feature is a best-effort staleness check, not a transaction.

---

### Disposition

**Update**: the live end-to-end confirmation this disposition was contingent on has since landed — two independent runs (different items, same zone), both `accepted` and confirmed genuinely playing, with the search half independently re-verified by me against the rig directly. That condition is met. Ship, with D2 (TIDAL/Qobuz category rows) still flagged as an unclosed, known-shaped gap rather than an assumed-fine one — the two live runs were both Library-source searches on Roon, so they say nothing about D2 one way or the other. D1 is fixed and tested; D2 is not, and should not be quietly folded into "the category-row bug is fixed" language anywhere this PR is described.

### Residual risk (mine, additive to the review's)

- **Medium**: D2 — TIDAL/Qobuz search-result category rows are unverified; the same ref-minting bug D1 fixed for Library may still be live for those two sources.
- **Low**: the pattern itself (D1's closing paragraph) -- three real bugs found in the same PR by looking harder at data already in hand is evidence this surface has more of the same kind still in it, not a completed audit.
- **Low**: TOCTOU window in `assert_library_id_exists` (D4), accepted as inherent to a best-effort check, not a defect.

muness:
## Response to CodeRabbit's review (3 actionable comments)

Checked each against current code before acting — one was real, one was already an accepted documented trade-off, one was a false positive.

**1. `src/mcp/refs.rs` — `evict_to_capacity` shared-budget gap. Real bug, fixed in `29102d1`.** `evict_to_capacity` only looped while `entries.len() >= capacity`. A ref expiring on read (`resolve`) shrinks `entries` without touching `order`, so a table whose refs mostly expire before ever being resolved — the common case — never trips that check, and `order` grows by one token per mint forever: a slow, unbounded memory leak over a long-running server's lifetime even though `entries` itself stayed correctly bounded. Fixed by gating the eviction loop on `order.len()` too. New test `order_does_not_leak_when_expiry_outpaces_eviction` proves it: verified red against the reverted condition (leaked to 50 entries), green with the fix.

**2. `src/adapters/lms.rs` — shared validation-search-limit budget across result kinds. Already an accepted, documented, bounded risk, not a new find.** `assert_library_id_exists`'s own doc comment (added in `cdf0a84`, the commit CodeRabbit reviewed) already states this exact scenario — a shared-title collision beyond `LIBRARY_VALIDATION_SEARCH_LIMIT` causing a false-negative refusal — by name, calls it "a known, bounded, fail-*closed* residual risk, not a claim that no library can ever trigger it," and has its own regression test (`lms_a_valid_ref_beyond_the_validation_search_limit_still_resolves`) that already forced the limit constant up once in response to exactly this failure mode. Per-kind independent limit windows would be a genuine future improvement, but this isn't an unaddressed oversight — leaving as documented.

**3. `src/mcp/tools/library.rs` — `operation` "next" vs "insert" is not a mismatch.** `hifi_play_ref`'s own accepted action vocabulary is `LMS_REF_ACTIONS = ["play", "queue", "next"]` — a closed set specific to this tool, intentionally distinct from `hifi_play`'s vocabulary (via `lms_action_name`, which uses "insert"). `env.operation = action_name.to_string()` correctly reports the value from *this tool's* vocabulary, which is exactly what `LMS_REF_ACTIONS` advertises as accepted and exactly what the client sent. Two tools legitimately using two different, each internally-consistent names for the same underlying LMS operation is not a bug.

The 7 nitpicks (extract-shared-helper suggestions, an undocumented `radio` action, missing action aliases, an additional LMS test) are real, low-severity code-quality suggestions I'm leaving as follow-up rather than blocking this PR further — none of them are correctness issues.

Issue #438, opened 2026-08-02
Aggregator projections: remove direct adapter state and status reads from every surface


muness:
Parent: #434

## Problem

Unified zone reads mostly use `ZoneAggregator`, but adapter-specific zone, now-playing, status, HQPlayer native state, and MCP status paths still query adapter caches or trigger adapter refreshes directly.

## Outcome

Expand aggregator-owned observations/projections so every readable HTTP, knob, web, and MCP value comes from admitted bus events. Refresh is an adapter lifecycle/polling concern, never initiated by a read handler.

## Acceptance

- Adapter-specific zone/now-playing endpoints project the aggregator while preserving payloads.
- Connection/health status for every adapter is retained from bus lifecycle/health events.
- HQPlayer pipeline/profile/matrix/options/native status are published on the bus and admitted coherently by the aggregator.
- Reads never call `get_status`, `get_zones`, cached-player, refresh, or instance lookup methods on adapters/managers.
- Staleness and unavailable state are explicit and consistently projected.
- Existing public API contracts remain unchanged.

Depends on: #434, #440
OH: 80222d6d

muness:
Dependency clarification: blocked by #440. Projection consistency requires source epoch + monotonic sequence, stale/duplicate rejection, explicit gap detection and bus-mediated snapshot recovery, atomic provider snapshots where fields must be coherent, freshness metadata, and a committed revision/watermark that command completion can await. Command results do not mutate canonical observed state.

Issue #437, opened 2026-08-02
Typed bus services: move browse, configuration, artwork, profiles, and instance operations behind the app boundary


muness:
Parent: #434

## Problem

Even after generic commands and state reads are corrected, surfaces directly invoke Roon browse/load/search, artwork fetching, LMS/HQPlayer configuration, HQPlayer profiles/matrix/pipeline, and instance-manager operations. Concrete adapters remain embedded in `AppState` to support those calls.

## Outcome

Define typed bus request families and dedicated adapter-side consumers for backend-specific operations. Keep configuration persistence and lifecycle ownership in application services/coordinator, but require all native adapter I/O to cross the bus. Remove concrete adapters and adapter-returning managers from surface `AppState`.

## Acceptance

- Roon browse/search/load and artwork proxy requests are correlated bus operations.
- HQPlayer pipeline/profile/matrix/instance native I/O is bus-routed with coherent verified observations.
- LMS and HQPlayer configure/reconnect operations use coordinator/application requests; surfaces never stop/start adapters directly.
- No service returns `Arc<ConcreteAdapter>` to a caller.
- Composition root and adapter workers are the only production owners/importers of concrete adapter types.
- Existing public API contracts and credential-handling behavior remain unchanged.

Depends on: #434, #435, #436
OH: 80222d6d

muness:
Dependency correction: implementation follows #435 and #438; #436 runs throughout as the shrinking enforcement baseline. Parent: #434.

muness:
HQPlayer latency contract: profile loads are long-running mutations, not ordinary 3-second RPCs. Replace the current shared HTTP-client timeout behavior with an operation-specific deadline and poll/readback convergence until the requested profile is observed, a terminal provider error occurs, or the long deadline expires. Run slow profile/config work on a serialized per-instance mutation lane so it cannot head-of-line block short transport/volume commands or aggregator reads. Publish pending/progress/confirmed/indeterminate observations through the bus; do not optimistically overwrite canonical settings. Fast operations retain their own short budgets.

Issue #436, opened 2026-08-02
Architecture lint: deterministically inventory and reject surface-to-adapter bypasses


muness:
Parent: #434

## Problem

`tests/architecture_lint.rs` is substring-based and its broad function-name exemptions permit nearly every current bypass. It can miss aliases, formatting changes, manager lookups, imports, and new methods.

## Outcome

Use `syn` to inspect production Rust syntax and identify concrete adapter dependencies from surface modules. Land an exact, reviewed debt baseline first so CI rejects any new bypass immediately. As #435, #437, and #438 remove entries, shrink the baseline; completion requires an empty baseline and removal of the migration escape hatch.

## Acceptance

- Findings use stable identities: module, enclosing item, adapter field/type, and operation—not line numbers.
- CI fails for new concrete adapter imports, `AppState` adapter fields, direct state field access, manager-to-adapter lookup, or adapter method calls in surface modules.
- Tests include positive fixtures for aliases/formatting/nested expressions and negative fixtures for composition root, adapter implementation, aggregator event application, and test doubles.
- The final gate has no handler-name exemptions and no baseline entries.
- Existing API route contracts are unchanged.

Depends on: #434
OH: 80222d6d

muness:
This issue lands first as no-new-debt enforcement, but is not complete until #435, #438, and #437 reduce its exact baseline to zero.

muness:
Implementation PR #460 extends the original AST guardrail with deterministic provider-recovery/projection checks: LMS snapshot readmission, Roon Core-loss removal, OpenHome/UPnP cached-device readmission, startup readiness, and stop/recovery ordering. #436 remains open because its broader surface-to-adapter migration baseline is intentionally not yet empty.

Issue #435, opened 2026-08-02
Bus command gateway: route all zone transport and volume commands through correlated events


muness:
Parent: #434

## Problem

HTTP, knob, and MCP transport paths call Roon, LMS, OpenHome, UPnP, and HQPlayer executors directly. `CommandReceived`/`CommandResult` and `AdapterLogic::handle_command` exist but are not wired in production.

## Outcome

Create one typed request/reply gateway over the in-app bus and migrate every generic transport, seek, mute, and volume surface to it. Adapter workers consume only commands for their prefix and publish one correlated terminal result.

## Acceptance

- HTTP, knob, web, and MCP use the same gateway and error taxonomy.
- Correlation IDs cannot cross-deliver concurrent replies.
- Timeout, cancellation, adapter unavailable/stopping, lagged receivers, duplicate results, and shutdown are deterministic and tested.
- HQPlayer verified readback is preserved before success is published.
- No surface calls `.control`, transport, or volume methods on an adapter.
- Existing public API contracts remain unchanged.

Depends on: #434, #440
OH: 80222d6d

muness:
Dependency clarification: blocked by #440. The gateway must register correlation before dispatch, use bounded per-adapter queues and absolute deadlines, classify pre-dispatch timeout separately from in-flight/indeterminate timeout, avoid unsafe automatic retries, and return `confirmed` only after the aggregator commits the referenced observation watermark. A provider-level ACK without readback is `accepted`, not `confirmed`.

PR #439, opened 2026-08-02
test(architecture): enforce adapter boundary debt

muness:
## Outcome

Makes the adapter boundary deterministic and ratcheting: production application modules may depend on the aggregator and in-app bus, but may not introduce new concrete adapter dependencies.

Advances #436
Parent: #434

## What changed

- replaces the substring-based API lint and its broad status/control/config exemptions with a `syn` AST walk
- inventories direct `AppState` adapter access, concrete adapter fields, imports, and types
- records the 108 existing findings as an exact, stable migration baseline
- rejects both new findings and stale baseline entries, so every migration must reduce the debt in the same PR
- limits exemptions to native adapter/bus/aggregator infrastructure and binary composition roots
- includes syntax mutation tests for aliases, nested/reference access, imports, concrete types, and sanctioned paths

The baseline is intentionally non-empty. This PR prevents new bypasses; #435, #437, and #438 remove the existing ones. #436 completes only when the baseline reaches zero.

## Verification

- `cargo test --test adapter_boundary_lint --quiet`
- `cargo test --test architecture_lint --quiet`
- `cargo clippy --all-features -- -D warnings`
- `cargo test --all-features --quiet`
- `git diff --check`

No API contract change.


muness:
Superseded by #428. The AST guardrail and its regenerated combined HQPlayer baseline were transplanted onto the single consolidated branch in commits 9d4701f and 2fc1fb7. All #434 execution continues there.

